PR #82, opened 2019-10-24
Fix softargmax-bench CMake target.

AshkanAliabadi:


AshkanAliabadi:
@googlebot I signed it!

PR #185, opened 2019-11-20
Rename f32-argmaxpool source files in CMake.

AshkanAliabadi:


CMakeLists.txt
```diff
@@ -156,16 +156,17 @@ SET(XNNPACK_OPERATOR_SRCS
 SET(XNNPACK_COLD_SRCS ${XNNPACK_OPERATOR_SRCS})
 LIST(APPEND XNNPACK_COLD_SRCS
   src/init.c
+  src/memory.c
```
AshkanAliabadi: Not sure you want to categorize this as hot or cold.

Maratyszcza: Cold LGTM. Only the code that runs inside `xnn_run_operator` (e.g. thread pool entry points) is considered hot.

PR #267, opened 2019-12-10
Rename strided gemm and igemm fma3 broadcast files.

AshkanAliabadi:


AshkanAliabadi:
Alternatively I can change CMakeLists.txt.  Which one do you prefer?

Maratyszcza:
I prefer to keep function names in sync with file names, so changing `BUILD.bazel` and codegen scripts is preferable.

PR #284, opened 2019-12-18
Tighten the compiler detection guards in intrinsics-polyfill.h.

AshkanAliabadi:


AshkanAliabadi:
Tested on godbolt.org.  Clang CI failures triggered this change.

AshkanAliabadi:
This still seems not to work on older versions of Apple Clang since the versioning is different.  Need to investigate.

PR #383, opened 2020-03-02
Update pthreadpoool for support Windows

mattn:
See https://github.com/Maratyszcza/pthreadpool/pull/6


mattn:
Ah, sorry. I seems mistake how to calculate sha256sum. What is this value?

PR #414, opened 2020-03-12
Remove non-trivial designated initializers.

syoyo:
gcc(confirmed with gcc 7.5.0) fails to compile some benchmark code due to this:

https://stackoverflow.com/questions/31215971/non-trivial-designated-initializers-not-supported

This PR fixes it.


Maratyszcza:
Fixed in #417

syoyo:
Awesome! Thanks!

PR #415, opened 2020-03-12
Fix building operator-size-test and subgraph-size-test using cmake.

syoyo:
Support building `operator-size-test` and `subgraph-size-test`  using cmake.

But running these tests will crash due to the variable assignment to null pointer address: 

https://github.com/google/XNNPACK/blob/master/test/subgraph-size.c#L23

https://github.com/google/XNNPACK/blob/b7dd29e2b63c699e62a3955dbc1d3db1cb204c13/src/subgraph.c#L53

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1631,17 +1631,19 @@ IF(XNNPACK_BUILD_TESTS)
 
   # ---[ Build size tests
 
-  ADD_EXECUTABLE(operator-size-test test/size.c)
+  ADD_EXECUTABLE(operator-size-test test/operator-size.c)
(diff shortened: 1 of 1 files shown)


Maratyszcza:
Thank you for your contribution, @syoyo! You'd need to accept Google CLA (see above) before I can import this pull request.

syoyo:
@Maratyszcza Changes are trivial, so I don't sign SLA. You can cherry pick or copy&pase my changes and close this PR. Same applies for https://github.com/google/XNNPACK/pull/414

Maratyszcza:
Fixed in #416

syoyo:
Thanks! :+1:

PR #768, opened 2020-06-26
Updated operator-strings.c

Shivam7-1:


Maratyszcza:
Why did you add `<stdlib.h>` header? This file doesn't seem to use anything declared in `stdlib.h`

Shivam7-1:
because <stdlib.h> this not available in code

PR #769, opened 2020-06-26
Updated neonfp16arith-x16.c

Shivam7-1:


Maratyszcza:
What is the motivation for this change?

PR #833, opened 2020-07-16
Update add-nd.cc

Shivam7-1:


Maratyszcza:
The change is not helpful for the project.

Issue #1091, opened 2020-10-22
Build is broken on Ubuntu Xenial with CMake 3.5 and GCC 5.4


peterjc123:
Reproducible steps:
```bash
docker run --it --name xnnpack_test ubuntu:16.04
apt install -y cmake git build-essential
# cmake --version shows 3.5.1
# gcc --version shows 5.4.0

git clone https://github.com/google/XNNPACK.git
cd XNNPACK
mkdir build
cd build
cmake -DXNNPACK_BUILD_TESTS=OFF -DXNNPACK_BUILD_BENCHMARKS=OFF ..
make -j8
```

Logs:
```
[  4%] Building C object CMakeFiles/XNNPACK.dir/src/operators/convolution-nhwc.c.o
cc: error: ../src/operators/average-pooling-nhwc.c../src/operators/average-pooling-nhwc.cNOT:../src/operators/average-pooling-nhwc.cCONFIG:Debug: No such file or directory
cc: error: CMakeFiles/XNNPACK.dir/src/operators/average-pooling-nhwc.c.o: No such file or directory
[  4%] Building C object CMakeFiles/XNNPACK.dir/src/operators/convolution-nchw.c.o
CMakeFiles/XNNPACK.dir/build.make:86: recipe for target 'CMakeFiles/XNNPACK.dir/src/operators/average-pooling-nhwc.c.o' failed
make[2]: *** [CMakeFiles/XNNPACK.dir/src/operators/average-pooling-nhwc.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
cc: error: ../src/operators/binary-elementwise-nd.c../src/operators/binary-elementwise-nd.cNOT:../src/operators/binary-elementwise-nd.cCONFIG:Debug: No such file or directory
cc: error: CMakeFiles/XNNPACK.dir/src/operators/binary-elementwise-nd.c.o: No such file or directory
cc: error: ../src/operators/constant-pad-nd.c../src/operators/constant-pad-nd.cNOT:../src/operators/constant-pad-nd.cCONFIG:Debug: No such file or directory
cc: error: CMakeFiles/XNNPACK.dir/src/operators/constant-pad-nd.c.o: No such file or directory
CMakeFiles/XNNPACK.dir/build.make:158: recipe for target 'CMakeFiles/XNNPACK.dir/src/operators/constant-pad-nd.c.o' failed
[...]

PR #1093, opened 2020-10-22
Fixes build for CMake < 3.8

peterjc123:
Fixes https://github.com/google/XNNPACK/issues/1091.


peterjc123:
cc @Maratyszcza

Maratyszcza:
Thanks for the fix!

PR #1252, opened 2020-12-18
Exclude aarch32 & aarch64 ASM code accroding to cmake flag

skyline75489:


skyline75489:
@googlebot I signed it!

PR #1257, opened 2020-12-24
Improve architecture detection for Apple devices

skyline75489:
Relaxed several restrictions to support arm64e.


skyline75489:
Ideally in the future we should also drop support for armv7s and support only arm64 (iPhone 5s and later). Both armv7k and arm64_32 does not seem to get along with neon fma code in the project so watchOS will still be a no go.

Maratyszcza:
Relaxing regex to support arm64e LGTM, but deprecating support for older architectures is not. If XNNPACK fails to build for old architectures, please file a separate issue.

skyline75489:
@Maratyszcza Thanks for the feedback. I guess we can agree to disagree. I've retargeted the PR to just relax the regex for arm64e.

PR #1265, opened 2021-01-15
WIP: update the ruy benchmark

bjacob:
to use caching of weights and use the same ordering of the matmul as in other xnnpack benchmarks.

I can't test this easily for now but once the Ruy CMakeLists.txt are merged, that will become far easier.


Maratyszcza:
You'd likely want to update `bench/qs8-gemm.cc` and `bench/qu8-gemm.cc` as well

PR #1308, opened 2021-02-22
last_input should be changed only when it induces recalculation of

kimishpatel:
indirection buffer.

This bug exhibited as a crash in pytorch integration of xnnpack.

Ran scripts/build-local.sh. Ran ./build/local/convolution-nhwc-test on mac.


kimishpatel:
Looks like @Maratyszcza already fix this. Closing this PR.

PR #1310, opened 2021-02-22
last_input should be only changed during indices recalculation

malfet:
This bug was discovered while investigating PyTorch crash in https://github.com/pytorch/pytorch/issues/52463


Maratyszcza:
Already landing in #1309

malfet:
Abandoning in favor of https://github.com/google/XNNPACK/pull/1309

PR #1329, opened 2021-03-03
Add Bazel config_settings for the official Emscripten toolchain

mattsoulanille:
Current Bazel config settings for Emscripten are specific to TensorFlow.js's custom Emscripten toolchain. This PR adds config_settings for [the official toolchain](https://github.com/emscripten-core/emsdk/tree/master/bazel).

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -7413,13 +7413,13 @@ config_setting(
 
 config_setting(
     name = "emscripten",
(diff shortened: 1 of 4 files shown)


mattsoulanille:
I considered using skylib's `config_setting_group` instead of adding an extra config to all rules, but decided against that since it would introduce another dependency that consuming projects would need to include.

mattsoulanille:
Creating this internally since copybara needs to be updated.

PR #1367, opened 2021-03-25
Update FP16 dependency.

JerryShih:
Update the cmake download path for fp16.


Maratyszcza (approved):
For consistency, please update the path in WORKSPACE file as well

JerryShih:
> For consistency, please update the path in WORKSPACE file as well

@Maratyszcza 
I already have updated the WORKSPACE for bazel setting in this pr.

Issue #1402, opened 2021-04-12
Does iPhoneOS armv7 support neon dot and neon v8?


larryliu0820:
In CMakeLists.txt we have:
```
IF(CMAKE_SYSTEM_PROCESSOR MATCHES "^armv[5-8]" OR IOS_ARCH MATCHES "^armv7")
...
  SET_PROPERTY(SOURCE ${XNNPACK_NEONV8_MICROKERNEL_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS " -march=armv8-a -mfpu=neon-fp-armv8 ")
  SET_PROPERTY(SOURCE ${XNNPACK_NEONDOT_MICROKERNEL_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS " -march=armv8.2-a+dotprod -mfpu=neon-fp-armv8 ")
...
ENDIF()
```
In pytorch we are not able to build this on iphoneos armv7 because in arm_neon.h (clang 10.0.0) it fails these checks:
```
#if __ARM_ARCH >= 8 && defined(__ARM_FEATURE_DIRECTED_ROUNDING)
...
#if defined(__ARM_FEATURE_DOTPROD)
```
Is it possible for us to not include neon_dot and neon_v8 for iphoneos armv7? Or can we have a macro to exclude those two features?

Maratyszcza:
I imagine the `neon_dot` build error is the same as tensorflow/tensorflow#48464, but what is failing in `neon_v8` microkernels? It is plausible that Apple toolchain for ARMv7 doesn't support `neon_dot` because the only Apple SoCs that support it are ARM64-only, but IMO `neon_v8` should still compile fine, at there are AArch32-compatible Apple SoCs with ARMv8.

Maratyszcza:
Currently, there is no way to exclude `neon_dot` and `neon_v8` microkernels from the build. I'm looking into a possible solution.

larryliu0820:
Are armv8 SoCs passing CMakeLists.txt check: `IOS_ARCH MATCHES "^armv7"` ?

Maratyszcza:
I'd expect so. `"armv7"` really means AArch32 with ARMv7 or higher instruction set.

larryliu0820:
> "armv7" really means AArch32 with ARMv7 or higher instruction set.

Somehow armv7 is passing this check
```
if (cpuinfo_has_arm_neon_v8())
```
in init.c, but eventually it fails arm_neon.h:
```
[...]

PR #1406, opened 2021-04-14
Fix neon v8 for ios armv7

larryliu0820:
Based on discussion in #1402


larryliu0820:
@googlebot I signed it!

PR #1462, opened 2021-05-14
Workaround -Wincompatible-pointer-types

georgthegreat:
The first operand of `InterlockedCompareExchangePointer` has type `void* volatile* Destination`, thus triggering the warning under clang-cl.

This PR adds explicit type casting to workaround this.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -3432,7 +3432,7 @@ enum xnn_status xnn_initialize(const struct xnn_allocator* allocator) {
     allocator = &xnn_default_allocator;
   }
   #ifdef _MSC_VER
-    _InterlockedCompareExchangePointer(&init_allocator, allocator, NULL);
+    _InterlockedCompareExchangePointer((void* volatile)(&init_allocator), allocator, NULL);
   #else
     __sync_bool_compare_and_swap(&init_allocator, NULL, allocator);
   #endif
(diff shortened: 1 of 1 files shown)


src/init.c
```diff
@@ -3432,7 +3432,7 @@ enum xnn_status xnn_initialize(const struct xnn_allocator* allocator) {
     allocator = &xnn_default_allocator;
   }
   #ifdef _MSC_VER
-    _InterlockedCompareExchangePointer(&init_allocator, allocator, NULL);
+    _InterlockedCompareExchangePointer((void* volatile)(&init_allocator), allocator, NULL);
```
Maratyszcza, line 3435: Just `(void* volatile) &init_allocator`, no need for extra brackets.

Maratyszcza (changes requested):

src/init.c
Maratyszcza: Also, [the argument in the function signature](https://docs.microsoft.com/en-us/cpp/intrinsics/interlockedcompareexchangepointer-intrinsic-functions?view=msvc-160) has `void* volatile*` type

georgthegreat: Should be fixed now

Maratyszcza (approved):

Issue #1612, opened 2021-07-02
There is no function for mod operation in include file of xnnpack.h


carrycooldude:
For addition we have 
> enum xnn_status xnn_create_add_nd_f32(
  float output_min,
  float output_max,
  uint32_t flags,
  xnn_operator_t* add_op_out);

>enum xnn_status xnn_setup_add_nd_f32(
  xnn_operator_t add_op,
  size_t num_input1_dims,
  const size_t* input1_shape,
  size_t num_input2_dims,
  const size_t* input2_shape,
  const float* input1,
  const float* input2,
  float* output,
  pthreadpool_t threadpool);

[This is the link for it](https://github.com/google/XNNPACK/blob/master/include/xnnpack.h)

Maratyszcza:
Right, `MOD` operator is not supported, and there are no short-term plans to support it.

PR #1613, opened 2021-07-02
 Added create and setup function for mod kernel

carrycooldude:
> I made a P.R for mod kernel in the [include file](https://github.com/google/XNNPACK/blob/master/include/xnnpack.h) for which I create an issue[ #1612](https://github.com/google/XNNPACK/issues/1612)

### DIFF
--- a/include/xnnpack.h
+++ b/include/xnnpack.h
@@ -829,7 +829,7 @@ enum xnn_status xnn_define_minimum2(
   uint32_t output_id,
   uint32_t flags);
 
-  /// Define a 2-Input Node Mod and add it to a Subgraph.
+  /// Define a 2-Input Node Mod and mod it to a Subgraph.
   ///
   /// The 2-Input Node for Mod computes elementwise minimum of two tensor inputs with numpy broadcasting rules.
   ///
--- a/include/xnnpack.h
+++ b/include/xnnpack.h
@@ -829,6 +829,29 @@ enum xnn_status xnn_define_minimum2(
   uint32_t output_id,
   uint32_t flags);
 
+  /// Define a 2-Input Node Mod and add it to a Subgraph.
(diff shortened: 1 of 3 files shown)


carrycooldude:
cc: @huningxin

Maratyszcza (changes requested):
Missing implementation

carrycooldude:
@Maratyszcza I updated it in the include folder of xnnpack.h file

carrycooldude (commented):
I hope so I update the xnnpack.h file let me know what I am doing wrong

Maratyszcza:
I see the declaration of the new functions in `xnnpack.h`, but the implementation of the functions is missing. In general, I suggest to start with microkernels for the new operators, and only add operator-level APIs once you have microkernels for a full set of platforms supported by XNNPACK (i.e. x86, ARM, and WebAssembly).

carrycooldude (commented):
Have a look over this.

carrycooldude:
cc : @Maratyszcza

Maratyszcza:
The code defines new functionality, but doesn't implement it. It is missing the implementation of the new functions declared in `xnnpack.h`.

carrycooldude:
> The code defines new functionality, but doesn't implement it. It is missing the implementation of the new functions declared in `xnnpack.h`.

Actually The main implementation is in the tfjs kernel and for implementation I implemented on form of test suite 
And after this I am wrong let me know where I have to implement the function

Maratyszcza:
If you implement this function in TF.js, you don't need it in XNNPACK. TF.js WAsm backend uses XNNPACK for perf-critical operators, not the other way around.

carrycooldude:
I tired in that but test fails so I think it's better to do in XNNPACK
directly

PR #1619, opened 2021-07-08
Add patch for MSVC intrinsics

mitchellspryn:
When building XNNPACK with MSVC, I get warnings about undefined intrinsic functions, which eventually turn into link errors:

```
up32x4-minmax-avx512f.c(153): warning C4013: '_cvtu32_mask16' undefined; assuming extern returning int

...

lld-link: error: undefined symbol: _cvtu32_mask16
```

Adding this patch properly defines these symbols and causes the warnings to go away.


mitchellspryn:
@googlebot I signed it!

mitchellspryn:
@googlebot I fixed it.

PR #1621, opened 2021-07-08
Polyfill missing AVX512 intrinsics on MSVC

mitchellspryn:
When building XNNPACK with MSVC, I got many warnings about missing intrinsics that turned into link errors:

```
f32-dwconv\gen\up32x4-minmax-avx512f.c(153): warning C4013: '_cvtu32_mask16' undefined; assuming extern returning int

...

lld-link: error: undefined symbol: _cvtu32_mask16
```

Adding this patch to correctly define the polyfill for MSVC solved this error.

### DIFF
--- a/src/xnnpack/intrinsics-polyfill.h
+++ b/src/xnnpack/intrinsics-polyfill.h
@@ -33,7 +33,7 @@ void _mm_storeu_si32(const void* address, __m128i v) {
 #ifdef __AVX512F__
 #include <immintrin.h>
 
-// GCC pre-7, Clang pre-8, Android NDK Clang pre-8.0.7, Apple Clang pre-11, and ICC pre-18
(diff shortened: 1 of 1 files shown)


Maratyszcza:
Are you using the latest version of MSVC?

Maratyszcza (changes requested):
Update comment as well

mitchellspryn:
Not quite. I'm currently using VS2017 (version 15.9.12). Is using the latest a requirement for building XNNPACK?

Maratyszcza:
If the later MSVC versions include these intrinsics, you need to check `_MSC_VER` value and only polyfill for versions which don't lack the intrinsics.

mitchellspryn:
I tried compiling this bit of code on a few different versions of visual studio:

```
#include <immintrin.h>

int main()
{
    __mmask16 i = _cvtu32_mask16(0);
    return 0;
}
```

It seems to compile fine on VS2019, and does not compile on versions prior (e.g. 2017). Accordingly, I added the version check to polyfill if the MSVC version is pre-2019.

PR #1631, opened 2021-07-14
Fix missing braces around initializer warning

malfet:
As `xnn_hmp_igemm_ukernel` is a struct containing array, it should be initialized as `{{element}}`, otherwise compiler warns about missing braces, as shown in  https://godbolt.org/z/77dsc88T4


Maratyszcza (approved):
LGTM

PR #2673, opened 2022-02-23
Add GitHub actions workflow to build using CMake

ngzhian:


ngzhian:
@Maratyszcza seems like the import might be stuck because of a new folder? I'll try creating this then export it to a PR vai copybara.

ngzhian:
Closed in favor of #2674.

Issue #2767, opened 2022-03-22
Build fail on native aarch64.


cbalint13:
Latest master fail to link references on native aarch64 targets.

```
xnnpack/src/init.c:2432: undefined reference to `xnn_x8_transposec_ukernel__16x16_reuse_dec_zip_neon'
xnnpack/src/init.c:2432: undefined reference to `xnn_x8_transposec_ukernel__16x16_reuse_dec_zip_neon'
xnnpack/src/init.c:2442: undefined reference to `xnn_x16_transposec_ukernel__8x8_reuse_dec_zip_neon'
xnnpack/src/init.c:2442: undefined reference to `xnn_x16_transposec_ukernel__8x8_reuse_dec_zip_neon'
ld: libXNNPACK.so.1.0: internal symbol `xnn_x16_transposec_ukernel__8x8_reuse_dec_zip_neon' isn't defined
ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/XNNPACK.dir/build.make:9103: libXNNPACK.so.1.0] Error 1
make[1]: *** [CMakeFiles/Makefile2:195: CMakeFiles/XNNPACK.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

* See [builder-live.log.gz](https://github.com/google/XNNPACK/files/8322439/builder-live.log.gz) for full environment logs and compile errors.
* The linkage error occurs across various [builds](https://copr.fedorainfracloud.org/coprs/rezso/ML/build/3811635) , all aarch64 natives.
* Prior to some 20220309 git ```master``` dates [all builds](https://copr.fedorainfracloud.org/coprs/rezso/ML/package/xnnpack/) are fine.

Thank you !
~cristian.

cbalint13:
A simple fix would be to include ```ALL_MICROKERNELS_SRC``` into the library:

```
 IF(XNNPACK_LIBRARY_TYPE STREQUAL "default")
-  ADD_LIBRARY(XNNPACK ${COLD_SRCS} ${HOT_SRCS} ${TABLE_SRCS} ${JIT_SRCS} ${PROD_MICROKERNEL_SRCS})
+  ADD_LIBRARY(XNNPACK ${COLD_SRCS} ${HOT_SRCS} ${TABLE_SRCS} ${JIT_SRCS} ${ALL_MICROKERNEL_SRCS})
[...]

PR #2789, opened 2022-03-25
Fix missing functions on native aarch64 build.

cbalint13:
Fix for #2767 issue.

The current PR builds fine on various [copr builders](https://copr.fedorainfracloud.org/coprs/rezso/ML/build/3840910) for ```aarch64```.
My guess is that ```PROD``` means **production**, so we promote here missing kernels to ```PROD```.

Thank you,
~cristian.

Cc @alankelly


alankelly:
Thanks for opening this PR. Our commits crossed paths so this isn't needed anymore. Everything should be working again now.

PR #3435, opened 2022-09-04
CMake: Enabled build for FreeBSD

stemann:
Cf. https://github.com/JuliaPackaging/Yggdrasil/pull/4577


Maratyszcza:
XNNPack can't be used on FreeBSD because `cpuinfo` dependency doesn't support FreeBSD

stemann:
OK, this PR makes little sense then :-)

PR #3616, opened 2022-10-08
fix disallowed option when built by Xcode

wejoncy:
**Target**:
To support build XNNPACK by Xcode

Building XNNPACK on Mac with CMake and set XCode as Generator simultaneously is not able to be done currently.
**Reason**:
According to https://cmake.org/cmake/help/latest/prop_sf/COMPILE_FLAGS.html
[Xcode](https://cmake.org/cmake/help/latest/generator/Xcode.html#generator:Xcode) does not support per-config per-source settings, so expressions that depend on the build configuration are not allowed with that generator.

**Solution**
We just disabled that option when Xcode is used.


Maratyszcza:
Does the build work with these changes? ISA-specific microkernels require proper COMPILE_FLAGS, I expect them to fail to build without it.

wejoncy:
> Does the build work with these changes? ISA-specific microkernels require proper COMPILE_FLAGS, I expect them to fail to build without it.

Thanks for your review.  Those ISA-specific kernels do fail without the hint of flag so I have to disable those features. But I found we only needs to suppress this flag `"$<$<NOT:$<CONFIG:Debug>>: -O2 >"`, while keeping `FP16/BF16/Dot` stuff.

CMakeLists.txt
```diff
@@ -70,6 +70,11 @@ IF(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   ADD_COMPILE_OPTIONS("-Wno-psabi")
 ENDIF()
 
+# Xcode does not support COMPILE_FLAGS according to https://cmake.org/cmake/help/latest/prop_sf/COMPILE_FLAGS.html
```
Maratyszcza, line 73: This is not what the linked document says: "However, Xcode does not support per-config per-source settings, so expressions that depend on the build configuration are not allowed with that generator." So, it is a combination of per-source-file and per-config settings that is not supported.

CMakeLists.txt
```diff
@@ -70,6 +70,11 @@ IF(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
...
+IF(CMAKE_GENERATOR STREQUAL Xcode)
+  set(XCODE ON)
+ENDIF()
+
```
Maratyszcza, line 77: Don't define `XCODE` variable, it may lead to name collisions. Just use `IF(CMAKE_GENERATOR STREQUAL Xcode)` instead of of `IF(XCODE)`

CMakeLists.txt
```diff
@@ -7402,7 +7407,7 @@ IF(MSVC)
   SET_PROPERTY(SOURCE ${ALL_MICROKERNEL_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS "$<$<NOT:$<CONFIG:Debug>>: /O2 >")
   SET_PROPERTY(SOURCE ${HOT_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS "$<$<NOT:$<CONFIG:Debug>>: /O2 >")
   SET_PROPERTY(SOURCE ${COLD_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS "$<$<NOT:$<CONFIG:Debug>>: /O1 >")
-ELSE()
+ELSEIF(NOT XCODE)
```
Maratyszcza, line 7410: With this change, optimization settings are not set at all for the compiled files, resulting in subpar performance. A proper solution would be to refactor CMakeLists so that the per-config flags can be set at a target level.

CMakeLists.txt
wejoncy: I got it, It actually means it doesn't support Generator expression collaborating with Compile_flag, it will lead  to per-config per-source settings

wejoncy: Fixed

CMakeLists.txt
wejoncy: Refactored

CMakeLists.txt
wejoncy: Done.

Using CMAKE_BUILD_TYPE to check its buid-type and decide if we should add those optimization flags.

CMakeLists.txt
```diff
@@ -8163,7 +8168,7 @@ IF(XNNPACK_BUILD_TESTS)
 
   ADD_EXECUTABLE(f16-dwconv-minmax-test test/f16-dwconv-minmax.cc)
   TARGET_INCLUDE_DIRECTORIES(f16-dwconv-minmax-test PRIVATE include src test)
-  TARGET_LINK_LIBRARIES(f16-dwconv-minmax-test PRIVATE cpuinfo fp16 pthreadpool gtest gtest_main)
+  TARGET_LINK_LIBRARIES(f16-dwconv-minmax-test PRIVATE cpuinfo fp16 pthreadpool gtest gtest_main logging)
```
Maratyszcza, line 8171: This line is for external dependencies. Internal dependencies are listed in alphabetical order on the line below. Ditto for other TARGET_LINK_LIBRARIES changes.

CMakeLists.txt
```diff
@@ -7402,10 +7403,14 @@ IF(MSVC)
   SET_PROPERTY(SOURCE ${ALL_MICROKERNEL_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS "$<$<NOT:$<CONFIG:Debug>>: /O2 >")
   SET_PROPERTY(SOURCE ${HOT_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS "$<$<NOT:$<CONFIG:Debug>>: /O2 >")
   SET_PROPERTY(SOURCE ${COLD_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS "$<$<NOT:$<CONFIG:Debug>>: /O1 >")
-ELSE()
-  SET_PROPERTY(SOURCE ${ALL_MICROKERNEL_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS "$<$<NOT:$<CONFIG:Debug>>: -O2 >")
-  SET_PROPERTY(SOURCE ${HOT_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS "$<$<NOT:$<CONFIG:Debug>>: -O2 >")
-  SET_PROPERTY(SOURCE ${COLD_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS "$<$<NOT:$<CONFIG:Debug>>: -Os >")
+ELSEIF(NOT (CMAKE_BUILD_TYPE AND (CMAKE_BUILD_TYPE STREQUAL "Debug")))
+  # Compataible with Xcode, which does not support per-config per-source settings,
```
Maratyszcza, line 7407: We can't set `"$<$<NOT:$<CONFIG:Debug>>: -O2 >"` `COMPILE_FLAGS` on source files with XCode, but we can do it on targets. So you need to set `COMPILE_FLAGS` for targets rather than source files, then it will work with XCode. Trace `ALL_MICROKERNELS_SRCS`, `HOT_SRCS`, `COLD_SRCS` to targets and set `COMPILE_FLAGS` for those targets accordingly.

scripts/build-ios-arm64.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
```
Maratyszcza, line 1: This change is unrelated to the PR, please leave it out

scripts/build-ios-arm64.sh
```diff
@@ -0,0 +1,63 @@
...
+  echo "Have you installed Xcode?"
+  exit 1
+fi
+
+mkdir -p build/iOS/arm64
+
+# Create Tool-chain file
+IOS_TOOL_CHAIN=build/iOS/arm64/ios_tmp.toolchain.cmake
+cat << EOF > ${IOS_TOOL_CHAIN}
+set(CMAKE_SYSTEM_NAME iOS)
```
Maratyszcza, line 24: If you need a toolchain file, create on in the top-level `cmake` subdirectory.

CMakeLists.txt
wejoncy: Done

scripts/build-ios-arm64.sh
wejoncy: Got it

scripts/build-ios-arm64.sh
wejoncy: removed.

CMakeLists.txt
wejoncy: TARGET_COMPILE_OPTIONS seems work. but not quite sure it 's the best way to do that.

wejoncy:
@Maratyszcza Could you please take a look again?

Maratyszcza:
This change is implemented in #3732. It is sufficient to make CMake successfully generate Xcode project, but not enough to make the Xcode project successfully build XNNPACK.

Maratyszcza:
Xcode project not successfully builds for ARM64/x86_64 macOS, ARM64 iOS, and x86_64 iOS Simulator. See the corresponding GitHub Actions.

wejoncy:
> Xcode project now successfully builds for ARM64/x86_64 macOS, ARM64 iOS, and x86_64 iOS Simulator. See the corresponding GitHub Actions.

Perfect!!! Thanks

PR #4289, opened 2023-02-13
Use #include <> instead of ""

kirklandsign:
Some compilers have different behavior and may include different files, so "#pragma once" does not protect from double inclusion. So far this could break some Windows build for PyTorch.

All other XNNPACK code is using #include <> so let's make it universal.


kirklandsign:
> Can you provide a scenario where this breaks a build?

For example in https://github.com/google/XNNPACK/blob/master/src/cache.c#L6 we have `#include "xnnpack/cache.h"` and in https://github.com/google/XNNPACK/blob/master/src/xnnpack/cache.h#L12 we have `#include <xnnpack/mutex.h>     // For xnn_mutex.
`

Then in https://github.com/google/XNNPACK/blob/master/src/cache.c#L16 we have `#include "xnnpack/mutex.h"
` again.

This should not cause any problem, because we have `#pragma once`.

However, when we have a windows build, when it's building cache.c I got this error:
```
stderr: D:\***\XNNPACK\src\xnnpack\mutex.h(26): error C2011: 'xnn_mutex': 'struct' type redefinition
D:\***\XNNPACK#default,private-headers\xnnpack\mutex.h(26): note: see declaration of 'xnn_mutex'
    When running <c preprocess_and_compile>.
    When building rule //***/XNNPACK:XNNPACK#compile-cache.c.o82d6894a,default (***/windows:x86_64-ms14.28.29910-dev-llvm-lld-dynamic-vs16.9.3).
Command failed with exit code 2.
```

PR #4329, opened 2023-02-22
Add darwin_x86_64 CPU handling

keith:
Currently bazel supports `darwin` and `darwin_x86_64` as meaning the same thing. The fully qualified version is normally used if you're cross compiling from M1 machines to intel machines. I'm also hoping to remove the unqualified version to reduce confusion. This change makes it compatible with both for now.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -11312,10 +11312,18 @@ config_setting(
 )
 
 config_setting(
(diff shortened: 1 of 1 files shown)


keith:
@fbarchard @alankelly could one of you take a look??

PR #4705, opened 2023-04-23
Add a data check of slope for prelu operation

### LINKED ISSUE #4692: An issue about the implementation of prelu operation
As I observed, XNNPACK checks the type and the data of the constant operand in the implementation of ops, such as the filter of  [conv2d](https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:third_party/xnnpack/src/src/subgraph/convolution-2d.c;l=600). 
But I also found that XNNPACK doesn't check whether slope is a constant before using its buffer in the implementation of [PRelu](https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:third_party/xnnpack/src/src/subgraph/prelu.c;l=165;drc=8ce14ef97f8607b1b57f8d02da575ed5150eea9e). It will cause a crash due to null pointer in some scenarios.

lisa0314:
Fix #4692.

@Maratyszcza  PTAL, thanks!

### DIFF
--- a/src/subgraph/prelu.c
+++ b/src/subgraph/prelu.c
@@ -165,6 +165,13 @@ enum xnn_status xnn_define_prelu(
     return xnn_status_invalid_parameter;
   }
 
+  if (slope_value->data == NULL) {
+    xnn_log_error(
+      "failed to define %s operator with slope ID #%" PRIu32 ": non-static Value",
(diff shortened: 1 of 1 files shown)


Maratyszcza:
Thank you for your contribution!

PR #5263, opened 2023-08-16
Fix typo in convolution-nhwc.c

bhbruce:


Maratyszcza:
Thank you for the fix. Committed in f8630341eb8aebdfe45a1ac8522bef80f0a133e8

PR #5491, opened 2023-09-18
add f16 square bench

chenkui164:
add f16 square bench


Maratyszcza (changes requested):
Replicate the `CMakeLists.txt` changes in `BUILD.bazel`

chenkui164:
It seems that there is no need to modify BUILD.bazel, square_bench already includes the existing dependencies for fp16.
I have successfully compiled it on my own computer.

PR #5739, opened 2023-11-10
[Request for Comments] Minimal Proof-of-concept of Dynamic Shape Support for xnn_runtime

digantdesai:
_This submission is not meant for an in-depth code review or for integration, but rather to initiate a discussion on the topic of dynamic shape support._

Based on this, I am looking forward to discussing:

1. The design for the runtime dynamic shape support in general.
    - The runtime and op->reshape() modifications for this feature, particularly if we plan to extend this to other operators and its implications for backwards compatibility.
1. Progress and concurrent development with the "Shape Inference" work for xnn_subgraph.
    - Support for upper bounded tensors by only specifying the maximum dimension size value, and with a default minimum value (= 1). This contrasts with calculating both the minimum and maximum for each tensor through shape inference.

### DIFF
--- a/include/xnnpack.h
+++ b/include/xnnpack.h
@@ -1826,6 +1826,8 @@ enum xnn_status xnn_create_runtime(
 struct xnn_external_value {
   uint32_t id;
   void* data;
+  size_t num_dims;
+  const size_t* dim;
 };
 
 /// Setup data pointers for external inputs and outputs in a Runtime object.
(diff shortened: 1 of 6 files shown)


include/xnnpack.h
```diff
@@ -1826,6 +1826,8 @@ enum xnn_status xnn_create_runtime(
 struct xnn_external_value {
   uint32_t id;
   void* data;
+  size_t num_dims;
+  const size_t* dim;
```
mcr229, line 1830: From API point of view, these are the only changes that would be noticed from building the subgraph? Even then, it would be fine to leave `num_dims` and `dim` to be empty/null?

PR #5869, opened 2023-12-15
RVV F32-Softmax Patch 1: Add RVV reduced min & max & minmax

bhbruce:
This PR is to add RVV version of reduced min & max & minmax.
We not only add RVV ukernels but also modify test-generator to enable `vector_tile`.
The `vector_tile` idea has been applied to `tools/generate-vunary-test.py` and we expand it to `tools/generate-reduce-test.py`.

c.c @phoebesv @rednoah91


bhbruce:
@fbarchard @alankelly If you have free time, please help to review this.

PR #5890, opened 2023-12-27
Replace riscv64 vector GCC CI with clang toolchain.

phoebesv:
In this PR, we replaced the gcc toolchain in the rvv CI with clang. For RVV development, Clang is a better option than GCC as it offers more comprehensive support for RVV, and the upstream clang toolchain has more recent intrinsic version support than GCC.
Since the target CPU architecture "riscv64" is not supported by the current version of cpuinfo, we enable the compilation definition GNU_SOURCE in rvv ci to avoid implicit function declarations in cpuinfo. The option USE_GNU_SOURCE can be removed after upgrading cpuinfo to include the target CPU architecture with riscv64 support.


phoebesv:
@alankelly @fbarchard Please help to review this if you have time, thanks.

.github/workflows/build.yml
```diff
@@ -85,7 +85,7 @@ jobs:
       - name: Run tests
         run: ctest --output-on-failure --parallel $(nproc)
         working-directory: ${{ github.workspace }}/build/linux/riscv64
-  cmake-linux-riscv64-rvv:
+  cmake-linux-riscv64-vector:
```
NonerKao, line 88: I wonder why the suffix need changing. The summary didn't mention this as well.

phoebesv: Because riscv64 already implies rv, the term rvv has been changed to a vector to make it more straightforward. The description has been updated to reflect these changes. Many thanks.

phoebesv: The suffix has been changed back.

fbarchard (approved):

PR #5910, opened 2024-01-10
[Dynamic Shapes] Fully-connected op

digantdesai:
Changes for dynamic shape support:
* Add output resize support for fully-connected op
* Remove shape_inference methods, no change in the subgraph file yet
* Add support for `XNN_FLAG_TENSORFLOW_RESHAPE_2D`. Not sure if this is used, so haven't added any tests
  
Other changes:
* Some of the QD tests were not setting the output rank correctly, fix that
* There is a flaky test - not sure if this fix is sufficient but should be in the right direction

### DIFF
--- a/src/subgraph/fully-connected.c
+++ b/src/subgraph/fully-connected.c
@@ -577,47 +577,6 @@ static enum xnn_status setup_fully_connected_operator(
   }
 }
 
-static enum xnn_shape_inference_status infer_shape_forward(
-  const struct xnn_node* node,
-  struct xnn_value* values)
-{
-  // Assert that filter tensor has static shape.
-  const uint32_t filter_id = node->inputs[1];
-  const struct xnn_value* filter = &values[filter_id];
-  assert(xnn_tensor_shape_is_static(filter));
-
-  enum xnn_shape_inference_status status = xnn_shape_inference_status_no_change;
-
-  // Infer output channels.
-  const uint32_t output_channel_index = (node->flags & XNN_FLAG_TRANSPOSE_WEIGHTS) ? 1 : 0;
-  const uint32_t output_id = node->outputs[0];
-  struct xnn_value* output = &values[output_id];
-  status = xnn_tensor_propagate_dimension(output, output->shape.num_dims - 1, filter->shape.dim[output_channel_index]);
-  if (status == xnn_shape_inference_status_error) {
-    return status;
-  }
-
-  if (node->flags & XNN_FLAG_TENSORFLOW_RESHAPE_2D) {
-    // No inference for input/output shape possible.
-    return status;
-  }
-
-  // Propagate input shape to output.
-  const struct xnn_value* input = &values[node->inputs[0]];
-  for (size_t cur_dim = 0; cur_dim < input->shape.num_dims - 1; cur_dim++) {
-    const enum xnn_shape_inference_status changed = xnn_tensor_propagate_dimension(output, cur_dim, input->shape.dim[cur_dim]);
-    if (changed == xnn_shape_inference_status_error) {
-      return changed;
-    } else if (changed == xnn_shape_inference_status_changed) {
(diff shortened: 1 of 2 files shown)


src/subgraph/fully-connected.c
```diff
@@ -322,6 +322,62 @@ static enum xnn_status create_fully_connected_operator(
...
+  const struct xnn_value* filter = &values[filter_id];
+
+  const uint32_t output_id = opdata->outputs[0];
+  struct xnn_value* output = (struct xnn_value*) &values[output_id];
+
+  const uint32_t input_id = opdata->inputs[0];
+  const struct xnn_value* input = &values[input_id];
+
+  // Infer output channels.
+  const uint32_t output_channel_index = (opdata->flags & XNN_FLAG_TRANSPOSE_WEIGHTS) ? 1 : 0;
```
alankelly, line 342: output channel index is independent of weights transpose and is always 1.

src/subgraph/fully-connected.c
```diff
@@ -322,6 +322,62 @@ static enum xnn_status create_fully_connected_operator(
...
+  enum xnn_shape_inference_status status =
+    xnn_tensor_propagate_dimension(output, output->shape.num_dims - 1, filter->shape.dim[output_channel_index]);
```
alankelly, line 344: Make this /*to_dim=*/ 1 and move the assert from line 352 before this

src/subgraph/fully-connected.c
```diff
@@ -322,6 +322,62 @@ static enum xnn_status create_fully_connected_operator(
...
+  const size_t new_size = xnn_tensor_get_size(output);
+  if (new_size > output->size || old_workspace_size < opdata->workspace_size) {
+    output->size = new_size;
+    return xnn_status_reallocation_required;
+  }
+
+  return xnn_status_success;
+}
+
+
```
alankelly, line 380: remove 2nd new line

src/subgraph/fully-connected.c
alankelly, line 342: can you rename this something like filter_channel_index or output_filter_channel_index as this refers to the channel index of the filter and not the output tensor?

src/subgraph/fully-connected.c
```diff
@@ -577,88 +653,6 @@ static enum xnn_status setup_fully_connected_operator(
...
-      // Only overwrite status if inference on this dimension changed. We could have changed something above, and not
-      // changed anything here, then overwriting status to be no_change will be incorrect.
-      status = changed;
-    }
-  }
-
-  return status;
-}
-
-static enum xnn_shape_inference_status infer_shape_backward(
```
alankelly, line 621: Can we please keep infer_shape_backward? It has not been superseded by the update reshape function and may be useful in the future.

src/subgraph/fully-connected.c
```diff
@@ -1045,8 +1039,6 @@ enum xnn_status xnn_define_fully_connected(
   node->create = create_fully_connected_operator;
   node->reshape = reshape_fully_connected_operator;
   node->setup = setup_fully_connected_operator;
-  node->infer_shape_forward = infer_shape_forward;
-  node->infer_shape_backward = infer_shape_backward;
```
alankelly, line 1049: let's keep this line too.

src/subgraph/fully-connected.c
digantdesai: Renamed the variables

src/subgraph/fully-connected.c
digantdesai: Not sure I follow, IIUC output rank can be >2 here.

alankelly: True, output can have > 2 rank

alankelly (approved):

digantdesai:
@alankelly - I don't think I can merge this, can you please do it for me? Thanks.

alankelly:
It's in progress. Should land soon.

PR #5957, opened 2024-01-26
[subgraph][sdpa] Add dynamic shape support

digantdesai:
* Add output resize function.
* Add tests for no resize required and resize required similar to
  `matches_operator_api`.
    * Rationale to add such tests is to make sure shape change works
      correctly end-to-end with the intermediate buffers on the
      workspace.
    * Add a `Resize()` helper function to better manage of input tensors
      for user provided shapes and changes in them.


alankelly (approved):
Thanks for the great testing

PR #5997, opened 2024-02-06
Add subgraph support for FP16 tensors

GregoryComer:
As per previous design discussions, update subgraph node definition methods to allow for FP16 inputs. Add test coverage for FP16 support using subgraph API.

Tested building with cmake and running tests via cmake:
```
cmake --build . -j32 && ctest -j32 --output-on-failure
...
100% tests passed, 0 tests failed out of 382
```

This is a fairly large PR, so feel free to suggest that I break it up, if preferred for easier review. As an additional note, I did notice that there is a significant amount of boilerplate code in many of the tests, I've followed the example of existing FP32/Q8 tests, but would be happy to refactor to reduce code duplication, if desired.

Also, bazel build appears to be broken (local errors match error in CI bazel build), so I have not tested via bazel. If that issue is resolved, I can go ahead and verify that it builds and runs via bazel.

### DIFF
--- a/src/subgraph/convolution-2d.c
+++ b/src/subgraph/convolution-2d.c
@@ -597,6 +597,11 @@ static inline enum xnn_compute_type validate_datatypes_with_bias(
           output_datatype == xnn_datatype_fp32)
       {
         return xnn_compute_type_fp32;
+      } else if (input_datatype == xnn_datatype_fp16 &&
+          bias_datatype == xnn_datatype_fp32 &&
+          output_datatype == xnn_datatype_fp16) {
+        // Flag: XNN_FLAG_FP32_STATIC_WEIGHTS
+        return xnn_compute_type_fp16;
       }
       break;
     case xnn_datatype_qint8:
@@ -648,6 +653,9 @@ static inline enum xnn_compute_type validate_datatypes_without_bias(
     case xnn_datatype_fp32:
       if (input_datatype == xnn_datatype_fp32 && output_datatype == xnn_datatype_fp32) {
         return xnn_compute_type_fp32;
+      } else if (input_datatype == xnn_datatype_fp16 && output_datatype == xnn_datatype_fp16) {
+        // Flag: XNN_FLAG_FP32_STATIC_WEIGHTS
+        return xnn_compute_type_fp16;
       }
       break;
     case xnn_datatype_qint8:
@@ -792,6 +800,7 @@ enum xnn_status xnn_define_convolution_2d(
   }
 
   switch (input_value->datatype) {
+    case xnn_datatype_fp16:
     case xnn_datatype_fp32:
     case xnn_datatype_qint8:
     case xnn_datatype_quint8:
@@ -838,6 +847,7 @@ enum xnn_status xnn_define_convolution_2d(
 
   switch (filter_value->datatype) {
     case xnn_datatype_fp32:
+    case xnn_datatype_fp16:
       break;
     case xnn_datatype_qint8:
       if (filter_value->quantization.zero_point != 0) {
@@ -885,6 +895,7 @@ enum xnn_status xnn_define_convolution_2d(
     }
 
     switch (bias_value->datatype) {
+      case xnn_datatype_fp16:
       case xnn_datatype_fp32:
       case xnn_datatype_qint32:
       case xnn_datatype_qcint32:
@@ -918,9 +929,9 @@ enum xnn_status xnn_define_convolution_2d(
   }
 
   switch (output_value->datatype) {
+    case xnn_datatype_fp16:
     case xnn_datatype_fp32:
     case xnn_datatype_qint8:
-    case xnn_datatype_fp16:
     case xnn_datatype_quint8:
       break;
     default:
--- a/src/subgraph/fully-connected.c
+++ b/src/subgraph/fully-connected.c
@@ -709,6 +709,11 @@ static inline enum xnn_compute_type validate_datatypes_with_bias(
           output_datatype == xnn_datatype_fp32)
       {
         return xnn_compute_type_fp32;
+      } else if (input_datatype == xnn_datatype_fp16 &&
+          bias_datatype == xnn_datatype_fp32 &&
+          output_datatype == xnn_datatype_fp16) {
+        // Flag: XNN_FLAG_FP32_STATIC_WEIGHTS
+        return xnn_compute_type_fp16;
       }
       break;
     case xnn_datatype_qcint4:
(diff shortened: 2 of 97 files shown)


digantdesai:
This is awesome. Thanks @GregoryComer.

fbarchard (approved):

src/subgraph/convolution-2d.c
```diff
@@ -561,6 +561,11 @@ static inline enum xnn_compute_type validate_datatypes_with_bias(
           output_datatype == xnn_datatype_fp32)
       {
         return xnn_compute_type_fp32;
+      } else if (input_datatype == xnn_datatype_fp16 &&
+          bias_datatype == xnn_datatype_fp32 &&
```
alankelly, line 565: Are you certain that the bias in this case is fp32? Typically for fp16 inference all float32 are transformed to float16. The microkernels only support fp16 bias: https://github.com/google/XNNPACK/blob/6ef9832d025fb385d69c8fec88e0cb6ffa950e7f/src/f16-igemm/gen/f16-igemm-1x16-minmax-neonfp16arith-ld64.c#L48

The bias is loaded here.

alankelly: Ignore this, the bias and weights are converted in the packing function.

This is inefficient, we should optionally be able to pass fp16 weights and biases. Do you fancy adding this in a separate PR? I think only some small changes to this file are necessary, at the operator level fp16 weights and biases are handled.

alankelly (approved):

test/concatenate5.cc
```diff
@@ -14,6 +14,7 @@
 #include <random>
 #include <vector>
 
+#include <fp16.h>
```
alankelly, line 17: Please all test fp16 includes to #include <fp16/fp16.h>

test/static-resize-bilinear-2d.cc
```diff
@@ -338,6 +379,71 @@ TEST_F(StaticResizeBilinear2DTestQU8, matches_operator_api)
...
+TEST_F(StaticResizeBilinear2DTestF16, matches_operator_api)
+{
+  ASSERT_EQ(xnn_status_success, xnn_initialize(/*allocator=*/nullptr));
+  std::generate(input.begin(), input.end(), [&]() { return fp16_ieee_from_fp32_value(f32dist(rng)); });
+  std::fill(operator_output.begin(), operator_output.end(), fp16_ieee_from_fp32_value(std::nanf("")));
+  std::fill(subgraph_output.begin(), subgraph_output.end(), fp16_ieee_from_fp32_value(std::nanf("")));
+
+  // Call operator API.
+  xnn_operator_t op = nullptr;
+  const xnn_status status = xnn_create_resize_bilinear2d_nhwc_f16(channels, channels, channels, /*flags=*/0, &op);
```
alankelly, line 391: Please rebase, the signature of this function has changed. Channel, input_pixel_stride and output_pixel_stride and now passed to reshape as part of the work on reshaping xnnpack subgraphs

test/softmax.cc
```diff
@@ -57,6 +96,67 @@ TEST_F(SoftmaxTestF32, define)
...
+  std::uniform_real_distribution<float> f32dist(90.0f, 100.0f);
+  std::generate(input.begin(), input.end(), [&]() { return fp16_ieee_from_fp32_value(f32dist(rng)); });
+  std::fill(operator_output.begin(), operator_output.end(), fp16_ieee_from_fp32_value(nanf("")));
+  std::fill(subgraph_output.begin(), subgraph_output.end(), fp16_ieee_from_fp32_value(nanf("")));
+
+  ASSERT_EQ(xnn_status_success, xnn_initialize(/*allocator=*/nullptr));
+
+  // Call operator API.
+  xnn_operator_t op = nullptr;
+  const xnn_status status = xnn_create_softmax_nc_f16(channels, channels, channels, /*flags=*/0, &op);
```
alankelly, line 112: Please rebase, all shape dimensions have moved to reshape

test/concatenate5.cc
alankelly: This may be the only one you have to fix, it's the only one failing to build internally

src/subgraph/convolution-2d.c
GregoryComer: Yeah, I noticed that as well. I'll create a follow-up PR to allow passing fp16 weights for conv & fully-connected.

GregoryComer:
I've rebased on top of the latest master commit and updated usages. I also added test coverage for fp16 convert, which I previously missed. I also addressed the issue with fp16.h import in concatenate5.cc. Tests and build are all passing locally (via cmake), so I believe everything has been addressed and is ready for review. Thanks!

alankelly (approved):

src/subgraph/fully-connected.c
```diff
@@ -994,6 +1003,7 @@ enum xnn_status xnn_define_fully_connected(
     }
 
     switch (bias_value->datatype) {
+      case xnn_datatype_fp16:
```
digantdesai, line 1006: Not allow this?

GregoryComer: It will get caught by the follow up logic. I was planning to release a follow up pr to allow passing fp16 weights, so I think it's fine to leave here.

alankelly:
Sorry for the broken bazel build, someone is working on it internally.

Can you please add  @FP16 to all the dependencies of all the tests you touched in test/BUILD.bazel

alankelly:
The bazel fix has landed so if you rebase it should work.

GregoryComer:
@alankelly 

I've rebased and updated the bazel test targets. Building locally, everything looks good.

alankelly:
Hi, can you please rebase? There are conflicts.

GregoryComer:
Sure, I've rebased and everything looks clean.

GregoryComer:
Actually, looks like during rebase some extra commits ended up getting pulled in somehow. I'll clean it up and then we should be good to go from my end.

GregoryComer:
Done.

GregoryComer:
@alankelly Is there anything else blocking this PR from merging? There's no rush, just want to ensure I'm not missing anything. Thanks.

alankelly:
Sorry forgot about this. There are some failing tests:

Add2, divide2, maximum2, multiply2, squared_difference, subtract2  fail now, looks like the new f16 tests needs:

  ASSERT_EQ(xnn_status_success, xnn_initialize(/*allocator=*/nullptr));

GregoryComer:
Thanks. I'll update the PR to address.

GregoryComer:
Hi @alankelly, regarding the test failures, are these internal tests? I'm not seeing any test failures running via cmake/ctest on linux/x86-64. Looking at divide2.cc, for example, I do see the xnn_initialize calls in both of the new fp16 tests. It does seem like the tests you mentioned do use the BinaryTest fixture, so maybe there's something there. Could you provide any info on the specific failures you're seeing, such as a stack trace or error message? Thanks.

GregoryComer:
I just pre-emptively rebased again. Looked like there was one conflict that needed to be addressed. Tests pass locally. I still do need a little more information on the test failures mentioned above.

PR #6009, opened 2024-02-10
Reshape Conv2d Output

mcr229:
Reshape Convolution 2d output

```shell
cd XNNPACK
bash script/build-local.sh
cd build/local
make -j9 convolution-2d-test
./convolution-2d-test
```

### DIFF
--- a/src/subgraph/convolution-2d.c
+++ b/src/subgraph/convolution-2d.c
@@ -345,108 +345,143 @@ static enum xnn_status reshape_convolution_operator(
 {
   const uint32_t input_id = opdata->inputs[0];
   assert(input_id < num_values);
+
+  const uint32_t output_id = opdata->outputs[0];
+  assert(output_id < num_values);
+
   const size_t batch_size = values[input_id].shape.dim[0];
   const size_t input_height = values[input_id].shape.dim[1];
   const size_t input_width = values[input_id].shape.dim[2];
+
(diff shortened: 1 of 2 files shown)


mcr229:
cc. @digantdesai @alankelly @fbarchard @gonnet for reviews

src/subgraph/convolution-2d.c
```diff
@@ -345,108 +345,143 @@ static enum xnn_status reshape_convolution_operator(
...
+
+  const uint32_t output_id = opdata->outputs[0];
+  assert(output_id < num_values);
+
   const size_t batch_size = values[input_id].shape.dim[0];
   const size_t input_height = values[input_id].shape.dim[1];
   const size_t input_width = values[input_id].shape.dim[2];
+
+  size_t output_height, output_width;
+  enum xnn_status status = xnn_status_invalid_state;
```
fbarchard, line 357: prefer early return and remove this variable.

mcr229: not sure what you mean. It seems like we have to update the output height width given to us by the reshape APIs, so not sure how to return early from there.

fbarchard: ah I see.  Normally subgraph would just call operators that do all the work, but I can approve this if thats the way you want to implement it.

fbarchard (approved):

PR #6015, opened 2024-02-12
Reshape static resize

mcr229:
Reshaping Static Resize, Allowing resizing input so long as numel of input = numel of output. We throw an xnn_status_invalid_parameter otherwise.


mcr229:
cc: @gonnet @digantdesai @alankelly @fbarchard for reviews

mcr229:
seems like this was already done so closing

PR #6027, opened 2024-02-16
Add RVV F32-IGEMM

bhbruce:
Support RVV F32-IGEMM with MR=1 & 7, NR=4v.


bhbruce:
Hi @fbarchard 
This is the RVV IGEMM implementation and NR is also determined by LMUL & VLEN.
The main idea is similar to https://github.com/google/XNNPACK/pull/5893.

scripts/generate-f32-igemm.sh
```diff
@@ -367,6 +367,14 @@ tools/xngen src/f32-igemm/avx512-broadcast.c.in -D MR=6 -D NR=16 -o src/f32-igem
 tools/xngen src/f32-igemm/avx512-broadcast.c.in -D MR=7 -D NR=16 -o src/f32-igemm/gen/f32-igemm-7x16-minmax-avx512f-broadcast.c &
 tools/xngen src/f32-igemm/avx512-broadcast.c.in -D MR=8 -D NR=16 -o src/f32-igemm/gen/f32-igemm-8x16-minmax-avx512f-broadcast.c &
 
+################################ RISC-V Vector ################################
+tools/xngen src/f32-igemm/MRxNRv-rvv.c.in -D MR=1 -D NR=m4 -D ACTIVATION=LINEAR -o src/f32-igemm/gen/f32-igemm-1x4v-rvv.c &
+tools/xngen src/f32-igemm/MRxNRv-rvv.c.in -D MR=7 -D NR=m4 -D ACTIVATION=LINEAR -o src/f32-igemm/gen/f32-igemm-7x4v-rvv.c &
+tools/xngen src/f32-igemm/MRxNRv-rvv.c.in -D MR=1 -D NR=m4 -D ACTIVATION=RELU   -o src/f32-igemm/gen/f32-igemm-1x4v-relu-rvv.c &
+tools/xngen src/f32-igemm/MRxNRv-rvv.c.in -D MR=7 -D NR=m4 -D ACTIVATION=RELU   -o src/f32-igemm/gen/f32-igemm-7x4v-relu-rvv.c &
+tools/xngen src/f32-igemm/MRxNRv-rvv.c.in -D MR=1 -D NR=m4 -D ACTIVATION=MINMAX -o src/f32-igemm/gen/f32-igemm-1x4v-minmax-rvv.c &
+tools/xngen src/f32-igemm/MRxNRv-rvv.c.in -D MR=7 -D NR=m4 -D ACTIVATION=MINMAX -o src/f32-igemm/gen/f32-igemm-7x4v-minmax-rvv.c &
```
fbarchard, line 376: Normally we'd do a few more sizes so benchmarks on different ML models or cpus can be compared.
These are likely the right choice, in my experience, so its fine if youre comfortable with 7x4.
In my experience 7 is best on machines with an 8 way cache.  7 for the inputs and 1 for the weights.
4v is best on cpus with 4 vector units, but it fine for 2 vector units.
In the past we use MR=6 for cpus with 2 or 3 vector units, but I'm finding short MR and wide NR scales better with more vector units, larger memory and more complex gemms that support 4 bit.
This gemm/igemm, for example, has the option to support 4 or 8 bit weights on arm and intel.  Each weight vector gets converted...e.g.  int8 to float.  That works best with a wide but short kernel.
The relu and linear variations have only been used with wasmsimd in the past, due to very expensive min/max instructions.  You dont really need to implement them - minmax will usually be a very small impact on overall performance.
It would be nice in the future if cpu included a since vminmax instruction or a clamp or convert that takes parameters.  The qs8 gemm internally accumulates int, converts to float, scales, clamps to min/max values and converts to int8.  Its outside the loop but steps could be saved.

src/f32-igemm/gen/f32-igemm-7x4v-minmax-rvv.c
```diff
@@ -0,0 +1,183 @@
...
+      size_t k = kc;
+      do {
+        const float va0 = *a0++;
+        const float va1 = *a1++;
+        const float va2 = *a2++;
+        const float va3 = *a3++;
+        const float va4 = *a4++;
+        const float va5 = *a5++;
+        const float va6 = *a6++;
+        vfloat32m4_t vb = __riscv_vle32_v_f32m4(w, vl);
```
fbarchard, line 135: overall this looks awesome.  But I wonder if this load is a bottleneck and should be scheduled carefully.
in arm assembly kernels I would preload some of the weights and inputs in a prologue, so in the loop the first instruction is a vfmacc, and then co-issue loads for the next iteration of the loop.
It usually starts by unrolling the loop to do 2 iterations, so you can load the weights for the next iteration while vfmacc-ing the current weights.
And then a prologue that is the same as the main loop, but no loads for the next iteration.
Also on larger models (LLMs) I'm finding a prefetch on weights helps.   That might be hard for RVV, not knowing how many bytes you're loading, so an alternative would be a redundent vle32 far into the future weights, but ignore the result?  The distance is usually fixed in practoce - 448 for arm and 960 for intel, depending on your cache.  448 is 7 cache lines of 64 for an 8 way cache.  960 is for a 15 cache lines of 64, for a 16 way cache.  Going beyond those sizes can cause collisions and performance can drop off.

fbarchard (approved):
Looks great!  Consider doing a 'pipelined' version of gemm/igemm where the loads are scheduled to be free.
Consider doing 4 and/or 8 bit weights.  8 bit should be easy for RVV... convert the int8 weights to float.

fbarchard (approved):

fbarchard (approved):

bhbruce:
Hi Frank,
This PR now has several conflicts since `tools/generate-gemm-test.py` has been refactored.
Let's try to approve and merge PR https://github.com/google/XNNPACK/pull/5893 at first.
After that, I'll refactor this PR.

fbarchard (approved):

bhbruce:
RVV IGEMM test cases are updated.

PR #6030, opened 2024-02-16
[RFC][NOT_FOR_COMMIT] 4-bit Weight Blockwise Quantization for qd8-f[16,32]

digantdesai:
# 4-bit Weight Blockwise Quantization for qd8-f[16,32]

## Introduction

This proposal aims to explore the implementation of blockwise quantization for 4-bit weights for qd8-f16 and qd8-f32 in XNNPACK, a concept that is particularly beneficial for Low Latency Models (LLM) and Transformers. I believe, this was initially introduced in a research [paper](https://arxiv.org/pdf/1909.05840.pdf) and has been implemented in slightly different forms across other frameworks. For the XNNPACK backend in PyTorch, we propose to investigate this concept for LLMs running on CPUs.

At a high level this can be seen as an extension of the `per_channel` weight quantization (qc4w) with multiple scales per weight output channels instead of one, based on the blocksize.

## Objective

The primary aim of this proposal is to get early feedback using this proof-of-concept PR for the blockwise quantization for the fully-connected operator. This type of quantization is static (weight only) and is planned to be used with f16 or f32 dynamically quantized fully-connected operators with 4-bit weights.

This proposal is **not** intended for a detailed review or merging at this stage. There are a lot of design decisions we need to make for this and for other similar quantization schemas likely to come up in the future. This RFC PR is for sharing the technical design aspects and using it as a baseline for effective design review and discussion.

I am hoping to get feedback around packing and precalculations, kernel designs, API design (operator and subgraph), and naming convention. I also hope to discuss a potential route to merge a version of this which we all agree upon to align XNNPACK with our timeline.

### WEIGHT PACKING
We propose a new weight packing routine that packs scale for a block in between 4-bit weights. Additionally, we introduce another packed weight update routine to fill-in the scales for each block after the weights have been packed.

### KERNEL
A minimally viable 1x2 scalar kernel is added for experimentation. It also illustrates a high level algorithm an optimized kernel might implement. We anticipate further development and refinement of next set of kernel based on feedback and suggestions without major changes in packing or higher level APIs.

### OPERATOR
We propose the addition of a new operator with a qb4w suffix, where b stands for blockwise. We also suggest the introduction of xnn_datatype_qbint4 as a new data type, and xnn_operator_type_fully_connected_nc_f32_qb4w as a new operator type. The block size will be passed down in xnn_create_fully_connected_nc_qd8_f32_qb4w.

### SUBGRAPH
A new tensor API xnn_define_blockwise_quantized_tensor_value with block_size and 2D scale parameters is introduced. The fully-connected subgraph operator will be updated to call the corresponding qb4w operator APIs.

## Limitations and Future Work
This proof-of-concept has certain known limitations, such as the lack of support for transposed weights or vectorized kernels. However, we believe that by taking a depth-first approach and getting a working end-to-end PR for RFC, we may iterate on it much faster than backtracking and updating frequently.


digantdesai:
> from [c2d5ea0 ](https://github.com/google/XNNPACK/commit/c2d5ea06f9e8efe0d4d10e18482cfc3e8395e5d0)to [faf230b ](https://github.com/google/XNNPACK/commit/faf230bbef3d8d32c747515ec559c30d14f7abaf)

Added a single 1x16c4 neon dotprod kernel for experimentation. Haven't added benchmarks yet for `qb4w` nor spent time looking at any performance yet.

src/amalgam/gen/neondot.c
```diff
@@ -1679,6 +1679,291 @@ void xnn_qd8_f32_qc8w_gemm_minmax_ukernel_4x8c4__neondot(
   } while (nc != 0);
 }
 
+void xnn_qd8_f32_qc4w_bl_gemm_minmax_ukernel_4x8c4__neondot(
```
digantdesai, line 1682: @fbarchard - FYI this doesn't spill on armv8. I haven't spend time to benchmark or even write a meta-kernel to codegen these yet.

digantdesai:
pushed v7

digantdesai:
pushed v8 - Add qd8-f16-qb4w support for early experimentation.

digantdesai:
Closing this RFC PR as we ramp up on developing qb4w as a feature. We will be upstreaming it in forms of multiple smaller PRs.

PR #6089, opened 2024-02-27
XNN_FLAG_KEEP_DIM --> XNN_FLAG_REDUCE_DIM

mcr229:
KEEP_DIM flag breaks backwards compatibility as global sum pooling, global avg pooling, and static mean now reduce dims by default. 

Instead I recommend switching the flag to xnn_flag_reduce_dim to opt-in to dim reducing so that prior behavior remains the same.


mcr229:
cc @alankelly

PR #6110, opened 2024-03-04
Add Clang version guard for AMX support in CMake

GregoryComer:
Support for x86 AMX was added in Clang 11 and GCC 11. There is existing logic in the CMake build to conditionally compile in the AMX kernels when the GCC version is greater than or equal to 11. This change adds an additional condition based on the version of Clang being used to compile the project. This change is desirable as PyTorch supports older Clang versions.


CMakeLists.txt
```diff
@@ -90,7 +90,8 @@ OPTION(XNNPACK_ENABLE_RISCV_VECTOR "Build XNNPACK with RISC-V Vector micro-kerne
 OPTION(XNNPACK_ENABLE_VSX "Build XNNPACK with VSX Vector micro-kernels for Power" ON)
 SET(XNNPACK_AVXVNNI_DEFAULT ON CACHE INTERNAL "Default value to enable AVX-VNNI in x86 builds.")
 SET(XNNPACK_AVX512AMX_DEFAULT ON CACHE INTERNAL "Default value to enable AVX512-AMX in x86 builds.")
-IF(CMAKE_C_COMPILER_ID STREQUAL "GNU")
+# Note that support for AMX was added in version 11 of Clang and version 11 of GCC.
+IF(CMAKE_C_COMPILER_ID STREQUAL "GNU" OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
```
fbarchard, line 94: the XNNPACK_AVXVNNI_DEFAULT variable doesnt seem to apply to the option.  I've refactored this to set the option first and then update the option if the compiler is old./
Also the more common issue is the amxintrinsic.h only defines intrinsics for x86_64 regardless of OS

PR #6144, opened 2024-03-08
Fix cmake for pthread and cpuinfo with USE_SYSTEM_LIBS

trixirt:
When -DUSE_SYSTEM_LIBS=ON is used this error is caused

CMake Error in CMakeLists.txt:
  Imported target "pthreadpool" includes non-existent path

    "/include"

...

CMake Error in CMakeLists.txt:
  Imported target "pthreadpool" includes non-existent path

    "/include"

In the USE_SYSTEM_LIBS=ON case, neither the CPUINFO_SOURCE_DIR nor the PTHREADPOLL_SOURCE_DIR variables are defined.

Revert the logic back to what it was before the change in Fixes: tag.

Fixes: e81b7f14eee7 ("Enable use of dependencies in arbitrary locations")


alankelly:
Are you sure that this isn't an issue with the setting of PTHREADPOOL_SOURCE_DIR and CPUINFO_SOURCE_DIR?

PR #6160, opened 2024-03-11
[reshape][xnnpack] use input for dq batch size

mcr229:
I believe dq batch size should be derived from the input shape not output shape in the convert node.


mcr229:
cc @alankelly

alankelly (approved):

PR #6181, opened 2024-03-14
Changes in the build recipe for Hexagon

bolgre:
1. I changed the paths in hexagon.toolchain to build with newer version of Hexagon DSK (5.4.x.x). 

2. $ENV{HEXAGON_SDK_ROOT}/rtos/qurt/computev66/lib/pic/debugmon.o was added to link together to resolve additional undefined reference errors with the path changes. 

3. I added ADD_COMPILE_DEFINITIONS("XNN_LOG_LEVEL=2") in CMakeLists.txt but this may need to be handled differently. 

4. I had errors with min/max function in test/vhswish-microkernel-tester.h for certain cases. Data type for min/max was assumed uint16_t (or it could be uint8_t) but using uint32_t for min/max function caused type mismatch issues. I directly added <uint32_t> for such cases.


bolgre:
I close this PR since I did not use the correct email address, so faced the Google CLA issue.

PR #6182, opened 2024-03-14
Changes to build XNNPACK for Hexagon with newer version of Hexagon SDK

ejparkqc:
1. The paths in hexagon.toolchain were changed to build with newer version of Hexagon DSK (5.x).

2. $ENV{HEXAGON_SDK_ROOT}/rtos/qurt/computev66/lib/pic/debugmon.o was added to link together to resolve additional undefined reference errors with the path changes.

3. ADD_COMPILE_DEFINITIONS("XNN_LOG_LEVEL=2") was added in CMakeLists.txt but this may need to be handled differently.

4. To resolve errors with min/max function in test/vhswish-microkernel-tester.h for certain cases, the data type was explicitly added. min/max assumed uint16_t (or it could be uint8_t) as data type, therefore using uint32_t for min/max function caused type mismatch issues. I directly added <uint32_t> for such cases.


test/vhswish-microkernel-tester.h
```diff
@@ -119,8 +119,8 @@ class VHSwishMicrokernelTester {
         const int32_t input_value = int32_t(uint32_t(input_zero_point() - input[i]) << 7);
         int32_t in = input_value * input_scale_div;
         in -= 16384;  // subtract 0.5 in Q15
-        in = std::min(in, 0);
-        in = std::max(in, -32768);
+        in = std::min<uint32_t>(in, 0);
+        in = std::max<uint32_t>(in, -32768);
```
dsharlet, line 123: It is suspicious to be calling `std::max<uint32_t>` with a negative constant argument and a signed (`int32_t`) argument, is that correct?

test/vhswish-microkernel-tester.h
```diff
@@ -170,8 +170,8 @@ class VHSwishMicrokernelTester {
...
```
dsharlet, line 174: Same comment here as above

CMakeLists.txt
```diff
@@ -138,6 +138,8 @@ ADD_COMPILE_DEFINITIONS("XNN_ENABLE_MEMOPT=$<BOOL:${XNNPACK_ENABLE_MEMOPT}>")
 ADD_COMPILE_DEFINITIONS("XNN_ENABLE_SPARSE=$<BOOL:${XNNPACK_ENABLE_SPARSE}>")
 ADD_COMPILE_DEFINITIONS("XNN_ENABLE_GEMM_M_SPECIALIZATION=$<BOOL:${XNNPACK_ENABLE_GEMM_M_SPECIALIZATION}>")
 ADD_COMPILE_DEFINITIONS("XNN_ENABLE_DWCONV_MULTIPASS=$<BOOL:${XNNPACK_ENABLE_DWCONV_MULTIPASS}>")
+ADD_COMPILE_DEFINITIONS("XNN_LOG_LEVEL=2")
```
dsharlet, line 141: Unrelated debugging change?

dsharlet (commented):
Thank you for updating this!

test/vhswish-microkernel-tester.h
bolgre: You are correct. I missed that. I think having int32_t make sense, and tried to recompile, successful. I will change this.

test/vhswish-microkernel-tester.h
bolgre: This also makes sense to take negative integers, so int32_t is the correct one to add. I will recommit this as well, thank you!

CMakeLists.txt
bolgre: I initially added this to remove the error: 

.../XNNPACK/src/packing.c:15:
.../XNNPACK/src/xnnpack/log.h:20:4: error: "Undefined XNN_LOG_LEVEL"
  #error "Undefined XNN_LOG_LEVEL"
   ^
1 error generated.

Although it worked, I knew this may not be the best solution because I was getting 'XNN_LOG_LEVEL' was redefined in some parts of the build. Any suggestions?

bolgre (commented):
1. Applied Dillon's comments and changed the min/max calls using uint32_t to int32_t to handle negative values properly. 
2. Setting XNNPACK_BUILD_LIBRARY=ON fixed the error related to "undefined XNN_LOG_LEVEL" while compiling packing.c and other files in src.
3. hardware-config need cpuinfo.h to compile, so added the right directory in CMakeLists.txt:906.

dsharlet (approved):
The code looks good to me, thanks for the fixes!

steven-johnson:
On this topic... are there any instructions in the README (or elsewhere) on how to build/test for Hexagon using CMake? If not, we should add some.

steven-johnson:
It would also be good to explicitly specify somewhere which version(s) of the Hexagon SDK are supported.

ejparkqc:
> On this topic... are there any instructions in the README (or elsewhere) on how to build/test for Hexagon using CMake? If not, we should add some.

Agree. I think scripts directory has several build scripts (build-*.sh) covering most supported targets and I used build-qurt-v66.sh for Hexagon build. Mentioning about them in README will be helpful.

ejparkqc:
> It would also be good to explicitly specify somewhere which version(s) of the Hexagon SDK are supported.

Yes good idea and I will update with the version I used. Later this week, I will update hexagon.toolchain to be able to choose the correct paths for Hexagon SDK versions we want to support. Thank you!

steven-johnson:
Thanks, I tried out the build scripts you mentioned (after applying this patch locally) and it appears that there are still some things that don't compile because they haven't been fully ported yet -- eg. `src/configs/binary-elementwise-config.c` and some others that fail with `#error "Unsupported architecture"`

ejparkqc:
> Thanks, I tried out the build scripts you mentioned (after applying this patch locally) and it appears that there are still some things that don't compile because they haven't been fully ported yet -- eg. `src/configs/binary-elementwise-config.c` and some others that fail with `#error "Unsupported architecture"`

@steven-johnson I just pushed several changes in src/configs. Since we don't have microkernels implemented for Hexagon yet, I added placeholders for Hexagon target to remove "Unsupported architecture" errors but they simply calls scalar versions for now. Once we have microkernels written for Hexagon, they will be all replaced by the right ones. 

I was not sure if this is a good solution, so wanted to check in the meeting before commit but now it happened :)

steven-johnson:
> I was not sure if this is a good solution,

Sounds like an excellent short-term solution :-)

steven-johnson:
FYI: with this patch in place, I can now run the `./scripts/build-qurt-v66.sh` script to build everything... but 100% of the tests immediately segfault (via `cd build/qurt/v66 && ctest`). What test(s) are expected to run and is there a different approach to running them?

test/vhswish-microkernel-tester.h
```diff
@@ -119,8 +119,8 @@ class VHSwishMicrokernelTester {
         const int32_t input_value = int32_t(uint32_t(input_zero_point() - input[i]) << 7);
         int32_t in = input_value * input_scale_div;
         in -= 16384;  // subtract 0.5 in Q15
-        in = std::min(in, 0);
-        in = std::max(in, -32768);
+        in = std::min<int32_t>(in, 0);
+        in = std::max<int32_t>(in, -32768);
```
fbarchard, line 123: I see about 600 occurances of  std::min or std::max that dont specify a type.  Is int 32 bits?

src/configs/binary-elementwise-config.c
```diff
@@ -399,6 +399,12 @@ static void init_f32_vadd_config(void) {
     f32_vadd_config.minmax.ropc_ukernel = (xnn_vbinary_ukernel_fn) xnn_f32_vaddc_minmax_ukernel__scalar_u8;
     f32_vadd_config.init.f32_minmax = xnn_init_f32_minmax_scalar_params;
     f32_vadd_config.minmax.element_tile = 8;
+  #elif XNN_ARCH_HEXAGON
```
fbarchard, line 402: normally we'd configure the kernels when they are ready to put into production.  e.g.  hvx kernels.
since these are placeholders, perhaps change the ppc above to
#elif XNN_ARCH_PPC64 || XNN_ARCH_HEXAGON and then add a hexagon section when there is hexagon specific kernels?

ejparkqc:
> FYI: with this patch in place, I can now run the `./scripts/build-qurt-v66.sh` script to build everything... but 100% of the tests immediately segfault (via `cd build/qurt/v66 && ctest`). What test(s) are expected to run and is there a different approach to running them?

Thank you for finding this out. I found the same error when I did ctest. I picked up several tests and manually ran them and all had issues like this:
```bash
/.../XNNPACK/test/abs.cc:104: Failure
  xnn_status_success
    Which is: 0
  xnn_initialize( nullptr)
    Which is: 5
```
Also, this error (This particular example was coming from abs):
```bash
Error in cpuinfo: processor architecture is not supported in cpuinfoError in cpuinfo: processor architecture is not supported in cpuinfoError in cpuinfo: processor architecture is not supported in cpuinfoError in cpuinfo: processor architecture is not sup/.../XNNPACK/test/abs.cc:160: Failure
```
Test commands all use hexagon-sim and it is unclear if the errors came from hexagon-sim itself or xnnpack configure. I will look into this and update soon.

test/vhswish-microkernel-tester.h
ejparkqc: Yes it's int 32bits and yes I usually didn't need to specify the type, but I thought specifying type would fix the issue based on the error message I got:
```bash
/.../XNNPACK/test/vhswish-microkernel-tester.h:123:14: error: no matching function for call to 'max'
        in = std::max(in, -32768);
             ^~~~~~~~
/.../Hexagon_SDK/5.3.0.0/tools/HEXAGON_Tools/8.6.06/Tools/bin/../target/hexagon/include/c++/v1/__algorithm/max.h:39:1: note: candidate template ignored: deduced conflicting types for parameter '_Tp' ('long' vs. 'int')
max(const _Tp& __a, const _Tp& __b)
```
I assume this min/max implementations come from Hexagon SDK, which are targeted for Hexagon. Do you see other min/max are also using namespace std?

src/configs/binary-elementwise-config.c
ejparkqc: I think that's very reasonable. I also see XNN_ARCH_RISCV is configured using scalar as well. Should we put all three cases together for now? So it will become:
#elif XNN_ARC_RISCV || XNN_ARCH_PPC64 || XNN_ARCH_HEXAGON

steven-johnson:
> Thank you for finding this out. I found the same error when I did ctest. I picked up several tests and manually ran them and all had issues like this:

FYI, mine still segfaulted when running manually; running in gdb didn't provide me with useful info. Thanks for investigating.

steven-johnson:
Hi, where do we stand on this? What work is remaining before this can land?

dsharlet:
I just re-ran the CLA check and it passed, I'm surprised that went as smoothly as it did :)

So I think we can now consider merging this. I think the only thing that makes me hesitate is the issue with `std::min`. It is weird we need to do this, because it looks like both types are already `int`. I second @fbarchard 's question re: what is `int` in the hexagon compiler?

steven-johnson:
Any update on being able to run the tests on simulator without crashing?

ejparkqc:
> I found my errors came from the missing files (executable microkernels) - I guess it's because we do not have Hexagon target for the most ukernels yet. (@steven-johnson if your issues are different for mine, please let me know!)
> 
> Only case the test passed was cs16-vsquareabs where we have hexagon ukernels. You can try the command in line 691 in CTestTestfile.cmake and see if this test also passes for you.
> 
> With the current build solution pointing scalar version would fail most cases for now but this will go away once we have all kernels written. Any better solutions?

I am running ctest again after making changes what @fbarchard suggested - having else statement calling scalar ukernel implementation instead of showing error "Unsupported Architecture". 
- Other tests and small benchs seems having issues - xnn_initialize failed and never completes until timeout, so I skipped those cases for now. 
- Kernel tests seems going well (starting from f16-f32-vcvt-test) except some gemm kernels. Once ctest completes I will push the latest changes.

ejparkqc:
Ctest results:

91% tests passed, 21 tests failed out of 240

Total Test time (real) = 6016.17 sec

The following tests FAILED:
          1 - bf16-gemm-minmax-test (Failed)
         10 - f16-f32acc-gemm-minmax-test (Failed)
         11 - f16-gemm-jit-test (Failed)
         12 - f16-gemm-minmax-test (Failed)
         15 - f16-igemm-jit-test (Failed)
         16 - f16-f32acc-igemm-minmax-test (Failed)
         17 - f16-igemm-minmax-test (Failed)
         67 - qd8-f16-qc8w-gemm-minmax-test (Failed)
         75 - f32-gemm-goi-minmax-test (Failed)
         80 - f32-gemm-jit-test (Failed)
         87 - f32-igemm-jit-test (Failed)
        158 - qd8-f16-qc4w-gemm-minmax-test (Failed)
        161 - qs8-qc8w-dwconv-minmax-multipass-fp32-test (Failed)
        163 - qs8-qc8w-gemm-jit-fp32-test (Failed)
        165 - qs8-qc8w-igemm-jit-fp32-test (Failed)
        167 - qs8-dwconv-minmax-multipass-fp32-test (Failed)
        185 - qu8-dwconv-minmax-multipass-fp32-test (Failed)
        192 - qu8-gemm-minmax-fp32-test (Failed)
        194 - qu8-igemm-minmax-fp32-test (Failed)
        195 - qu8-igemm-minmax-rndnu-test (Failed)
        227 - x8-packw-test (Failed)

ejparkqc:
I believe the conflicts are gone now and ready to be merged.

dsharlet:
I still see the `std::max<int32_t>` changes, did we ever figure out what was going on there?

ejparkqc:
> I still see the `std::max<int32_t>` changes, did we ever figure out what was going on there?

I think compilation error comes from hexagon-clang++ where int32_t is considered as 'long' and constants are considered was 'int'. I used version 8.6 in HexagonSDK first then tried 8.8 as well but no luck. Other possible solutions without modifying hexagon-clang++, I could use were:

1. std::max(in,  (int32_t)-32768);
2. std::max(in, -32768l); // Adding 'l' in the end

Not really better than the first solution though. I could not think of any solutions giving specific flags or anything yet.

test/vhswish-microkernel-tester.h
steven-johnson: This is definitely something weird with this compiler/runtime; either integer literals are of type `long` (which seems very unlikely) or their `int32_t` is an alias for `long` (which is plausible). FWIW, an alternate way to fix it is

```
        constexpr int32_t zero = 0;
        constexpr int32_t most_neg = -32768;
        in = std::min(in, zero);
        in = std::max(in, most_neg);
```

but this isn't really any more appealing.

steven-johnson: Yeah, `int32_t` is an alias for `long` rather than `int` in this environment, which is weird but legal I guess:

```
        static_assert(std::is_same<int32_t, int>::value, "int");  <-- fails
        static_assert(std::is_same<int32_t, long>::value, "long");  <-- passes
```

Since integer literals are all of type `int`, std::min that mixes an integer literal is mixing an `int` and a `long`, hence the error. The right way to fix this is to go back in time and change the definition of `int32_t` to be plain `int` here (since `int` and `long` are the same size here according to a similar static_assert() check) but that would be a nightmare to change now I presume, if for no other reason than it would require a new rev to the Hexagon SDK, I suspect.

TL;DR: adding the casts is probably the simplest way to address this.

dsharlet:
I was thinking about this and did some digging around in the XNNPACK code. A few thoughts:
- We had XNNPACK building on the Hexagon SDK before, so it seems like something must have changed between the last version we were attempting to build with and this new version? Or did the XNNPACK code change?
- Can we just #define int32_t (on Hexagon only) somewhere to work around this?
- I was looking at the XNNPACK code, there are already many places that we explicitly specialize std::min/std::max, for similar reasons. So I think maybe we can just go with this change as-is for now. I expect more hassles in the future due to int32_t = long, but we can deal with those as we come across them.

dsharlet (approved):

PR #6196, opened 2024-03-20
Fix Bazel build for bench subdir

steven-johnson:
Several targets were including a nonexistent file in `srcs`, meaning they would not build properly. The file they (apparently) want to include is already in the deps set, so simple removal of these srcs fixes the build. (Perhaps this is a wart that appeared from a Google -> GitHub change migration?)


fbarchard (approved):
this file is generated, but sure

steven-johnson:
> this file is generated, but sure

In that case the bug is that the Bazel build isn't generating the files in all cases that are needed?

PR #6234, opened 2024-03-28
Add cmake flag for building microkernels-all lib

mcr229:
When building and linking static XNNPACK Library, it takes a while because microkernels-all is built(and takes a while) despite not being used by XNNPACK Library. As a result, I'd like to add a flag "XNNPACK_BUILD_ALL_MICROKERNELS" which can be set to turn off building all Microkernels so that this isn't unnecessarily built.

microkernels-all is also only built when building benchmarks and tests, as those targets link against microkernels-all.


mcr229:
cc @alankelly

fbarchard (approved):
Sounds good.  How much build time did it save?

mcr229:
Saved around 1 minute building XNNPACK Library

PR #6267, opened 2024-04-09
fc_qc8w should use qcint32 bias

mcr229:
In the subgraph APIs, FC Layer supports:

input: qs8
weights: qc8
bias: qint32

However, based on what is cone for qc8 filters in convolution, I believe it should be:

input: qs8
weights: qc8
bias: _**qcint32**_

### DIFF
--- a/test/fully-connected.cc
+++ b/test/fully-connected.cc
@@ -171,8 +171,8 @@ TEST_F(FullyConnectedTestQC8, define)
 
   uint32_t bias_id = XNN_INVALID_VALUE_ID;
   ASSERT_EQ(
-    xnn_status_success, xnn_define_quantized_tensor_value(
-                          subgraph, xnn_datatype_qint32, 0, 1.0f, bias_dims.size(), bias_dims.data(), bias.data(),
+    xnn_status_success, xnn_define_channelwise_quantized_tensor_value(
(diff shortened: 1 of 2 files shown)


mcr229:
@alankelly

test/fully-connected.cc
```diff
@@ -759,8 +759,8 @@ TEST_F(FullyConnectedTestQC8, matches_operator_api)
 
   uint32_t bias_id = XNN_INVALID_VALUE_ID;
   ASSERT_EQ(
-    xnn_status_success, xnn_define_quantized_tensor_value(
-                          subgraph, xnn_datatype_qint32, 0, kernel_scale, bias_dims.size(), bias_dims.data(),
```
gonnet, line 763: The variable `kernel_scale` is no longer used and should be removed.

alankelly:
Good catch, this is definitely a bug

test/fully-connected.cc
mcr229: good catch, thanks! Updated.

gonnet (approved):

mcr229:
@gonnet @alankelly Is it possible to merge these changes? Or are there some fixes internally that need to be made in order to safely land this? Sorry, I have no visibility into any internal CIs that may be breaking, and unable to help fix those with this change.

alankelly:
Sorry for the delay, this slipped under the radar. Landed and TFLite fix incoming

PR #6326, opened 2024-04-23
Change Hexagon minimum supported version to v68 (from v66)

ejparkqc:
This modification is to support v68 as the minimum supported Hexagon ISA version. We can also remove scripts/build-qurt-v66.sh not included in this patch yet.


dsharlet:
Does scripts/build-qurt-v66.sh still work, after the cmake paths are changed to v68? If not, we should remove it in this PR.

ejparkqc:
No, scripts/build-qurt-v66.sh will simply pick up the build recipe for v68, so no need to keep this script. I just removed it, so this should be good to go.

dsharlet (approved):

PR #6339, opened 2024-04-25
Amalgam RVV X32-Transpose microkernels

phoebesv:


phoebesv:
Hi @fbarchard @alankelly, 
This PR only enables RVV x32-transpose microkernels. If you have free time, please help to review. Thanks

src/configs/transpose-config.c
```diff
@@ -214,10 +214,41 @@ static void init_transpose_config(void) {
...
+          .tile_size = 32,
+        };
+      } else if (hardware_config->vlenb == 16) {
+        transpose_config.x32 = (struct xnn_transpose_subconfig) {
+          .const_size_ukernel = (xnn_transposec_ukernel_fn) xnn_x32_transposec_ukernel__4x4_rvv,
+          .tile_size = 32,
+        };
+      } else {
+        transpose_config.x32 = (struct xnn_transpose_subconfig) {
+          .const_size_ukernel = (xnn_transposec_ukernel_fn) xnn_x32_transposec_ukernel__2x4_scalar_int,
```
fbarchard, line 242: Is scalar needed?  Can't we use 4x4_rvv for all other vector sizes?

src/configs/transpose-config.c
```diff
@@ -214,10 +214,41 @@ static void init_transpose_config(void) {
...
     };
-    transpose_config.x32 = (struct xnn_transpose_subconfig) {
-      .const_size_ukernel = (xnn_transposec_ukernel_fn) xnn_x32_transposec_ukernel__2x4_scalar_int,
-      .tile_size = 32,
-    };
+    #if XNN_ENABLE_RISCV_VECTOR
+      const struct xnn_hardware_config* hardware_config = xnn_init_hardware_config();
+      assert(hardware_config != NULL);
+      if (hardware_config->vlenb >= 128) {
+        transpose_config.x32 = (struct xnn_transpose_subconfig) {
```
fbarchard, line 221: having all these rvv kernels pulls in all of them... is this really best?

fbarchard (approved):

src/configs/transpose-config.c
phoebesv: Thank you for the feedback. It is the best choice for each vlenb.

src/configs/transpose-config.c
phoebesv: When vlenb is 8, according to the rule, the maximum tile size would be 2x2. However, this tile size is too small, falling back to scalar would be preferable.

fbarchard (approved):

PR #6373, opened 2024-05-07
[blockwise] GOI packing routine for qb4w

digantdesai:
Test `packing-test --gtest_filter="PACK_QD8_F32_QB4W_GEMM_GOI_W.*"`

### DIFF
--- a/src/microparams-init.c
+++ b/src/microparams-init.c
@@ -930,6 +930,45 @@ void xnn_init_qs8_to_qs8_qc8w_scale_fp32_params(
   }
 }
 
+void xnn_init_qs8_qb8w_scale_fp32_params(
+  size_t channels,
+  size_t channels_tile,
+  size_t channels_subtile,
+  size_t stride,
+  size_t substride,
+  size_t num_blocks,
+  size_t block_stride,
+  size_t stride_offset,
+  const float scale[XNN_MIN_ELEMENTS(1)],
+  void* packed_w)
+{
+  void* packed_w_saved = packed_w;
+  for (size_t block_start = 0; block_start < num_blocks; block_start++) {
+    packed_w = (void*)((uintptr_t) packed_w_saved + block_start * block_stride);
+    const size_t tiled_channels = round_down_po2(channels, channels_tile);
+    size_t tile_start = 0;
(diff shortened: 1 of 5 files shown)


digantdesai:
cc @alankelly - here is the first PR to start building blockwise support i.e. qb4w. I closed the Draft PR #6030

src/xnnpack/pack.h
```diff
@@ -202,6 +202,38 @@ XNN_INTERNAL void xnn_pack_qs8_qc4w_gemm_goi_w(
   size_t extra_bytes,
   const struct xnn_qs8_qc4w_packing_params* params);
 
+typedef void (*xnn_pack_qs8_qb4w_gemm_fn)(
+  size_t g,
```
alankelly, line 206: Traditionally these parameters were undocumented and cryptic. Could you please break this tradition and give them better names?!

digantdesai: Updated. I feel nc/kc/nr/kr are quite well "baked" in the code and we should document is somewhere like in the docs dir. It might be more meaningful for readability. I can put up a doc change tomorrow.

src/microparams-init.c
```diff
@@ -930,6 +930,45 @@ void xnn_init_qs8_to_qs8_qc8w_scale_fp32_params(
   }
 }
 
+void xnn_init_qs8_qb8w_scale_fp32_params(
```
mcr229, line 933: should this be named qb4w?

digantdesai: Though you may be coming from consistency point of view with-in qb*w. I would argue that, it shouldn't matter actually because the routine is independent of weight bit width. If you [look](https://github.com/google/XNNPACK/blob/master/src/operators/fully-connected-nc.c#L521) qc4w, it uses qc8w routine. I am not sure if we will ever write qb8w but this is just an artifact from qc.

digantdesai: Sounds good. Let me update.

digantdesai:
Thanks @alankelly  - let me push another PR to fix these name changes and other small things.

digantdesai:
#6434 is the follow up PR.

PR #6384, opened 2024-05-07
Upgrade android_ndk_repository rule and upgrade NDK to 25b

dsharlet:


fbarchard (approved):
ndk 25 comes with clang 1, which supports all current intel cpus supported by XNNPack, so this is good for now.  Thanks!

PR #6386, opened 2024-05-07
Hexagon HVX implementation of F32 vbinary

ejparkqc:
We needed to adjust XNN_ALLOCATION_ALIGNMENT to 128 Byte for HVX and use predicate store for the tail part. Ctest passed but performance is not good yet. The next step naturally becomes the performance improvements.


scripts/generate-f32-vbinary.sh
```diff
@@ -591,4 +591,18 @@ tools/xngen src/f32-vbinary/vopc-avx512f.c.in -D OP=SQRDIFF  -D BATCH_TILE=32 -D
 tools/xngen src/f32-vbinary/vopc-avx512f.c.in -D OP=SUB      -D BATCH_TILE=16 -D ACTIVATION=MINMAX -o src/f32-vbinary/gen/f32-vsubc-minmax-avx512f-u16.c &
 tools/xngen src/f32-vbinary/vopc-avx512f.c.in -D OP=SUB      -D BATCH_TILE=32 -D ACTIVATION=MINMAX -o src/f32-vbinary/gen/f32-vsubc-minmax-avx512f-u32.c &
 
+################################### HEXAGON HVX ##################################
+tools/xngen src/f32-vbinary/vop-hexagon.c.in -D OP=ADD     -D BATCH_TILE=32 -D ACTIVATION=MINMAX -o src/f32-vbinary/gen/f32-vadd-minmax-hexagon-u32.c &
```
dsharlet, line 595: Based on the other kernels nearby, it seems like the kernel should be suffixed `hvx`. If I understand correctly, `hexagon`:`x86`::`hvx`:`avx512f` (for example).

scripts/generate-f32-vbinary.sh
```diff
@@ -591,4 +591,18 @@ tools/xngen src/f32-vbinary/vopc-avx512f.c.in -D OP=SQRDIFF  -D BATCH_TILE=32 -D
 tools/xngen src/f32-vbinary/vopc-avx512f.c.in -D OP=SUB      -D BATCH_TILE=16 -D ACTIVATION=MINMAX -o src/f32-vbinary/gen/f32-vsubc-minmax-avx512f-u16.c &
 tools/xngen src/f32-vbinary/vopc-avx512f.c.in -D OP=SUB      -D BATCH_TILE=32 -D ACTIVATION=MINMAX -o src/f32-vbinary/gen/f32-vsubc-minmax-avx512f-u32.c &
 
+################################### ARM NEON ##################################
```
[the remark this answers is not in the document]
ejparkqc: Yes, after I copied the command line examples from ARM NEON section, missed this line to change to HEXAGON HVX. Done!

scripts/generate-f32-vbinary.sh
ejparkqc: Good point and I will commit after making all related changes.

src/f32-vbinary/gen/f32-vmax-hvx-u64.c
```diff
@@ -0,0 +1,65 @@
...
+  assert(input_b != NULL);
+  assert(output != NULL);
+
+
+  const HVX_Vector *vptr_a = (const HVX_Vector *) input_a;
+  const HVX_Vector *vptr_b = (const HVX_Vector *) input_b;
+  HVX_Vector *vptr_o = (HVX_Vector*) output;
+
+  for (; batch >= 64 * sizeof(float); batch -= 64 * sizeof(float)) {
+    HVX_Vector va0 = *vptr_a++;
```
fbarchard, line 35: Is this the best way to load on HVX?
Would it be better to load all va values and then all vb values?
Would it be better to use offsets and then increment the pointers after?
Are there LDP or LD4 instructions like ARM?

src/f32-vbinary/gen/f32-vmax-hvx-u64.c
```diff
@@ -0,0 +1,65 @@
...
+#include <xnnpack/common.h>
+#include <xnnpack/math.h>
+#include <xnnpack/vbinary.h>
+
+void xnn_f32_vmax_ukernel__hvx_u64(
+    size_t batch,
+    const float* input_a,
+    const float* input_b,
+    float* output,
+    const union xnn_f32_default_params params[restrict XNN_MIN_ELEMENTS(1)]) XNN_OOB_READS
```
fbarchard, line 21: Could you mask the remainder read so you can remove OOB_READs?  (Out of Bounds)
HVX has a large over read, and the masked remainder wont normally be a performance bottleneck... the main loop will be.  On AVX512, RiSCV and SVE I expect we'll move towards asan friendly reads using predicates.

fbarchard (commented):
Easy!  Looks good overall.

scripts/generate-f32-vbinary.sh
```diff
@@ -591,4 +591,18 @@ tools/xngen src/f32-vbinary/vopc-avx512f.c.in -D OP=SQRDIFF  -D BATCH_TILE=32 -D
 tools/xngen src/f32-vbinary/vopc-avx512f.c.in -D OP=SUB      -D BATCH_TILE=16 -D ACTIVATION=MINMAX -o src/f32-vbinary/gen/f32-vsubc-minmax-avx512f-u16.c &
 tools/xngen src/f32-vbinary/vopc-avx512f.c.in -D OP=SUB      -D BATCH_TILE=32 -D ACTIVATION=MINMAX -o src/f32-vbinary/gen/f32-vsubc-minmax-avx512f-u32.c &
 
+################################### HEXAGON HVX ##################################
```
fbarchard, line 594: missing OP=DIV

src/f32-vbinary/gen/f32-vmax-hvx-u64.c
ejparkqc: I changed to load all va values then vb values for now. I will improve this part later using special instructions or offset.

ejparkqc:
> I would ask you how fast it is, but we seem to be missing benchmarks for vbinary.
> 
> I'm adding x86 kernels for FP16, which you could refer to for what needs to be done. #6394
> 
> For benchmark I used a real model and tflite benchmark_model with perf record/report to measure improvement:
> 
> Was AVX with VCVT 7.15% xnn_f16_vmulc_minmax_ukernel__f16c_u16 Now AVX512 FP16 2.94% xnn_f16_rmax_ukernel__avx512fp16_u128_acc4
> 
> But I'll try to make a benchmark we can use for F16 and F32 vbinary

The current PR should work with unaligned load as well without changing tester. I will work on the related issue now.

dsharlet:
This PR still contains the 10k+ line additions of microkernels.bzl and microkernels.cmake. You need to revert those changes, and I also think you're missing the microkernels_hexagon.bzl changes. You shouldn't have to make these changes manually, I think you need to run:

`python3 tools/update-microkernels.py -a`

ejparkqc:
> This PR still contains the 10k+ line additions of microkernels.bzl and microkernels.cmake. You need to revert those changes, and I also think you're missing the microkernels_hexagon.bzl changes. You shouldn't have to make these changes manually, I think you need to run:
> 
> `python3 tools/update-microkernels.py -a`

Yeap, I realized that I modified the generated file which is clearly saying "Do not edit" :) I am working on this. 
Also, I will revert the changes for microkernels.bzl and microkernels.cmake.

One related question: We currently have 'hexagon' in ISA_LIST in update-microkernels.py but coming kernels having 'hvx' as isa_name instead of 'hexagon'. cs16-vsquareabs has hexagon kernels using scalar vectors instead of HVX vectors, so renaming cs16-vsqaureabs kernels with including 'hvx' seems not right.

The simplest solution is to have both 'hvx' and 'hexagon' in the ISA_LIST and make related changes. This will generate kernels separately for HVX and Hexagon (non-HVX). If I add 'hexagon' in ARCH_LIST, it tries to generate asm_kernels and jit_kernels, so I want avoid this for now. Do you think this is okay?

dsharlet (approved):
Thanks, this looks good!

BUILD.bazel
```diff
@@ -9,7 +9,7 @@
...
... "ALL_AVXVNNI_MICROKERNEL_SRCS", "ALL_AVX_MICROKERNEL_SRCS", "ALL_F16C_MICROKERNEL_SRCS", "ALL_FMA3_MICROKERNEL_SRCS", "ALL_FMA_MICROKERNEL_SRCS", "ALL_FP16ARITH_MICROKERNEL_SRCS", "ALL_HEXAGON_MICROKERNEL_SRCS", "ALL_HVX_MICROKERNEL_SRCS", "ALL_NEONBF16_AARCH64_MICROKERNEL_SRCS", "ALL_NEONBF16_MICROKERNEL_SRCS", "ALL_NEONDOTFP16ARITH_MICROKERNEL_SRCS", "ALL_NEONDOT_AARCH64_MICROKERNEL_SRCS", "ALL_NEONDOT_MICROKERNEL_SRCS", "ALL_NEONFMA_AARCH64_MICROKERNEL_SRCS", "ALL_NEONFMA_MICROKERNEL_SRCS", "ALL_NEONFP16ARITH_AARCH64_MICROKERNEL_SRCS", "ALL_NEONFP16ARITH_MICROKERNEL_SRCS", "ALL_NEONFP16_MICROKERNEL_SRCS", "ALL_NEONI8MM_MICROKERNEL_SRCS", "ALL_NEONV8_MICROKERNEL_SRCS", "ALL_NEON_AARCH64_MICROKERNEL_SRCS", "ALL_NEON_MICROKERNEL_SRCS", "ALL_RVVFP16ARITH_MICROKERNEL_SRCS", "ALL_RVV_MICROKERNEL_SRCS", "ALL_SCALAR_MICROKERNEL_SRCS", "ALL_SSE2_MICROKERNEL_SRCS", "ALL_SSE41_MICROKERNEL_SRCS", "ALL_SSE_MICROKERNEL_SRCS", "ALL_SSSE3_MICROKERNEL_SRCS", "ALL_WASMRELAXEDSIMD_MICROKERNEL_SRCS", "ALL_WASMSIMD_MICROKERNEL_SRCS", "ALL_WASM_MICROKERNEL_SRCS", "WASM32_ASM_MICROKERNEL_SRCS", "WASM32_JIT_MICROKERNEL_SRCS", "WASMRELAXEDSIMD32_JIT_MICROKERNEL_SRCS", "WASMSIMD32_JIT_MICROKERNEL_SRCS")
```
fbarchard, line 12: this shouldnt change in your CL... its likely a BUILD file change we committed and you need to rebase

BUILD.bazel
```diff
@@ -203,7 +203,7 @@ JIT_WASMRELAXEDSIMD32_SRCS = JIT_WASMSIMD32_SRCS + JIT_WASMRELAXEDSIMD32_COMMONS
...
...ICROKERNEL_SRCS + ALL_AVX_MICROKERNEL_SRCS + ALL_AVXVNNI_MICROKERNEL_SRCS + ALL_AVX2_MICROKERNEL_SRCS + ALL_AVX512F_MICROKERNEL_SRCS + ALL_AVX512SKX_MICROKERNEL_SRCS + ALL_AVX512VBMI_MICROKERNEL_SRCS + ALL_AVX512VNNI_MICROKERNEL_SRCS + ALL_AVX512VNNIGFNI_MICROKERNEL_SRCS + ALL_AVX512AMX_MICROKERNEL_SRCS + ALL_AVX512FP16_MICROKERNEL_SRCS + ALL_F16C_MICROKERNEL_SRCS + ALL_FMA_MICROKERNEL_SRCS + ALL_FMA3_MICROKERNEL_SRCS + ALL_FP16ARITH_MICROKERNEL_SRCS + ALL_HEXAGON_MICROKERNEL_SRCS + ALL_HVX_MICROKERNEL_SRCS + ALL_NEON_MICROKERNEL_SRCS + ALL_NEONBF16_MICROKERNEL_SRCS + ALL_NEONDOT_MICROKERNEL_SRCS + ALL_NEONDOTFP16ARITH_MICROKERNEL_SRCS + ALL_NEONFMA_MICROKERNEL_SRCS + ALL_NEONFP16_MICROKERNEL_SRCS + ALL_NEONFP16ARITH_MICROKERNEL_SRCS + ALL_NEONV8_MICROKERNEL_SRCS + ALL_RVV_MICROKERNEL_SRCS + ALL_RVVFP16ARITH_MICROKERNEL_SRCS + ALL_SCALAR_MICROKERNEL_SRCS + ALL_SSE_MICROKERNEL_SRCS + ALL_SSE2_MICROKERNEL_SRCS + ALL_SSE41_MICROKERNEL_SRCS + ALL_SSSE3_MICROKERNEL_SRCS + ALL_WASM_MICROKERNEL_SRCS + ALL_WASMRELAXEDSIMD_MICROKERNEL_SRCS + ALL_WASMSIMD_MICROKERNEL_SRCS + AARCH32_ASM_MICROKERNEL_SRCS + AARCH64_ASM_MICROKERNEL_SRCS + WASM32_ASM_MICROKERNEL_SRCS + ["src/microparams-init.c"],
```
fbarchard, line 206: ditto.. revert this change

CMakeLists.txt
```diff
@@ -586,6 +586,7 @@ IF(XNNPACK_TARGET_PROCESSOR MATCHES "^x86(_64)?$")
 ENDIF()
 IF(XNNPACK_TARGET_PROCESSOR MATCHES "Hexagon")
   LIST(APPEND ALL_MICROKERNEL_SRCS ${ALL_HEXAGON_MICROKERNEL_SRCS})
+  LIST(APPEND ALL_MICROKERNEL_SRCS ${ALL_HVX_MICROKERNEL_SRCS})
```
fbarchard, line 589: ditto... revert

BUILD.bazel
dsharlet: It's not because of a missing rebase, it's because of https://github.com/google/XNNPACK/pull/6386/files#diff-3b5d59a3985ecd23f243252f1697e786fc5beec50a965c8acd0189155ce89c63. I think it's an OK change to make. Subsequent HVX kernel changes would not have this diff.

src/f32-vbinary/gen/f32-vmax-hvx-u64.c
```diff
@@ -0,0 +1,64 @@
...
+{
+  assert(batch != 0);
+  assert(batch % sizeof(float) == 0);
+  assert(input_a != NULL);
+  assert(input_b != NULL);
+  assert(output != NULL);
+
+
+  const HVX_UVector *vptr_a = (const HVX_UVector *) input_a;
```
fbarchard, line 31: UVector is unaligned?  Does it hurt performance much?

BUILD.bazel
dsharlet: IIRC we suggested making this change at some point (these kernels are HVX, not general hexagon kernels)

scripts/generate-f32-vbinary.sh
```diff
@@ -591,4 +591,18 @@ tools/xngen src/f32-vbinary/vopc-avx512f.c.in -D OP=SQRDIFF  -D BATCH_TILE=32 -D
...
+tools/xngen src/f32-vbinary/vop-hvx.c.in -D OP=ADD     -D BATCH_TILE=32 -D ACTIVATION=MINMAX -o src/f32-vbinary/gen/f32-vadd-minmax-hvx-u32.c &
+tools/xngen src/f32-vbinary/vop-hvx.c.in -D OP=ADD     -D BATCH_TILE=64 -D ACTIVATION=MINMAX -o src/f32-vbinary/gen/f32-vadd-minmax-hvx-u64.c &
```
fbarchard, line 596: v73 hvx has 4 vector units?  So it may be good to generate 1, 2 and 4 vector variations

src/f32-vbinary/gen/f32-vmax-hvx-u64.c
ejparkqc: > UVector is unaligned? Does it hurt performance much?

UVector is aligned with 4 Byte, and I actually found the performance got better for some cases with simulator (e.g., when we have the number of elements that fills up 128 bytes completely). We will find more accurate performance later with actual device.

fbarchard (changes requested):
Needs rebasing.

gen/microkernels.bzl
```diff
@@ -23,6 +23,7 @@ load("fma3_microkernels.bzl", _ALL_FMA3_MICROKERNEL_SRCS = "ALL_FMA3_MICROKERNEL
 load("fma_microkernels.bzl", _ALL_FMA_MICROKERNEL_SRCS = "ALL_FMA_MICROKERNEL_SRCS")
 load("fp16arith_microkernels.bzl", _ALL_FP16ARITH_MICROKERNEL_SRCS = "ALL_FP16ARITH_MICROKERNEL_SRCS")
 load("hexagon_microkernels.bzl", _ALL_HEXAGON_MICROKERNEL_SRCS = "ALL_HEXAGON_MICROKERNEL_SRCS")
+load("hvx_microkernels.bzl", _ALL_HVX_MICROKERNEL_SRCS = "ALL_HVX_MICROKERNEL_SRCS")
```
alankelly, line 26: This line is causing internal bazel builds to fail. Is this file missing?

ejparkqc: Thank you for letting me know. I realized that I did not add into the repo. Just added.

scripts/generate-f32-vbinary.sh
ejparkqc: I just added the the microkernels using 4 vector units. I tried with simulator v68 vs. v73, but they took exact same time. This is the results coming from f32-vadd-minmax:

[----------] 9 tests from F32_VADD_MINMAX__HVX_U32 (37 ms total)
[----------] 9 tests from F32_VADD_MINMAX__HVX_U64 (96 ms total)
[----------] 9 tests from F32_VADD_MINMAX__HVX_U128 (282 ms total)

dsharlet: I think that is measuring the time for the simulator to run, not the simulated time it would take to run the kernel. I wouldn't make any performance judgements based on that.

That said, I wouldn't be surprised if clang was unrolling it for you anyways, and so these kernels are probably generating the same code.

ejparkqc:
I think the changes I made for bazel build fails. If the build bot fails again, I will revert all the changes for bazel build for now.

BUILD.bazel
ejparkqc: As we want to have regular hexagon kernels and hvx kernels separately, I think I can leave like this, but please let me know if I have to revert this change as well!

CMakeLists.txt
ejparkqc: Same as above, I will leave as it is to separate hvx and regular hexagon kernels, but please let me know if this should be reverted.

test/f32-vmax.cc
```diff
@@ -1316,6 +1316,183 @@
 #endif  // XNN_ARCH_WASM || XNN_ARCH_WASMSIMD || XNN_ARCH_WASMRELAXEDSIMD
 
 
+TEST(F32_VMAX__HVX_U32, batch_eq_32) {
```
dsharlet, line 1319: This is failing internal presubmits because this needs to be wrapped in an `#if XNN_ARCH_HEXAGON` or similar, because it's trying to run these tests on inappropriate architectures.

dsharlet: (Same with the other tests too)

ejparkqc: Thank you Dillon! I just fixed the error and generated the proper if macro for HVX kernels.

ejparkqc: Clearly the new change did not fix the error. Do you see any new errors from build bot result? I tried to build locally for X86 and checked if any build/ctest fails for f32-vbinary but seems all fine.

dsharlet: Actually it looks like you did fix the errors.

This failure is different, and unrelated to your change. You don't need to make any changes to fix it. The system tries to ignore pre-existing failures like this when testing new changes, but it's not always able to do that. I'll check with this change in a bit and see if I can make it retry the internal tests.

PR #6388, opened 2024-05-08
Hexagon build changes after rebase

ejparkqc:
I had two new errors, so I updated the build recipe for Hexagon. 

1. I found XNNPACK_ENABLE_RISCV_VECTOR=ON caused compilation error, so disabled this cmake variable. If you find any issue with this change for Hexagon build, please let me know.
2. This is related to our long discussion about using <int32_t> with min/max function. I found one more tester had compilation issue, so added <int32_t>.


test/deconvolution-operator-tester.h
```diff
@@ -1148,7 +1148,7 @@ class DeconvolutionOperatorTester {
       for (size_t g = 0; g < groups(); g++) {
         for (size_t oc = 0; oc < group_output_channels(); oc++) {
           int32_t range = w8dist(rng);
-          auto weights_dist = std::uniform_int_distribution<int32_t>(std::min(range, 0), std::max(range, 0));
+          auto weights_dist = std::uniform_int_distribution<int32_t>(std::min<int32_t>(range, 0), std::max<int32_t>(range, 0));
```
fbarchard, line 1151: INT32_C(0) for the constants instead?

fbarchard (approved):

test/deconvolution-operator-tester.h
ejparkqc: Quickly tried INT32_C(0) but no luck unfortunately. I will keep investigating if we have better solution than this.

PR #6411, opened 2024-05-14
Enable RVV GEMM/IGEMM 7 x m4 in operator config

bhbruce:
This PR aims to enable RVV GEMM/IGEMM/X32-PACKW in GEMM config. 
It leads to enabling RVV implementation in operator API.


bhbruce:
@alankelly @fbarchard Could you help to review it?
Also, I would like to ask about what's the appropriate way to enable RISC-V only nr2 selection logic in following files.
```
src/operators/convolution-nhwc.c
1653:  const struct xnn_gemm_config* gemm_nr2_config = xnn_init_f32_gemm_nr2_config();
src/operators/dynamic-fully-connected-nc.c
215:  const struct xnn_gemm_config* gemm_nr2_config = xnn_init_f32_gemm_nr2_config();
src/operators/fully-connected-nc.c
754:  const struct xnn_gemm_config* gemm_nr2_config = xnn_init_f32_gemm_nr2_config();
src/operators/deconvolution-nhwc.c
898:  const struct xnn_gemm_config* gemm_nr2_config = xnn_init_f32_gemm_nr2_config();
```

The current logic determines to use nr2_config(half nr) if `gemm_config->nr > output_channels`.
For the RISC-V vector, I would like to specialize in either 
```
gemm_nr2_config->nr <= output_channels
or
gemm_config->nr / 2 <= output_channels
```
However, there is no arch-specifc definition macro used in `src/operators/`.

fbarchard:
nr 2 is an MRx2 GEMM - 2 floats wide.
On SSE and NEON that normally use 4 floats per vector it allows a faster GEMM.
But it is optional... any gemm can output NC of less than a full vector, and on RVV is shouldnt make a difference.

src/configs/gemm-config.c
```diff
@@ -967,6 +967,30 @@ static void init_f32_gemm_config(void) {
       f32_gemm_config.mr = 4;
       f32_gemm_config.nr = 4;
     }
+  #elif XNN_ARCH_RISCV && XNN_ENABLE_RISCV_VECTOR
+    const struct xnn_hardware_config* hardware_config = xnn_init_hardware_config();
+    assert(hardware_config != NULL);
+    if (hardware_config->use_riscv_vector) {
+      f32_gemm_config.linear.gemm[XNN_MR_TO_INDEX(7)] = xnn_init_hmp_gemm_ukernel((xnn_gemm_ukernel_fn) xnn_f32_gemm_ukernel_7x4v__rvv);
+      f32_gemm_config.linear.gemm[XNN_MR_TO_INDEX(1)] = xnn_init_hmp_gemm_ukernel((xnn_gemm_ukernel_fn) xnn_f32_gemm_ukernel_1x4v__rvv);
```
fbarchard, line 975: although faster, we dont normally do linear except on wasmsimd where minmax is very slow.
Its okay, but increases the code size (3 kernels instead of 1)

fbarchard (approved):

src/configs/gemm-config.c
bhbruce: oh... I finally got what you mentioned...

src/configs/gemm-config.c
```diff
@@ -967,6 +967,22 @@ static void init_f32_gemm_config(void) {
...
+    if (hardware_config->use_riscv_vector) {
+      f32_gemm_config.minmax.gemm[XNN_MR_TO_INDEX(7)] = xnn_init_hmp_gemm_ukernel((xnn_gemm_ukernel_fn) xnn_f32_gemm_minmax_ukernel_7x4v__rvv);
+      f32_gemm_config.minmax.gemm[XNN_MR_TO_INDEX(1)] = xnn_init_hmp_gemm_ukernel((xnn_gemm_ukernel_fn) xnn_f32_gemm_minmax_ukernel_1x4v__rvv);
+      f32_gemm_config.minmax.igemm[XNN_MR_TO_INDEX(7)] = xnn_init_hmp_igemm_ukernel((xnn_igemm_ukernel_fn) xnn_f32_igemm_minmax_ukernel_7x4v__rvv);
+      f32_gemm_config.minmax.igemm[XNN_MR_TO_INDEX(1)] = xnn_init_hmp_igemm_ukernel((xnn_igemm_ukernel_fn) xnn_f32_igemm_minmax_ukernel_1x4v__rvv);
+      f32_gemm_config.init.f32 = xnn_init_f32_minmax_scalar_params;
+      f32_gemm_config.pack_gemm_gio = (xnn_packw_gemm_gio_ukernel_fn) xnn_pack_f32_gemm_gio_w;
+      f32_gemm_config.pack_gemm_goi = (xnn_packw_gemm_goi_ukernel_fn) xnn_x32_packw_gemm_goi_ukernel_x4v__rvv_u8;
+      f32_gemm_config.mr = 7;
+      // nr is set to vlmax_e32m4 = 4*VLEN/SEW = 4*VLEN/32 = 4*VLENB*8/32 = VLENB
```
fbarchard, line 982: So really its 4 vectors, so (vlen * 4)  (because m4), but expressed in elements, so vlen * 4 / sizeof(float).
You comment is hard to understand for people that dont know RVV terminology, so maybe use the vlen * 4 / sizeof(float) so it will be more clear when we do fp16 and other types or m2/m8 values.
7x4v makes sense to me.  7 works well on cpus with 8 way l1 cache, and 4 works well on cpus with 2 or 4 vector units.  There is likely no advantage to m8 over m4 on any hardware near term, and a drawback of using more registers, causing spill.
But if you can, test this PR after doing this config, using end2end_bench and/or use gemm_e2e_bench to confirm this is the best kernel.

fbarchard (approved):

src/configs/gemm-config.c
bhbruce: I've modified the comment.
Also, I'll follow your guide to do some tests on different m. I.e m(4~7) x 4v using end2end_bench and/or use gemm_e2e_bench.

bhbruce: Do you think I can add different m size microkernels to XNNPACK like what you do for arm kernels?
I see you add 2x16__aarch64 ~ 6x16__arrch64.

bhbruce:
Hi @fbarchard @alankelly 
Could you help to merge this PR?

fbarchard (approved):

fbarchard:
Re nr2 - if you didnt have such huge vectors you wouldnt have this problem :-)

nr2 doesnt come up much, and you dont have to specialize for it, especially on rvv.
a regular gemm can do nr=2... its just handled as a remainder case.

Add an entry to
static void init_f32_gemm_nr2_config(void) {
with a pack function that can do nr=2   e.g. xnn_x32_packw_gemm_goi_ukernel_x2__scalar_float_u4
normally it would be
      f32_gemm_nr2_config.nr = 2;
meaning 2 floats.  hmmm... I see your issue.  You want something like
      // nr is set to vlen * 4 / sizeof(float) = 4 * VLENB * 8 / 32 = VLENB
      f32_gemm_config.nr = hardware_config->vlenb;
      
what if you break from convention and fill in nr=2, meaning 2 floats = 8 bytes.
and implement the gemm using u1v.  which will work most of the time.
you could check, in the gemm-config, that vlenb >= 8.
you could also check if  vlenb >= 16, and configure an nr2 gemm
but considering how rarely these come up, I'd just do the basic u1v and add a todo to revisit it.

Its also possible to implement nr=2 gemm's more efficiently than the obvious.  I did some for neon, using 4 floats per vector.  and for nr=1 you can do 4 floats at a time.  If thats possible on rvv, it would likely be faster.
I forget the exact method, but look at the 4x2-aarch64-neonfma-ld128.S.in
which does a trick to load 2 blocks at a time (4 floats) and then a paired add outside the loop.
        # Main loop - 4 floats of A (16 bytes)
1:
        LDR         q0,  [x3], 16
        LD2         {v20.4s, v21.4s}, [x5], 32
        LDR         q1, [x11], 16
        LDR         q2, [x12], 16
        LDR         q3,  [x4], 16
        SUBS        x0, x0, 16
        FMLA        v24.4s, v20.4s, v0.4s
        FMLA        v25.4s, v21.4s, v0.4s
        FMLA        v26.4s, v20.4s, v1.4s
        FMLA        v27.4s, v21.4s, v1.4s
        FMLA        v28.4s, v20.4s, v2.4s
        FMLA        v29.4s, v21.4s, v2.4s
        FMLA        v30.4s, v20.4s, v3.4s
        FMLA        v31.4s, v21.4s, v3.4s
        B.HS        1b

        FADDP       v24.4s, v24.4s, v25.4s
        FADDP       v26.4s, v26.4s, v27.4s
        FADDP       v28.4s, v28.4s, v29.4s
        FADDP       v30.4s, v30.4s, v31.4s

fbarchard:
Enable RVV GEMM/IGEMM 7 x m4 is landed in https://github.com/google/XNNPACK/pull/7035
you can close this PR and if add an nr2 enable as followup

bhbruce:
@fbarchard  Thanks for your help.

PR #6493, opened 2024-05-28
Hexagon HVX implementation of F32 gemm

mallick-qc:
Initial functional implementation of f32-gemm for Hexagon HVX. 
Some known outstanding issues
- scalar loop for remainder
- broadcast with vsplat 
- performance analysis of kernels


gen/hvx_microkernels.bzl
```diff
@@ -0,0 +1,16 @@
...
+Auto-generated file. Do not edit!
+  Generator: tools/update-microkernels.py
+"""
+
+ALL_HVX_MICROKERNEL_SRCS = [
+    "src/f32-gemm/gen/f32-gemm-1x32-minmax-hvx-broadcast.c",
+    "src/f32-gemm/gen/f32-gemm-1x64-minmax-hvx-broadcast.c",
+    "src/f32-gemm/gen/f32-gemm-1x128-minmax-hvx-broadcast.c",
+    "src/f32-gemm/gen/f32-gemm-2x128-minmax-hvx-broadcast.c",
+    "src/f32-gemm/gen/f32-gemm-4x64-minmax-hvx-broadcast.c",
```
fbarchard, line 13: Consider 7x64 which is 2 vectors wide, and works well with 8 way cache.

fbarchard (approved):

src/xnnpack/microparams.h
```diff
@@ -170,6 +170,12 @@ union xnn_f32_minmax_params {
     XNN_ALIGN(8) float max[2];
   } wasmsimd;
 #endif  // XNN_ARCH_WASMSIMD || XNN_ARCH_WASMRELAXEDSIMD
+#if XNN_ARCH_HEXAGON
+  struct {
+    XNN_ALIGN(128) float min[128];
```
dsharlet, line 175: Should this be 32 (the array size, not the alignment)? This is a 512 byte array, I think we want 128 bytes here?

mallick-qc: Ah yes that's a mistake should be 32

PR #6507, opened 2024-05-30
 Hexagon HVX implementation of F32 gemm

mallick-qc:
Initial functional implementation of f32-gemm for Hexagon HVX.
Some known outstanding issues

- scalar loop for remainder
- broadcast with vsplat
- performance analysis of kernels


src/f32-gemm/gen/f32-gemm-16x32-minmax-hvx-broadcast.c
```diff
@@ -0,0 +1,326 @@
...
+
+
+#include <assert.h>
+#include <hexagon_types.h>
+#include <hexagon_protos.h>
+#include <hvx_hexagon_protos.h>
+#include <xnnpack/gemm.h>
+
+static XNN_INLINE
+void vstu_variable_scalar(char *bytes, size_t num_bytes, HVX_Vector vin) {
```
fbarchard, line 18: I took a look at EJ's remainder handler, which is a logic nightmare, but it looks correct.  Go ahead and use this for now but followup with trying better ways to handle remainder.
Also I dont think calling a function that is in the same file will work for our amalgamation process.
So this function will need to be global or inlined into the code.
But thats only when you want to hook it up in src/configs/gemm-config.c

There are 3 other ways you can handle remainder
1. power of 2 test/branch.  Store 16 floats.  Copy upper 16 floats to lower
Store 8 floats... copy upper 8 floats
Store 4 floats, 2 floats, 1 float
These stores likely need to handle unaligned pointers
2. create a predicate, which is what EJ did.  Hers does 2 vectors with 2 predicates
3. use a mask table.  On cpus like SSE we create a 'table' which is basically 2 vectors, so in your case, 64 int64 values.  The first 32 are 0xffffffff.  The next 32 are 0
do an unaligned load.
4. generate a mask.  You can use C to make a byte mask
int32_t mask = 1 << (count * 8) - 1;
and then convert that from 32 bytes to 32 int in hvx
or fill a register with 1's  (on intel vcmpeq reg,reg does this) and shift the entire vector by count * 4 bytes.
everything on hvx would be simplier/faster with aligned pointers, so if you somehow ensure allocations are 128 byte aligned, you could add an assert that the output pointer and strides are aligned.  if not possible, it may be common enough that its worth a run time test and/or create 2 kernels.

On newer Intel cpus its looking like 'S4' method of shuffling is a win.  If you were to do this on HVX you can load 4 channels of input at a time and broadcast that, then multiply by each weight vector, and then rotate the 4 floats of input
$if L + 1 != 4:
  $for M in range(MR):
    va${M} = _mm256_permute_ps(va${M}, _MM_SHUFFLE(0, 3, 2, 1));
and FMA against the next set of weights.
instead of 4 loads it becomes 1 load and 3 shuffles.

or if you have registers to spare do KR=4
load and broadcast 4 floats and FMA against the weights.
Outside the loop, do horizontal adds for the accumulators, adding 4 adjacent values.
Works best if you use 4 vectors wide so you end up with 1 full vector.
Compared to doing an 8x128 and 8x32C4 would do the same number of FMA but fewer loads.

fbarchard (approved):

PR #6557, opened 2024-06-12
Update test and benchmark generation for blockwise kr>2

GregoryComer:
This PR updates test generation for blockwise (qb4w) kernels in preparation for ISA-specific kernels with kr > 2. Blockwise kernels currently enforce several constraints:

1) Kc is divisible by block size (no partial blocks in a column).
2) Block size is divisible by 2\*kr\*sr (inner loop is 2x unrolled for 2 planes, no partial epilogue).

The majority of the existing gemm tests cover edge cases that are incompatible with the above constraints, so this PR conditionally disables the irrelevant tests for blockwise kernels. Blockwise kernels still have tests for strides, M subtiles, N subtiles, and varying block/k sizes.

Blockwise tests also benefit from linear stepping of nc and kc. Existing test logic always calls NextPrime after incrementing nc and kc values. To allow for linear stepping in certain blockwise tests while also maintaining the existing behavior, I've tentatively added a step_type parameter to LoopParams, allowing for either linear or prime steps. The default value for step_type is set in GemmTestParams to maintain the existing behavior of prime steps for nc and kc.

Additionally, benchmark logic for blockwise kernels now rounds kc up to a multiple of 2\*kr\*sr.

All tests were regenerated via scripts/generate-tests.sh.

### DIFF
--- a/test/gemm-microkernel-tester.h
+++ b/test/gemm-microkernel-tester.h
@@ -10,6 +10,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <cstdlib>
 #include <functional>
 #include <string>
 
@@ -32,6 +33,25 @@
 #include <xnnpack/post-operation.h>
 #endif  // XNN_PLATFORM_JIT
 
+inline bool IsPrime(size_t n) {
+    if (n == 1 || n == 2) {
+      return true;
+    }
+    for (size_t k = 3; k * k <= n; k += 2) {
+      if (n % k == 0) {
+        return false;
+      }
+    }
+    return true;
+}
+
+inline size_t NextPrime(size_t n) {
+    n = (n + 1) | static_cast<size_t>(1);
+    while (!IsPrime(n)) {
+      n++;
+    }
+    return n;
+}
 
 class GemmMicrokernelTester {
  public:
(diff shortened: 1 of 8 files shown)


test/gemm-microkernel-tester.h
```diff
@@ -32,6 +32,25 @@
 #include <xnnpack/post-operation.h>
 #endif  // XNN_PLATFORM_JIT
 
+inline bool IsPrime(size_t n) {
```
GregoryComer, line 35: These methods are unchanged - I moved them up in the header to allow use from code in this file.

test/gemm-microkernel-tester.h
```diff
@@ -491,14 +510,29 @@ class GemmMicrokernelTester {
...
+enum class LoopStepType {
+  Linear,
+  NextPrime
+};
+
 struct LoopParams {
   LoopParams() = default;
-  explicit LoopParams(size_t from, size_t to, size_t step)
-      : is_set(true), from(from), to(to), step(step) {}
+  explicit LoopParams(size_t from, size_t to, size_t step, LoopStepType step_type)
```
GregoryComer, line 520: If preferred, I could change this to a boolean step_prime variable or similar. I figured the enum class approach is slightly more readable, but does feel less idiomatic to the codebase. Feel free to weigh in on this if you prefer it differently.

test/gemm-microkernel-tester.h
```diff
@@ -549,23 +583,3 @@ struct GemmTestParams {
 };
 
 using GemmTest = testing::TestWithParam<GemmTestParams>;
-
```
GregoryComer, line 552: See above comment - methods are moved to the top of the header.

tools/generate-gemm-test.py
```diff
@@ -236,93 +236,52 @@ def split_ukernel_name(name):
         , test_func, isa_check)
         .loop_n(1, nr)
         .loop_m(1, mr));
-  if (k_block > 1) {
```
GregoryComer, line 239: The diff looks messy, but all that actually changed here is that block of tests that are incompatible with the blockwise kernel assumptions are wrapped in the $if KERNELTYPE not in ['qb4w']. I did re-generate all tests via scripts/generate-tests.sh and there are no functional changes to existing tests.

GregoryComer:
@alankelly Hey Alan, could you take a look at this PR when you have time? Once this is merged, we can start posting the ISA-specific qb4w kernels. Thanks.

alankelly:
@gonnet Can you please review?

tools/generate-gemm-test.py
```diff
@@ -244,464 +244,465 @@ def split_ukernel_name(name):
...
+        gemm_tests.push_back(GemmTestParams(
+            "k_lt_" + akbs,
+            GemmMicrokernelTester()
+                $if EXTENDED_WEIGHTS:
+                  .extended_weights(true)
+                .mr(mr).nr(nr).kr(kr).sr(sr).m(mr).n(nr)
+                $if KERNELTYPE in ['qb4w', 'qc4w']:
+                  .b_zero_point(8)
+                $if KERNELTYPE in ['qb4w']:
+                  .bl(kr * sr * 2)
```
digantdesai, line 258: I guess these aren't necessary

alankelly:
Please rebase and re-run test generation. This PR will be difficult to land since it touches the frequently changing generated bench/test files so after you re run generation, verify that no other tests/benches are changed and then revert the generated files. I will take over the PR internally, re-generate and land. Thanks

PR #6567, opened 2024-06-14
Fix crash in Android Emulator with armeabi-v7a.

API92:
Fix crash of TensorFlow Lite in function xnn_create_convert_nc_qs8 due to dereferencing null pointer in Android Emulator on armeabi-v7a.


alankelly (approved):
Thanks!

alankelly:
Sorry for the delay, importing now

PR #6637, opened 2024-06-27
Fix pointer casting to avoid CFI issues

rajendrant:
To avoid the CFI invalid cast failures observed in http://b/349625080


alankelly:
This has been fixed internally. Thanks

PR #6673, opened 2024-07-03
QB4W Fully Connected Operator

GregoryComer:
Add operator-API level support for qb4w (4-bit blockwise-quantized weight) fully connected operators. Note that this PR depends on the following PRs:

https://github.com/google/XNNPACK/pull/6748
https://github.com/google/XNNPACK/pull/6751
https://github.com/google/XNNPACK/pull/6757


src/operators/fully-connected-nc.c
```diff
@@ -460,6 +460,310 @@ enum xnn_status xnn_create_fully_connected_nc_qd8_f16_qc4w(
     fully_connected_op_out);
 }
 
+enum xnn_status xnn_create_fully_connected_nc_qd8_f16_qb4w(
```
digantdesai, line 463: do you think we can share logic with other `create` functions and use `create_fully_connected_nc`? For both fp16 and fp32?

digantdesai:
Looks good at a high level. Left some comments. Thanks.

PR #6693, opened 2024-07-09
HVX `f32-spmm`

ejparkqc:
Add two implementations so far:

- Default implementation with/without unrolling
   Tested with MR=32,64 and 128, k-block:1,2 and 4.
- Software pipeline with/without unrolling
 Tested with MR=32,64 and 128, k-block:1, 2 and 4.

TODO:
- Block implementation if necessary
- Packing weights from sparse to dense can be implemented using HVX intrinsic as well.


ejparkqc:
@fbarchard @dsharlet Can you review this PR? Thanks!

alankelly:
@Aelphy can you please review?

src/f32-spmm/gen/f32-spmm-32x1-minmax-hvx-pipelined-x2.c
```diff
@@ -0,0 +1,263 @@
...
+        vw = xnn_set1_f32(*w); w += 1;
+        xnn_prefetch_to_l1(w + 32);
+
+        vi0 = xnn_loadu_f32(input + 0);
+      }
+      if XNN_LIKELY(nnz != 0) {
+        do {
+          vacc0 = xnn_fmadd_f32(vi0, vw, vacc0);
+          input = (const float*) ((uintptr_t) input + (uintptr_t) diff);
+          xnn_prefetch_to_l1(input + 128);
```
dsharlet, line 83: I think on Hexagon, vectors only go to L2? This might still be a useful way to prefetch, but perhaps we need to add a `xnn_prefetch_to_l2` that maps to Hexagon's l2fetch.

This should probably wait until we can dig into performance on real hardware more thoroughly.

dsharlet (approved):

src/f32-spmm/gen/f32-spmm-32x1-minmax-hvx-pipelined-x2.c
ejparkqc: I revisited the manual to confirm and I found the correct way to use `dcfetch` is - (1) bring data into L2 first, then (2) bring data into L1 from L2. This is the description from the manual - 

"It is common for algorithms to use the vector engine to produce results that must further process on the scalar core. Use VMEM stores to get the data into L2, then use DCFETCH to get the data in L1, followed by scalar load instructions. Execute the DCFETCH anytime after the VMEM store, however, software should budget at least 30 cycles before issuing 
the scalar load instruction".

Probably I can change xnn_prefetch_to_l1 to have:
```
xnn_prefetch_to_l2(addr, size)
dcfetch(addr)
```
and add `xnn_prefetch_to_l2` to call `l2fetch(addr, size)`. And agree with you about performance analysis first, and we can improve this. Running on devices is one of my active items using Steven's script.

dsharlet: I think we should add an `xnn_prefetch_to_l2` function. But IIRC, Hexagon l2fetches can be 2D, so I think the arguments should be something like: `xnn_prefetch_to_l2_2d(addr, width_bytes, height, stride)` or similar. Maybe we should also have an overload like you described that assumes height = 1 (and then stride is irrelevant).

ejparkqc: I just added `xnn_prefetch_to_l2` in `prefetch.h`. The format is in `xnn_prefetch_to_l2(address, stride, width, height, iter)` - this calls `l2fetch(addr, info)` where `info` has the all the information about stride, width, height and iteration. I got hints from Halide how to prepare the `info`.

When we optimize this code later, we can check - if prefetch is called at the right timing and with right amount (the manual says at least few hundreds cycle before using data).

ejparkqc: Changed slightly having prefetch_to_l2_1D and 2D. In this code, we use 1D function so it is more clear than using 2D function like a overload.

src/f32-spmm/gen/f32-spmm-128x1-minmax-hvx-pipelined-x2.c
```diff
@@ -0,0 +1,396 @@
...
+        vacc1 = xnn_fmadd_f32(vi1, vw, vacc1);
+        vacc2 = xnn_fmadd_f32(vi2, vw, vacc2);
+        vacc3 = xnn_fmadd_f32(vi3, vw, vacc3);
+        input = (const float*) ((uintptr_t) input + (uintptr_t) diff);
+
+        // Use `xnn_prefetch_to_l2` first to prefetch data into the L2 cache,
+        // then, use `xnn_prefetch_to_l1` to bring data into the L1 cache.
+        // format: xnn_prefetch_to_l1(addr)
+        //         xnn_prefetch_to_l2(addr, stride, width, height, iter)
+        xnn_prefetch_to_l2_1d(input + 32, 128);
```
dsharlet, line 97: I don't understand why we prefetch `input + 32`. Shouldn't just be `input`?

src/f32-spmm/gen/f32-spmm-128x1-minmax-hvx-pipelined-x2.c
```diff
@@ -0,0 +1,396 @@
...
+        xnn_prefetch_to_l1(input + 32);
```
dsharlet, line 98: I don't think prefetching vector data to L1 is useful. If nothing else, this is only fetching one cache line of 8(?) that are needed. But since vector data only reaches L2, I don't think we should do it at all.

src/f32-spmm/gen/f32-spmm-128x1-minmax-hvx-pipelined-x2.c
dsharlet, line 97: Should this be prefetching `128 * sizeof(float)`?

ejparkqc: You are correct, should be input. I missed the part that we already modified the input pointers. Changes are coming.

src/f32-spmm/gen/f32-spmm-128x1-minmax-hvx-pipelined-x2.c
ejparkqc: I will leave l2 prefetch only for now.

src/f32-spmm/gen/f32-spmm-128x1-minmax-hvx-pipelined-x2.c
ejparkqc: https://birch/gitbuilder/arch/pkg/v68_stable/doc/page-SYSTEM_USER_l2fetch.html

I think I was using not correctly again! Based on the descriptions, I should give how many 'lines' of 128 bits are going to be brought into L2. so I probably should give 4 in this case? What do you think?

ejparkqc: Also I will change the function name to
- `xnn_prefetch_l2_linear`
- `xnn_prefetch_l2_box`
instead of using 1d and 2d.

src/xnnpack/prefetch.h
```diff
@@ -6,7 +6,29 @@
...
 #ifdef _MSC_VER
-  #include <intrin.h>
+  #include <intrin.h>  
+#endif
+
+#ifdef __hexagon__
+  #include <hexagon_protos.h>
+
+  // Use fetchl2 at least several hundred cycles prior to using the data.
+  // The second parameter (lines) sets the number of lines of 128bits to bring into L2 cache.
```
dsharlet, line 16: 128 bits? Or 128 bytes?

ejparkqc: 128 bytes. Sorry I corrected them. Also I added more description + fixed prefetch_linear function to call right intrinsic.

For linear case, I needed to prepare `info` having
0-7: height
8-15: width
16-31: stride
I set width = stride = 128bytes here.

src/xnnpack/prefetch.h
```diff
@@ -6,7 +6,38 @@
...
+  #include <hexagon_protos.h>
+
+  // Use fetchl2 at least several hundred cycles prior to using the data.
+  
+  // height: the number of 128 bytes to fetch.
+  // width: we set 128 bytes here.
+  // stride: we set 128 bytes here.
+  XNN_INLINE static void xnn_prefetch_to_l2_linear(const void *address, uint8_t height)
+  {
+    uint16_t h0 = (((HEXAGON_Vect32)((128) & 0xff)) << 8) | ((HEXAGON_Vect32)((height) & 0xff));
```
dsharlet, line 22: Isn't `128 & 0xff` just 128?

Why HEXAGON_Vect32? I don't think any casts are needed here.

In general, I think it would be better if the API was just a number of bytes. Wouldn't that make the implementation simpler too? This expression seems like it's basically just computing the number of bytes?

ejparkqc: 1. Done - 128 & 0xff -> 128.

2. Agree with your simpler way - 
I tried to utilize HEXAGON_V32_CREATE_H and HEXAGON_V64_CREATE_H in `hexagon_types.h` - 
```
# define HEXAGON_V32_CREATE_H(h1, h0)
  ((((HEXAGON_Vect32)(h1)) << 16) | ((HEXAGON_Vect32)((h0) & 0xffff)))

#define HEXAGON_V64_CREATE_H(h3, h2, h1, h0)
  ((((HEXAGON_Vect64)(h3)) << 48LL) | (((HEXAGON_Vect64)((h2) & 0xffff)) << 32LL) |
   (((HEXAGON_Vect64)((h1) & 0xffff)) << 16LL) | ((HEXAGON_Vect64)((h0) & 0xffff)))
```
Just verified the info is same with/without casting, so I will change like this:
For linear: `uint32_t info = uint32_t info = ((128 << 16) | ((128 & 0xffff) << 8) | (height & 0xffff));`
For box: `uint64_t info = ((h3 << 48) | ((h2 & 0xffff) << 32) | ((h1 & 0xffff) << 16) | (h0 & 0xffff));`

src/xnnpack/prefetch.h
```diff
@@ -6,7 +6,37 @@
...
+  #include <hexagon_protos.h>
+
+  // Use fetchl2 at least several hundred cycles prior to using the data.
+  
+  // height: the number of 128 bytes to fetch.
+  // width: we set 128 bytes here.
+  // stride: we set 128 bytes here.
+  XNN_INLINE static void xnn_prefetch_to_l2_linear(const void *address, uint32_t height)
+  {
+    uint32_t info = ((128 << 16) | ((128 << 8) & 0xffff) | (height & 0xffff));
```
dsharlet, line 22: This doesn't look right. The bits of h1 and h2 overlap the height.

It looks like you are trying to emulate the macro you were using earlier. Why not just use that? 

More generally, I think maybe we should not do this unless you can run on device and measure a performance improvement, so we know this is working as intended, since it's difficult to test prefetching otherwise.

src/xnnpack/prefetch.h
```diff
@@ -6,7 +6,37 @@
...
+  // stride: an unsigned byte offset which is used to increment the pointer
+  //         after each width-sized block is fetched.
+  // direction: If clear, row major. If set, cloumn major.
+  XNN_INLINE static void xnn_prefetch_to_l2_box(const void *address,
+                                        uint64_t stride,
+                                        uint64_t width,
+                                        uint64_t height,
+                                        uint64_t direction)
+  {
+    uint64_t info = ((h3 << 48) | ((h2 & 0xffff) << 32) | ((h1 & 0xffff) << 16) | (h0 & 0xffff));
```
dsharlet, line 37: Where do h0, h1, h2, h3 come from?

ejparkqc: Fixed, it was wrong paste from tester code.

src/xnnpack/prefetch.h
ejparkqc: For box prefetch, I can use the macro again - 
`uint64_t info = HEXAGON_V64_CREATE_H(direction, stride, width, height);`

For linear prefetch, I need to prepare info into having
0-7: height
8-15: width
16-32: stride
There is no macro defined to prepare this particular dataset, so I can do like this:
```
// set width (128) and height and put into the lower16 bytes.
uint32_t info =(128 << 8) & 0xffff | (height & 0xffff); 

// put stride (128) into the upper 16 bytes.
info =  HEXAGON_V32_CREATE_H(128, info);
```

Yes it would make more sense with performance measurement as I cannot tell if the prefetcher works or not from simulator. Later we also need to make sure to have input and weight is aligned to have prefetch works well.

dsharlet: Why not `HEXAGON_V64_CREATE_H(0, 0, size_bytes, 1)`?

ejparkqc: I guess the last parameter '1' is the height? 

And yes we can use that macro but, shouldn't we still need `(128 << 8) & 0xffff) | (height & 0xffff);` for the lower half-word? Otherwise `width` will be 0 but stride is set to `size_bytes`. If having stride correctly set is sufficient, then we can use `HEXAGON_V64_CREATE_H(0, 0, size_bytes, height)`.

dsharlet: Why is there a height at all? Aren't we trying to describe an N x 1 rectangle, i.e. a single row? The stride doesn't matter, and height is 1? 

Doesn't this macro handle all of the bit shuffling for you?

ejparkqc: Ahh, I assumed `size_bytes` as 128 bytes and `height` as the number of lines of 128 bytes. You meant `size_bytes = BATCH_TILE * sizeof(float)` correct? Then using your method makes perfect sense. And yes confirmed from the manual that `width` is the important part and `stride` doesn't matter. I can change in this way and we can come back and test the performance. Please let me know if I still misunderstood!

ejparkqc:
I kept the two prefetch function defined in prefetch.h, but removed the calls from the kernels for now. As Dillon suggested, we will put this back again when we evaluate on-device performance.

dsharlet (approved):

PR #6694, opened 2024-07-09
Cleanup/Simplify HVX `f32-vbinary/f32-vbinaryc` implementations

ejparkqc:
- Utilize simd/f32-hvx.h instead of using intrinsic directly.
- Remove unnecessary pointers.


ejparkqc:
@fbarchard @dsharlet This is just cleanup/simplified version of f32-binary (both vop and vopc).

alankelly:
@Aelphy @dsharlet can you please review?

dsharlet (approved):

PR #6755, opened 2024-07-22
[QB4][AVX256] Master qb4 avx256skx

mcr229:
QB4 AVX256SKX Kernels


fbarchard (changes requested):
missing the avx2.c.in template?

mcr229:
> missing the avx2.c.in template?

Gregory added a PR https://github.com/google/XNNPACK/pull/6618, to update avx2 kernels and added the block loop in the metakernels there. It looked like the difference between avx=2 vs avx=10 was small, so I didn't edit the meta kernel and just  generated the avx256skx kernels. Let me know if that looks fine

fbarchard (approved):
you've got generator for avx256 but no template change needed?
likely requires a rebase

PR #6756, opened 2024-07-23
Require bl % 32 = 0 for qb4w gemm kernels

GregoryComer:
Add an assertion to qb4w kernels to enforce bl % 32 == 0. Update tests accordingly.


digantdesai:
Looks good. I assume there will be more checks/asserts in the operator PR. I haven't looked at that yet.

PR #6780, opened 2024-07-29
Guard qb4w AVX512-VNNIGFNI kernel references in gemm-config

GregoryComer:
Needed to build on older GCC/Clang versions.


digantdesai:
Thanks @GregoryComer. 

@fbarchard also saw some failures. I am not sure if this is sufficient. See the Slack message, and also see the XNNPACK GH CI email on this PR especially around CMake builds.

GregoryComer:
I'm closing this PR since it looks like the #ifdef was added upstream. I'll open a new PR for the other tensor scale validation fix.

PR #6790, opened 2024-08-01
Add f32 Avgmaxpool RVV implementation microkernels, tests and config changes.

KaustubhIMG:


src/f32-argmaxpool/f32-argmaxpool-4x-rvv-c1v.c
```diff
@@ -0,0 +1,73 @@
...
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include <assert.h>
+#include <riscv_vector.h>
+
+#include "xnnpack/argmaxpool.h"
+
+void xnn_f32_argmaxpool_ukernel_4x__rvv_c1v(
```
fbarchard, line 11: we use 'u1v' for kernels that use 'm1' meaning unrolled to 1 vector.
Normally we'd make this into a template and generate versions for m1, m2, m4 and maybe m8 to allow some performance testing with different hardware and later put the best of them in the config.
the filename should also be -u1v.c

fbarchard (approved):
Looks okay, but I would
1. rename them to u1v instead of c1v
2. use a template to generate variations
But also I think we're moving away from 'pool' kernels to rdsum.
So you might want to implement/optimizes rsum and rdsum.
If you do fp16, there is a regular fp16 sum, but it tends to overflow, so theres a more useful version that accumulates fp32.

src/f32-argmaxpool/f32-argmaxpool-4x-rvv-c1v.c
KaustubhIMG: Ok. I will create a multiple LMUL versions and update it in separate PR.

KaustubhIMG: done

KaustubhIMG:
@fbarchard @allenkelly @bhbruce @keeranroth, Can someone merge this PR?

KaustubhIMG:
@fbarchard @allenkelly @bhbruce @keeranroth, Can someone merge this PR?

bhbruce (approved):
LGTM.

PR #6813, opened 2024-08-07
Xnn s32 and

umadevimcw:
Implements s32 AND op


umadevimcw:
AND op is part of https://github.com/google/XNNPACK/pull/6836. Hence closing it

PR #6823, opened 2024-08-08
Xnn s32 or

umadevimcw:
Bitwise OR implementation


src/amalgam/gen/avx512f.c
```diff
@@ -5557,3 +5553,103 @@ void xnn_s32_vmulc_ukernel__avx512f_u32(
...
+    input_b += 32;
+
+    xnn_simd_s32_t vy_0 = xnn_or_s32(vin1_0, vin2_0);
+    xnn_simd_s32_t vy_1 = xnn_or_s32(vin1_1, vin2_1);
+
+    xnn_storeu_s32(output, vy_0);
+    xnn_storeu_s32(output + 1 * xnn_simd_size_s32, vy_1);
+    output += 32;
+  }
+  for (; batch >= xnn_simd_bytes_s32; batch -= xnn_simd_bytes_s32) {
```
fbarchard, line 5587: you use batch -= 32 * sizeof(int32_t)) { for the main loop.  which is the correct amount for avx512
should this one be batch -= 16 * sizeof(int32_t)) {

src/amalgam/gen/avx512f.c
```diff
@@ -5557,3 +5553,103 @@ void xnn_s32_vmulc_ukernel__avx512f_u32(
...
+    xnn_simd_s32_t vin2 = xnn_loadu_s32(input_b);
+    input_b += xnn_simd_size_s32;
+
+    xnn_simd_s32_t vy = xnn_or_s32(vin1, vin2);
+
+    xnn_storeu_s32(output, vy);
+    output += xnn_simd_size_s32;
+  }
+  if XNN_UNLIKELY(batch != 0) {
+    xnn_simd_s32_t vin1 = xnn_load_tail_s32(input_a, batch >> XNN_LOG2_SIZEOF_INT32_T);
```
fbarchard, line 5600: normally for native we'd create a mask based on the remainder and use it for all the instructions.
the loop here would be the same as the previous loop, but with a mask
its also prudent to put an assert on the expected batch size to ensure we dont accidently have too much or too little for the remainder masking to work

src/amalgam/gen/avx512f.c
umadevimcw: @fbarchard  vin1_0, vin1_1 two set of inputs are processed here, like loop unrolling of 2 so its 32

fbarchard: vin1 is from input_a and is 16 ints
vin2 is from input_b and is 16 ints
should the loop be doing:
for (; batch >= 16 * sizeof(int32_t); batch -= 16 * sizeof(int32_t)) {

umadevimcw:
OR op  is part of https://github.com/google/XNNPACK/pull/6836. Hence closing it

PR #6739, opened 2024-07-19
Enable wasmusdot qs8_qc8w gemm/igemm in gemm-config

fanchenkong1:
This PR is a follow-up work for https://github.com/google/XNNPACK/pull/6505.

| v8/meteor lake                                           | median(Time) / us |
|----------------------------------------------------------|-------------------|
| qs8_qc8w_gemm_2x4c16__wasmusdot/mobilenet_v1/real_time   | 8952              |
| qs8_qc8w_gemm_2x4c16__wasmusdot/mobilenet_v2/real_time   | 7598              |
| qs8_qc8w_gemm_3x4c16__wasmusdot/mobilenet_v1/real_time   | 9629              |
| qs8_qc8w_gemm_3x4c16__wasmusdot/mobilenet_v2/real_time   | 7698              |
| qs8_qc8w_gemm_4x4c16__wasmusdot/mobilenet_v1/real_time   | 9715              |
| qs8_qc8w_gemm_4x4c16__wasmusdot/mobilenet_v2/real_time   | 7994              |
| qs8_qc8w_gemm_2x8c16__wasmusdot/mobilenet_v1/real_time   | 9189              |
| qs8_qc8w_gemm_2x8c16__wasmusdot/mobilenet_v2/real_time   | 7590              |
| qs8_qc8w_gemm_3x8c16__wasmusdot/mobilenet_v1/real_time   | 9773              |
| qs8_qc8w_gemm_3x8c16__wasmusdot/mobilenet_v2/real_time   | 7969              |
| qs8_qc8w_gemm_4x8c16__wasmusdot/mobilenet_v1/real_time   | 10211             |
| qs8_qc8w_gemm_4x8c16__wasmusdot/mobilenet_v2/real_time   | 8783              |

PR #6863, opened 2024-08-14
Enable wasmusdot qd8_f32_qc8w gemm/igemm in gemm-config

fanchenkong1:
This PR is a follow-up work for https://github.com/google/XNNPACK/pull/6771

Performance data for reference: https://github.com/google/XNNPACK/pull/6739

### DIFF
--- a/src/amalgam/gen/wasmrelaxedsimd.c
+++ b/src/amalgam/gen/wasmrelaxedsimd.c
@@ -12381,6 +12381,256 @@ void xnn_qd8_f32_qc8w_gemm_minmax_ukernel_1x4c16__wasmsdot(
   } while (nc != 0);
 }
 
+void xnn_qd8_f32_qc8w_gemm_minmax_ukernel_1x4c16__wasmusdot(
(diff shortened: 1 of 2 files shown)


fanchenkong1:
Hi @fbarchard , could you please take a look at this change? Thanks!

fbarchard (approved):
2x4c16 eh.  Thats a very small tile with a large KR (c16)
a more typical size these days would be 4x8c8
the nature of dot products is KR=4 would process 4 channels at a time.  But reading just 4 bytes of input is a bottleneck.  C8 seems to work well, with 64 bits reads/broadcast and 2 dotproducts per row.

fanchenkong1:
Sorry, I just found an issue with this change as I only enabled the wasmusdot kernels for gemm. 

I'm currently working on [c8 gemm/igemm microkernels for wasm](https://github.com/google/XNNPACK/pull/6921) and plan to update gemm-config again after that is merged. May I close this PR and create a new one once the c8 kernels are finished?

PR #6895, opened 2024-08-19
Minor changes to adjust new params initialization.

ejparkqc:
Minor changes in modifying HVX code to new parameter initialization method.

- f32-rsum
- f32-raddstoreexpminus
- qs8-vadd
- f32-qs8


dsharlet (approved):
Thanks for doing this. Sorry if there were any breakages.

ejparkqc:
@dsharlet I added XNN_FORCE_REALIZATION for HVX but not sure it is necessary for HVX. Please feel free to revert this change if necessary.

dsharlet:
It's probably not necessary on HVX but it's also harmless until we turn it on for the target. I generally am adding it to all kernels that have a lot of broadcasted values, but only enabling it for targets that show a benefit.

dsharlet (approved):

PR #6897, opened 2024-08-20
Simplify the CMake build rules for tests and benchmarks

steven-johnson:
This rewrites the CMake (but not Bazel) build rules for the vast majority of tests and benchmarks to be table-driven, dramatically reducing the size of the CMake file and (arguably) making is simpler to reason about the build rules for test and benchmarks. They are grouped in the same manner they were previously.

Note that one simplifying assumption is that each group of tests/benchmarks ends up linking in the union of all dependencies for everything in that group (relying on the linker to dead-strip unnecessary stuff); for production code this would probably be unacceptable, but for tests and benchmarks, I think it's an acceptable price to pay for simplifying the build rules.


steven-johnson:
Finally green, PTAL

fbarchard (approved):

steven-johnson:
Refactored to make the new lists a bit more granular, PTAL

PR #6898, opened 2024-08-20
Add QS8 QC8W GEMM AVXVNNI e2e Bench

Ch3nYuY:


fbarchard (approved):
Looks good, thanks, but you may need to make a tweek for params, as there have been changes related to gemm/igemm params

dsharlet:
Sorry for the churn here, but we are moving away from this style of benchmark, in favor of subgraph API benchmarks, which are "more end-to-end". These operator API benchmarks really aren't end-to-end, because they execute the model differently than actually happens in real use cases.

PR #6899, opened 2024-08-20
Introduce flags for qb4 scale format in xnn_define_blockwise_quantized_tensor_value

GregoryComer:
This change extends the xnn_define_blockwise_quantized_tensor_value API to accept flags to control block scale format, though only bf16 is currently supported. The intent of this change is to allow for other block scale formats (fp16 or fp32) in the future without breaking API backwards compatibility.

### DIFF
--- a/include/xnnpack.h
+++ b/include/xnnpack.h
@@ -236,6 +236,15 @@ enum xnn_status xnn_delete_subgraph(
 #define XNN_VALUE_FLAG_EXTERNAL_OUTPUT 0x00000002
 #define XNN_VALUE_FLAG_PERSISTENT      0x00000004
 
+/// Quantized block scales for the tensor are in fp32 format.
+#define XNN_VALUE_FLAG_FP32_SCALES 0x00000008
+
+/// Quantized block scales for the tensor are in fp16 format.
+#define XNN_VALUE_FLAG_FP16_SCALES 0x00000010
+
+/// Quantized block scales for the tensor are in bf16 format.
+#define XNN_VALUE_FLAG_BF16_SCALES 0x00000020
+
 #define XNN_INVALID_VALUE_ID UINT32_MAX
 
 /// Type of elements in a Value object.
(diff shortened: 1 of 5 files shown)


src/tensor.c
```diff
@@ -475,6 +475,12 @@ enum xnn_status xnn_define_blockwise_quantized_tensor_value(
     return xnn_status_invalid_parameter;
   }
 
+  if ((flags & XNN_VALUE_FLAG_FP16_SCALES) || (flags & XNN_VALUE_FLAG_FP32_SCALES)) {
```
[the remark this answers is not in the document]
digantdesai: Also I like your design here that regardless of the scale dtype - XNNPACK does not do conversion internally, i.e. if users have FP16 scales and want us to consume that then they can pass a pointer to a FP16 buffer and set a flag.

include/xnnpack.h
```diff
@@ -423,11 +432,12 @@ enum xnn_status xnn_define_channelwise_quantized_tensor_value_v2(
 ///                     number of input channel element per output channel.
 ///                     For Fully connected operators with 2d filters of size [output_channels, input_channels],
 ///                     expecting number of scale values to be = output_channels * (input_channels / block_size).
+/// @param flags - binary features of the Value. Supported values are XNN_VALUE_FLAG_BF16_SCALES.
```
digantdesai, line 435: Please add more documentation to make it crystal clear for someone who has no idea about internal implementations.

include/xnnpack.h
```diff
@@ -273,8 +282,8 @@ enum xnn_datatype {
   /// 32-bit signed integers.
   xnn_datatype_int32 = 11,
   /// Quantized 4-bit signed integer with shared per-channel-block quantization
-  /// parameters.
-  xnn_datatype_qbint4 = 12,
+  /// parameters (bf16 format).
+  xnn_datatype_qbint4_bf16 = 12,
```
digantdesai, line 286: How does this specialization trickle down to gemm-config and kernels selection? It might be OK for now since we don't have other d-types but imagine similar set of kernels but for fp16 scales. We need to disambiguate them at some level. 

This is also applicable to an extent to other APIs like create/reshape etc. They might need to know what is the scale dtype.

include/xnnpack.h
```diff
@@ -273,8 +282,8 @@ enum xnn_datatype {
   /// 32-bit signed integers.
   xnn_datatype_int32 = 11,
   /// Quantized 4-bit signed integer with shared per-channel-block quantization
-  /// parameters.
-  xnn_datatype_qbint4 = 12,
+  /// parameters (bf16 format).
```
digantdesai, line 285:
```suggestion
  /// parameters (scale dtype = bf16).
```

PR #7003, opened 2024-09-03
Integrating KleidiAI's QB4W kernels into XNNPACK.

mcr229:
WIP of integrating KleidiAI's QB4W kernels. The current status is that all neon dot product kernels (mr=1) are verified to be working. They are hooked up to operator and subgraph level APIs with tests.


src/configs/gemm-config.c
```diff
@@ -1745,6 +1747,28 @@ static void init_qp8_f32_qc4w_gemm_config(void) {
 #endif  // XNN_ARCH_ARM64 && XNN_ENABLE_KLEIDIAI
 }
 
+static void init_qp8_f32_qb4w_gemm_config(void) {
+#if XNN_ARCH_ARM64 && XNN_ENABLE_KLEIDIAI
+  const struct xnn_hardware_config* hardware_config =
+      xnn_init_hardware_config();
+  assert(hardware_config != NULL);
+  if (XNN_ENABLE_ARM_DOTPROD && hardware_config->use_arm_neon_dot) {
+#if XNN_ENABLE_ARM_DOTPROD
```
alankelly, line 1756: indentation

src/configs/gemm-config.c
```diff
@@ -1745,6 +1747,28 @@ static void init_qp8_f32_qc4w_gemm_config(void) {
...
+    qp8_f32_qb4w_gemm_config.init.f32_qb4w = xnn_init_f32_qb4w_minmax_scalar_params;
+    qp8_f32_qb4w_gemm_config.pack_weights_and_biases = xnn_pack_kai_qb4_weights_and_biases;
+    qp8_f32_qb4w_gemm_config.packed_stride_weights_and_biases = xnn_packed_stride_kai_qb4_weights_and_biases;
+    qp8_f32_qb4w_gemm_config.mr = 1;
+    qp8_f32_qb4w_gemm_config.nr = 8;
+    qp8_f32_qb4w_gemm_config.log2_kr = 4;
+    qp8_f32_qb4w_gemm_config.log2_sr = 1;
+    qp8_f32_qb4w_gemm_config.planes = 2;
+    qp8_f32_qb4w_gemm_config.mr_packed = 1;
+#endif  // XNN_ENABLE_ARM_DOTPROD
```
alankelly, line 1767: indentation

src/configs/gemm-config.c
```diff
@@ -1745,6 +1747,28 @@ static void init_qp8_f32_qc4w_gemm_config(void) {
...
 }
 
+static void init_qp8_f32_qb4w_gemm_config(void) {
+#if XNN_ARCH_ARM64 && XNN_ENABLE_KLEIDIAI
+  const struct xnn_hardware_config* hardware_config =
+      xnn_init_hardware_config();
+  assert(hardware_config != NULL);
+  if (XNN_ENABLE_ARM_DOTPROD && hardware_config->use_arm_neon_dot) {
+#if XNN_ENABLE_ARM_DOTPROD
+    qp8_f32_qb4w_gemm_config.minmax.qp8gemm_bl[XNN_MR_TO_INDEX(1)] = xnn_init_hmp_qp8gemm_bl_ukernel(xnn_qp8_f32_qb4w_gemm_minmax_ukernel_1x8c16s2__aarch64_neondot);
```
gmiodice, line 1757: Just a note: From our experiments, we have found that the 8(mr)x4(kr) ukernel variant performs better than the 4(mr)x8(kr). If this is the case, we might need the 1x4 variant here to re-use the same RHS packed matrix. For now, we can keep this variant as we haven't integrated the GEMM variant kernel for the e2e test yet.

mcr229: hmm is the idea here that we fall back on the neon dot 1xkr ukernels when computing tiles where m < mr?

mcr229: This makes sense. When I create the config for i8mm i'll have the fallback (mr=1) be the 1x4 neondot kernel and i'll use the 8x4 neoni8mm kernel

mcr229:
Some Initial Benchmarks with Llama 3.1 with ExecuTorch:

Benchmarks were run on S22 Device. Llama Benchmarks were run on 4 threads.

## BaseLine:
Gemm Config Contains: 4x16 i8mm ukernel and 1x16 i8mm ukernel:
```
I 00:00:35.791154 executorch:stats.h:84]        Prompt Tokens: 8    Generated Tokens: 119
I 00:00:35.791162 executorch:stats.h:90]        Model Load Time:                18.548000 (seconds)
I 00:00:35.791225 executorch:stats.h:100]       Total inference time:           17.223000 (seconds)              Rate:  6.909365 (tokens/second)
I 00:00:35.791256 executorch:stats.h:108]               Prompt evaluation:      2.234000 (seconds)               Rate:  3.581021 (tokens/second)
I 00:00:35.791262 executorch:stats.h:119]               Generated 119 tokens:   14.989000 (seconds)              Rate:  7.939155 (tokens/second)
I 00:00:35.791266 executorch:stats.h:127]       Time to first generated token:  2.234000 (seconds)
I 00:00:35.791271 executorch:stats.h:134]       Sampling time over 127 tokens:  0.194000 (seconds)
```

## QP8 (8x4):
Gemm Config Contains: 8x4 i8mm ukernel and 1x4 neondot ukernel
```
I 00:00:47.736503 executorch:stats.h:84]        Prompt Tokens: 8    Generated Tokens: 119
I 00:00:47.736511 executorch:stats.h:90]        Model Load Time:                28.177000 (seconds)
I 00:00:47.736546 executorch:stats.h:100]       Total inference time:           19.536000 (seconds)              Rate:  6.091319 (tokens/second)
I 00:00:47.736553 executorch:stats.h:108]               Prompt evaluation:      2.304000 (seconds)               Rate:  3.472222 (tokens/second)
I 00:00:47.736558 executorch:stats.h:119]               Generated 119 tokens:   17.232000 (seconds)              Rate:  6.905757 (tokens/second)
I 00:00:47.736564 executorch:stats.h:127]       Time to first generated token:  2.304000 (seconds)
I 00:00:47.736568 executorch:stats.h:134]       Sampling time over 127 tokens:  0.208000 (seconds)
```

## QP8 (4x8):
Gemm Config Contains 4x8 i8mm ukernel and 1x8 neondot ukernel
```
I 00:00:48.158178 executorch:stats.h:84]        Prompt Tokens: 8    Generated Tokens: 119
I 00:00:48.158186 executorch:stats.h:90]        Model Load Time:                29.479000 (seconds)
I 00:00:48.158328 executorch:stats.h:100]       Total inference time:           18.661000 (seconds)              Rate:  6.376936 (tokens/second)
I 00:00:48.158336 executorch:stats.h:108]               Prompt evaluation:      2.249000 (seconds)               Rate:  3.557137 (tokens/second)
I 00:00:48.158343 executorch:stats.h:119]               Generated 119 tokens:   16.412000 (seconds)              Rate:  7.250792 (tokens/second)
I 00:00:48.158349 executorch:stats.h:127]       Time to first generated token:  2.249000 (seconds)
I 00:00:48.158354 executorch:stats.h:134]       Sampling time over 127 tokens:  0.202000 (seconds)
```

Some perf differences we notice are regressions in Model Load Time (this is likely due to changes in RHS packing, as that is generally where all the time is spent at model load). We don't notice much change in TTFT (prefill), but we do  notice some regressions in Token Generation. We also notice 4x8 Kernel slightly outperforms 8x4 Kernel, which is curious because 8x4 Kernel is faster. 

## Further Benchmarks and Analysis

Digging further into the benchmarks of the ukernels I decided to investigate the prod kernels used in gemm config
|     |                       |                   | MC:128/NC:16/KC:1024/BL:32 | MC:128/NC:32000/KC:4096/BL:32 |
|-----|-----------------------|-------------------|----------------------------|-------------------------------|
| QB4 | QD8 Config (BaseLine) | 4x16c8_neoni8mm   | 208.77                     | 172.806                       |
|     |                       | 1x16c8_neoni8mm   | 118.26                     | 90.3884                       |
|     | QP8 Config (8x4)      | 8x4c16s2_neoni8mm | 249.26                     | 215.159                       |
|     |                       | 1x4c16s2_neondot  | 80.9344                    | 66.7615                       |
|     | QP8 Config (4x8)      | 4x8c16s2_neoni8mm | 236.42                     | 199.24                        |
|     |                       | 1x4c16s2_neondot  | 100.383                    | 80.204                        |

While when MR > 1 the QP8 Microkernels vastly out perform the QD8 microkernels, we notice that for MR=1 the performance drops considerably. Which explains why the llama performance of the QP8 Kernels show most regressions on performance of token generation rather than prefill. I wanted to compare this against the QP8 Kernels and found the following:

|     |                       |                   | MC:128/NC:16/KC:1024/ | MC:128/NC:32000/KC:4096/ |
|-----|-----------------------|-------------------|----------------------------|-------------------------------|
| QC4 | QD8 Config (BaseLine) | 4x16c8_neoni8mm   | 353.66                     | 287.92                        |
|     |                       | 1x16c8_neoni8mm   | 161.632                    | 109.536                       |
|     | QP8 Config (8x8)      | 8x8c16s2_neoni8mm | 495.69                     | 414.648                       |
|     |                       | 1x8c16s2_neondot  | 132.685                    | 102.366                       |

Here we notice that for MR > 1 the QP8 Kernels outperform the QD8 by a much wider margin. ~42% And the the MR=1  Kernels regress by ~9.5%. Compared to the Blockwise Kernels, we see MR > 1 outperform by ~22%(8x4) and ~14%(4x8) but regress by ~24%(8x4) and ~13%(4x8).

It seems like to improve performance of QP8 Kernels We will need to address the perf regression for kernels where MR=1

digantdesai:
@mcr229,

For kernels with MR > 1: let's look deeper at these numbers. Getting lower tokens per second for prefill (encode) compared to decode (mr=1) doesn't align well with my expectations. 

For kernels with MR = 1: it does seems like QP8 is slower than QD8. Can you compare against 1x8 QD8 kernels to make sure both gemv uses same number of neon pipes from the Cortex-X cores on S22, it could be I haven't looked at Kleidi kernel so correct if this is already apples-apples. Also seems it is also slower for QC4, I wonder if this is a known issue.

mcr229:
> @mcr229,
> 
> For kernels with MR > 1: let's look deeper at these numbers. Getting lower tokens per second for prefill (encode) compared to decode (mr=1) doesn't align well with my expectations.
> 
> For kernels with MR = 1: it does seems like QP8 is slower than QD8. Can you compare against 1x8 QD8 kernels to make sure both gemv uses same number of neon pipes from the Cortex-X cores on S22, it could be I haven't looked at Kleidi kernel so correct if this is already apples-apples. Also seems it is also slower for QC4, I wonder if this is a known issue.

@digantdesai I was able to get more accurate numbers and updated them above.

mcr229:
@alankelly this is ready for review now thank you!

src/operators/fully-connected-nc.c
```diff
@@ -822,6 +833,130 @@ enum xnn_status xnn_create_fully_connected_nc_qp8_f32_qc4w(
...
+
+  if (kernel_zero_point != 8) {
+    xnn_log_error(
+      "failed to create %s operator with %" PRIu8 " kernel zero point: kernel zero point must be equal to 8 "
+      "(unsigned weights) or 0 (signed weights)",
+      xnn_operator_type_to_string(xnn_operator_type_fully_connected_nc_qp8_f32_qb4w), kernel_zero_point);
+    return xnn_status_invalid_parameter;
+  }
+  // Assuming kernel_scale.size() is output_channels * num_blocks.
+  size_t num_blocks = input_channels / block_size;
```
digantdesai, line 909: can we share this with qd8-f32-qb4w with moving this to a common fn and passing an arg?

src/operators/fully-connected-nc.c
```diff
@@ -2020,6 +2189,23 @@ enum xnn_status xnn_reshape_fully_connected_nc_qp8_f32_qc4w(
...
 
+enum xnn_status xnn_reshape_fully_connected_nc_qp8_f32_qb4w(
+    xnn_operator_t fully_connected_op, size_t batch_size,
+    pthreadpool_t threadpool) {
+  return reshape_fully_connected_nc(
+      fully_connected_op, xnn_operator_type_fully_connected_nc_qp8_f32_qb4w,
+      batch_size,
+      /*log2_input_element_size=*/0,
+      // Pass 1 byte even though it is half byte, we handle the division via
+      // filter_is_nibble == true.
```
digantdesai, line 2200: remove this since we have it below?

src/packing.cc
```diff
@@ -1676,6 +1677,64 @@ void xnn_pack_kai_qs4_weights_and_biases(
         &kai_params);
   }
 }
+
+size_t xnn_packed_stride_kai_qb4_weights_and_biases(
```
digantdesai, line 1681: should this be in Kleidi?

digantdesai (approved):

src/packing.cc
mcr229: this is a wrapper around the kleidi api

src/packing.cc
```diff
@@ -1676,6 +1677,64 @@ void xnn_pack_kai_qs4_weights_and_biases(
...
+  const uint32_t nr = gemm_config->nr;
+  const uint32_t kr = UINT32_C(1) << gemm_config->log2_kr;
+  const uint32_t sr = UINT32_C(1) << gemm_config->log2_sr;
+  const struct xnn_qs8_qc4w_packing_params* xnn_params =
+      reinterpret_cast<const struct xnn_qs8_qc4w_packing_params*>(params);
+
+  if (flags & XNN_FLAG_TRANSPOSE_WEIGHTS) {
+    // no nxk as of now
+    xnn_log_fatal(
+      "KleidiAI does not currently have gio packing routine"
```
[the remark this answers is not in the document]
mcr229: it doesn't work unforutnately... They have a function name in the kxn packing header that is the same as another packing functions. So it is causes compiler errors.

src/operator-run.c
```diff
@@ -527,6 +527,33 @@ void xnn_compute_qp8gemm(
                           nr_block_start, mr_block_size, nr_block_size);
 }
 
+void xnn_compute_hmp_qp8gemm_bl(
```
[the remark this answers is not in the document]
mcr229: i kept this pattern because we have differentiate dq_gemm and dq_gemm_bl. I don't want to break that pattern here.

src/operators/fully-connected-nc.c
mcr229: We do the same as qp8_qc4w. While I know that qc4 doesn't necessarily have this extra code, I don't want to break off from how other create fns are doing this.

src/qp8-f32-qb4w-gemm/qp8-f32-qb4w-gemm-minmax-1x4c16s2-aarch64-neondot.c
```diff
@@ -0,0 +1,32 @@
...
+    const void* rhs_packed, float* dst, size_t dst_stride_row,
+    size_t dst_stride_col,
+    const struct xnn_f32_qb4w_minmax_params
+        minmax_params[XNN_RESTRICT XNN_MIN_ELEMENTS(1)]) {
+#if XNN_ENABLE_KLEIDIAI
+  kai_run_matmul_clamp_f32_qai8dxp1x8_qsi4c32p4x8_1x4x32_neon_dotprod(
+      m, n, k, minmax_params->scalar.blocksize, lhs_packed, rhs_packed, dst, dst_stride_row, dst_stride_col,
+      minmax_params->scalar.min, minmax_params->scalar.max);
+#else
+  xnn_log_fatal(
```
alankelly, line 28: Please replace with an assert like in https://github.com/google/XNNPACK/commit/211291c864f0db9b78d7e4439a1d577a672e6410

Some linkers fail with a duplicate symbol if we use logging in the kernels.

alankelly:
> @mcr229,
> 
> For kernels with MR > 1: let's look deeper at these numbers. Getting lower tokens per second for prefill (encode) compared to decode (mr=1) doesn't align well with my expectations.
> 
> For kernels with MR = 1: it does seems like QP8 is slower than QD8. Can you compare against 1x8 QD8 kernels to make sure both gemv uses same number of neon pipes from the Cortex-X cores on S22, it could be I haven't looked at Kleidi kernel so correct if this is already apples-apples. Also seems it is also slower for QC4, I wonder if this is a known issue.

src/operator-run.c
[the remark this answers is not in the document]
gonnet: Given that the functions are identical to `xnn_compute_qp8gemm` and `xnn_compute_hmp_qp8gemm`, respectively, I would just use the existing ones.

We can clean up and/or unify the different `*_gemm` functions later.

src/operators/fully-connected-nc.c
```diff
@@ -1780,31 +1922,58 @@ static enum xnn_status reshape_fully_connected_nc(
     if (xnn_is_hmp_gemm_ukernel(gemm_ukernel)) {
       fully_connected_op->compute[0].type = xnn_parallelization_type_2d_tile_2d_with_uarch;
       if (dynamic_quantization) {
-        fully_connected_op->compute[0].task_2d_tile_2d_with_id = (pthreadpool_task_2d_tile_2d_with_id_t) xnn_compute_hmp_dqgemm;
+        if (is_blockwise_kernel) {
+          fully_connected_op->compute[0].task_2d_tile_2d_with_id = (pthreadpool_task_2d_tile_2d_with_id_t) xnn_compute_hmp_dqgemm_bl;
+        } else {
+          fully_connected_op->compute[0].task_2d_tile_2d_with_id = (pthreadpool_task_2d_tile_2d_with_id_t) xnn_compute_hmp_dqgemm;
+        }
```
gonnet, line 1929: Actually, I would remove the `xnn_compute_hmp_dqgemm_bl` function and just use `xnn_compute_hmp_dqgemm` to avoid this special-casing here and below.

src/operators/fully-connected-nc.c
gonnet: I think this line should be read with the comment line above it, which makes it quite useful.

src/packing.cc
```diff
@@ -1676,6 +1677,64 @@ void xnn_pack_kai_qs4_weights_and_biases(
...
+
+  const size_t kai_num_bytes_sum_rhs = sizeof(float);
+  const size_t kai_num_bytes_bias = sizeof(float);
+  // perhaps derive Bf16 from gemm-config?
+  // This needs to be updated in the kleidi branch to be in header
+  // return kai_rhs_packed_stride(k, /*nr=*/1, kr, block_size, Bf16);
+  const size_t num_bytes_multiplier_rhs = sizeof(uint16_t);
+  const size_t num_blocks_per_row = k/block_size;
+  const size_t num_bytes_per_block = (block_size / 2) + num_bytes_multiplier_rhs;
+  return 1 * ((num_bytes_per_block * num_blocks_per_row) + kai_num_bytes_sum_rhs + kai_num_bytes_bias);
```
gonnet, line 1694: Is there no corresponding `kai_get_rhs_packed_stride_rhs_pack_kxn_*` that this should call instead of doing the computation itself?

The problem is that this code assumes that the KleidiAI implementation will never change.

src/subgraph/convert.c
```diff
@@ -530,7 +530,8 @@ enum xnn_status xnn_define_convert(
   if ((flags & XNN_FLAG_MAYBE_PACK_FOR_GEMM) &&
       input_value->datatype == xnn_datatype_fp32 &&
       output_value->datatype == xnn_datatype_qdint8 &&
-      xnn_init_qp8_f32_qc4w_gemm_config() != NULL) {
+      (xnn_init_qp8_f32_qc4w_gemm_config() != NULL ||
```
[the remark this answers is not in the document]
gonnet: Also, I think the logic is not correct since the `xnn_init_qp8_f32_qc4w_gemm_config()` and `xnn_init_qp8_f32_qb4w_gemm_config` would have to be matched to the `input_datatype`.

test/gemm-microkernel-tester.cc
```diff
@@ -1838,6 +1839,179 @@ void GemmMicrokernelTester::Test(
...
+    const size_t input_packed_size =
+        xnn_x8_packq_f32qp8_packed_size(m(), k2, mr_packed(), kr(), sr());
+    std::vector<int8_t> input_qp8(input_packed_size);
+    xnn_x8_packq_f32qp8_ukernel__scalar_u1(
+      m(), k2, mr_packed(), kr(), sr(),
+      /*m_idx_start=*/0, reinterpret_cast<const float*>(input_f32.data()),
+      /*lhs_stride=*/k2 * sizeof(float),
+      input_qp8.data()
+    );
+
```
gonnet, line 1905: Remove double newline.

mcr229:
@alankelly @gonnet 

Thank you for your comments, I addressed all of them in the commits above, please take a look. Thank you!

src/packing.cc
digantdesai: this is much better. Thanks.

gonnet (approved):
Thank you for contributing!

src/operators/fully-connected-nc.c
```diff
@@ -1825,8 +1962,8 @@ static enum xnn_status reshape_fully_connected_nc(
       if (dynamic_quantization) {
         fully_connected_op->compute[0].task_2d_tile_2d = (pthreadpool_task_2d_tile_2d_t) xnn_compute_dqgemm;
       } else if (is_qp8_ukernel) {
-        fully_connected_op->compute[0].task_2d_tile_2d =
-            (pthreadpool_task_2d_tile_2d_t)xnn_compute_qp8gemm;
+          fully_connected_op->compute[0].task_2d_tile_2d =
```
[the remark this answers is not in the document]
mcr229: fixed! Thank you!

mcr229:
hi @gonnet @alankelly I noticed that i had some errors in calculating stride, because we want the stride for nr =1. kleidi has a constraint which enforces nr can not be 1. So I modified the calculation to scale down the stride calculation by nr. Please take a look, and let me know if its possible to include these changes into the merge.

Thanks!
Max

mcr229:
merged in so closing

PR #7030, opened 2024-09-05
Add RVV qs8/qu8 vlrelu kernels, configs and tests.

oliIMG:


test/qs8-vlrelu.yaml
dsharlet, line 1: Can you revert these (accidentally?) added files? we're trying to move away from yaml to describe kernels that exist in favor of the headers (which you've correctly modified).

oliIMG: I'm sorry for this, this got applied when patching, will revert these changes.

dsharlet (approved):

KaustubhIMG (approved):

PR #7056, opened 2024-09-09
Update CMakeLists to avoid compile flags on top-level interface

GregoryComer:
A previous change (https://github.com/google/XNNPACK/commit/066628d7dd9a804d0a98c22069c32035f15d719e) appears to have inadvertently caused the XNNPACK target to apply compiler flags on the xnnpack-base interface (-O2) to targets that depend on XNNPACK. This change makes xnnpack-base private to avoid putting compiler flags on the public interface. It also explicitly links pthreadpool as PUBLIC on the XNNPACK to mimic the previous behavior.

I believe that since xnnpack-base is an interface target and part of the link interface for XNNPACK, all targets that link against XNNPACK also depend on xnnpack-base, and thus the compiler flags. It is possible to repro this by creating an empty main.cpp and a basic CMakeLists.txt top-level target with xnnpack as a subdirectory. Note that when building without depending on xnnpack, no -O2 is passed to the compiler. When adding xnnpack as a dependency, the top-level target is compiled with -O2.


gonnet:
Good catch, thanks for fixing this!

PR #7081, opened 2024-09-12
Add support for RVV f32 vrnd microkernels and config.

KaustubhIMG:
Microkernel for RNDNE, RNDZ, RNDU and RNDD are added. The LMUL variations are 1, 2, 4 and 8.
The config is added for LMUL 4 version.


KaustubhIMG:
@fbarchard @bhbruce @gonnet, If you have time, please help review this PR.

dsharlet (approved):

dsharlet:
Unfortunately, the approach taken in this PR doesn't work. We're trying to improve our test coverage, and this microkernel fails because converting to int and back only works for floats that are in the range of an integer. I tried a bit to make these kernels work (https://github.com/google/XNNPACK/pull/7971), but at this point I think they need a full rewrite with a different approach.

PR #7096, opened 2024-09-16
Fix f32-qs8/qu8 vcvt rvv kernel.

KaustubhIMG:
Do zero_point calculation before min/max


KaustubhIMG:
@dsharlet added the fix.

fbarchard (approved):

PR #7185, opened 2024-09-25
Follow Up Patch for Kleidi Integration PR #7003

mcr229:
There is a bug in stride calculations. We need to scale the stride down by nr. Adding the fix and corresponding test here.


mcr229:
cc. @alankelly @gonnet for review

alankelly (approved):

alankelly (approved):

mcr229:
@alankelly fixed on x86 and likely all other ISAs

PR #7192, opened 2024-09-26
Xnn f32 reduce window

LokeshReddyOVS-MCW:


src/f32-rwsum/f32-rwdsum-1p1x-scalar-c1.c
```diff
@@ -0,0 +1,79 @@
...
+    const struct xnn_f32_default_params params[XNN_RESTRICT XNN_MIN_ELEMENTS(1)])
+{
+  assert(rows != 0);
+  assert(channels != 0);
+  assert(input != NULL);
+  assert(output != NULL);
+
+  int padded_size = rows + MAX((rows - 1),0) * (base_dilation - 1) + padding[0] + padding[1];
+  int output_size = (padded_size < (window_dimensions - 1) * window_dilations + 1) ? 
+                    0 : FLOOR((padded_size - (window_dimensions - 1) * window_dilations - 1) / (float)window_strides) + 1;
```
dsharlet, line 41: This is going to convert the size to a float, which could drop some precision, which is a major problem for size/shape values. I think this needs to be done with integer arithmetic.

src/f32-rwsum/f32-rwdsum-1p1x-scalar-c1.c
```diff
@@ -0,0 +1,79 @@
...
+
+  // replaced modulo and division by multiplicative scaled inverse
```
dsharlet, line 43: This is an approximation, isn't this a problem for something like this? Similar to float comment above: approximation for the "values" of buffers is usually OK, but not for the "shape" parameters.

src/f32-rwsum/f32-rwdsum-1p1x-scalar-c1.c
```diff
@@ -0,0 +1,79 @@
...
+  int64_t inverse_base_dilation = (1LL << 32) / base_dilation;
+  int64_t inverse_win_dilation = (1LL << 32) / window_dilations;
+
+  for (int i = 0; i < output_size; i++) {
+        float sum = init_value;
+        int window_start = i * window_strides;
+        int pad_high_boundary = CEIL((((padding[0] - window_start) * inverse_win_dilation) >> 32));
+        int pad_low_boundary = CEIL((((padded_size - padding[1] - window_start) * inverse_win_dilation) >> 32));
```
dsharlet, line 51: Isn't the argument to CEIL here an integer? I think you need something more like: `(((padded_size - padding[1] - window_start) * inverse_win_dilation + ((1 << 32) - 1)) >> 32)`

But of course, this assumes it's OK to approximate this division in the first place (see above comment).

src/f32-rwsum/f32-rwdsum-1p1x-scalar-c1.c
```diff
@@ -0,0 +1,79 @@
...
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include <assert.h>
+
+#include "xnnpack/common.h"
+#include "xnnpack/reduce.h"
+
+#define CEILING_POS(X) ((X-(int)(X)) > 0 ? (int)(X+1) : (int)(X))
```
dsharlet, line 11: These macros should be helper functions in math.h. I think we already have some of the functions you need, e.g. `divide_round_up` (we shouldn't divide first and then try to compute the ceiling of the result, that approach assumes approximating the result with float or fixed point arithmetic).

src/f32-rwsum/f32-rwdsum-1p1x-scalar-c1.c
```diff
@@ -0,0 +1,79 @@
...
+#define CEILING_NEG(X) (int)(X)
+#define CEIL(X) ( ((X) > 0) ? CEILING_POS(X) : CEILING_NEG(X) )
+#define FLOORING_POS(X) (int)(X)
+#define FLOORING_NEG(X) ((X-(int)(X)) > 0 ? (int)(X-1) : (int)(X))
+#define FLOOR(X) ( ((X) > 0) ? FLOORING_POS(X) : FLOORING_NEG(X) )
+#define MAX(X,Y) (X > Y ? X : Y)
```
dsharlet, line 17: Use functions in math.h for these

src/f32-rwsum/f32-rwdsum-1p1x-scalar-c1.c
```diff
@@ -0,0 +1,79 @@
...
+#define MIN(X,Y) ((X) < (Y) ? (X) : (Y))
+
+// reduce window for vertical reduction 
+void xnn_f32_rwdsum_ukernel_1p1x__scalar_c1(
+    size_t rows,
+    size_t channels,
+    const float* input,
+    float init_value,
+    int* padding, 
+    int base_dilation,
```
dsharlet, line 27: Do these need to be `int`, i.e. signed? If we can assume they are positive, a lot of the necessary arithmetic gets simpler (e.g. `divide_round_up` is pretty simple for `size_t`, not so much for `int`).

test/f32-rwdsum.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright 2024 Google LLC
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# SCALAR
+- name: xnn_f32_rwdsum_ukernel_1p1x__scalar_c1
```
dsharlet, line 7: Please don't add new yaml files, and instead try to use this new system: https://github.com/google/XNNPACK/blob/master/doc/microkernel-enumerators.md, e.g. https://github.com/google/XNNPACK/blob/master/src/f16-vabs/f16-vabs.h.

We are working on converting existing yamls to such headers.

dsharlet (commented):
I only looked at microkernels so far

src/f32-rwsum/f32-rwdsum-1p1x-scalar-c1.c
```diff
@@ -0,0 +1,79 @@
...
+    int window_strides,
+    float* output,
+    const struct xnn_f32_default_params params[XNN_RESTRICT XNN_MIN_ELEMENTS(1)])
+{
+  assert(rows != 0);
+  assert(channels != 0);
+  assert(input != NULL);
+  assert(output != NULL);
+
+  int padded_size = rows + MAX((rows - 1),0) * (base_dilation - 1) + padding[0] + padding[1];
```
dsharlet, line 39: The mixing of int and size_t here is concerning, I think there's a lot of risk of integer overflow, especially in the approximated division below which will multiply values that should be size_t with very large fixed point reciprocals (2^32 if dilation/stride is 1).

PR #7211, opened 2024-10-01
Kleidi introduce new 16x4 kernels

mcr229:
Kleidi is adding new 16x4 optimized kernels here with specialized groupsize:

https://gitlab.arm.com/kleidi/kleidiai/-/merge_requests/117/diffs#ca37d1189498b4e5ccb6f6dcb9e43501a28cb5c6

We see some good performance, so bumping the kleidi pin and adding them here


src/configs/gemm-config.c
```diff
@@ -1741,11 +1741,11 @@ static void init_qp8_f32_qb4w_gemm_config(void) {
     if (XNN_ENABLE_ARM_I8MM && hardware_config->use_arm_neon_i8mm) {
       #if XNN_ENABLE_ARM_I8MM
         qp8_f32_qb4w_gemm_config.minmax.qp8gemm_bl[XNN_MR_TO_INDEX(1)] = xnn_init_hmp_qp8gemm_bl_ukernel(xnn_qp8_f32_qb4w_gemm_minmax_ukernel_1x4c16s2__aarch64_neondot);
-        qp8_f32_qb4w_gemm_config.minmax.qp8gemm_bl[XNN_MR_TO_INDEX(8)] = xnn_init_hmp_qp8gemm_bl_ukernel(xnn_qp8_f32_qb4w_gemm_minmax_ukernel_8x4c16s2__neoni8mm_mstep2);
+        qp8_f32_qb4w_gemm_config.minmax.qp8gemm_bl[XNN_MR_TO_INDEX(16)] = xnn_init_hmp_qp8gemm_bl_ukernel(xnn_qp8_f32_qb4w_gemm_minmax_ukernel_16x4c16s2__neoni8mm_mstep4);
```
digantdesai, line 1744: This is OK for now. 

For future, I am not sure if this will be the best tile size for the budget phones or slightly older phones with i8mm. Perhaps in the future we can use cpu specific tile size i.e. larger on Cortex-x cores and smaller on mid cores like we do in other hmp cases?

src/packing.cc
```diff
@@ -1631,7 +1631,7 @@ size_t xnn_packed_stride_kai_qs4_weights_and_biases(
     size_t extra_bytes) {
   const uint32_t kr = UINT32_C(1) << gemm_config->log2_kr;
   const uint32_t sr = UINT32_C(1) << gemm_config->log2_sr;
-  return kai_get_rhs_packed_stride_rhs_pack_kxn_qsi4cxp_qsu4cxs1s0(k, /*nr=*/1,
+  return kai_get_rhs_packed_stride_rhs_pack_kxn_qsi4cxp_qs4cxs1s0(k, /*nr=*/1,
```
digantdesai, line 1634: Just a rename? No change in the zp==8 and packing 2 unsigned values?

mcr229: yea this seems like purely a renaming. They no longer assert zero_point==8, but we pass that in anyways. So I believe this is purely a renaming for us. Packing 2 unsigned values stays the same

https://gitlab.arm.com/kleidi/kleidiai/-/commit/49f0ca8b20505e9a5a2c1449a8f715b587593280

for reference

mcr229:
@alankelly @gonnet

src/packing.cc
digantdesai: Nice. Thanks for fixing this.

WORKSPACE
```diff
@@ -112,10 +112,10 @@ http_archive(
 # KleidiAI library, used for ARM microkernels.
 http_archive(
     name = "KleidiAI",
-    sha256 = "d8f2b5bf6eba7ab8fe3cedd97c4adc967c1befa69a6f4c4f6cbb3c102a7dd3c9",
+    sha256 = "2666071382cfe1a705aaf96d938bca4436db59fc28518e49899140b652fa9a17",
     strip_prefix = "kleidiai-32384cde728f444afdb92eecbb65e293fc6a6315",
```
gonnet, line 116: This needs to be updated as well.

Also, would it make more sense to wait for the next official release, which I think should be happening soon?

mcr229: thanks i'll update that.

I'm not supre familiar with their release plan/schedule. Do you know how soon it is? We can wait if its coming up soon (next week-ish)

i'll be on PTO next week, so i'll cc @digantdesai @GregoryComer to bring this to the finish line.

digantdesai: Asked Gian Marco from Arm. Worst case can we update again? We would like to pull this kernel soon-ish.

digantdesai: Ok https://gitlab.arm.com/kleidi/kleidiai/-/releases/v0.3.0 is out, let me use that.

digantdesai: Created a new PR #7238 on top of this. Here is a [delta](https://github.com/mcr229/XNNPACK/compare/kleidi_p2..digantdesai:XNNPACK:kleidi_pin_bump_v0.3.0?expand=1) (ignore rebased stuff)

PR #7238, opened 2024-10-08
Pull in a new Kleidi kernel for QB4W

digantdesai:
This is reincarnation of Max's PR #7211 with a different pin.


digantdesai:
cc @alankelly  @gonnet - this replaces #7211

alankelly (approved):

Issue #7206, opened 2024-10-01
Unable to Enable Debug Logging in XNNPACK Despite Setting `XNN_LOG_LEVEL=4` and `-DCMAKE_BUILD_TYPE=Debug`


wewe5215:
Hi XNNPACK Developers,

I'm experiencing an issue where I'm unable to see debug messages from XNNPACK despite setting `XNN_LOG_LEVEL=4` during the build process. I've attempted this on both macOS (arm64) and Ubuntu(amd64) platforms, but the debug messages do not appear in either case.

**Problem Description:**

I am trying to enable debug logging in XNNPACK to understand which microkernels are being selected during the execution of my program. Despite setting `XNN_LOG_LEVEL=4` and rebuilding the library and my application multiple times, I do not see any debug messages from XNNPACK when running my application.

**Steps Taken to Resolve the Issue:**

1. **Built XNNPACK in Debug Mode:**
   - Used `-DCMAKE_BUILD_TYPE=Debug` during the CMake configuration.
   - Confirmed that the build configuration is set to `Debug`.
2. **Set `XNN_LOG_LEVEL=4` During Build:**
   - Modified the build script to include `-DXNN_LOG_LEVEL=4` in `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS`.
   - Verified via verbose build output that `-DXNN_LOG_LEVEL=4` is being passed to the compiler.
3. **Ensured No Conflicting Definitions:**
   - Removed any custom definitions of `XNN_LOG_LEVEL` in my application code.

4. **Checked for Logging Symbols in the Library:**
   - Ran `nm libXNNPACK.a | grep xnn_log_debug` and found the following:

     ```
     0000000000000c98 t _xnn_log_debug_
     00000000000004e8 t _xnn_log_debug_
     ...
     ```

   - Noted that the symbols are local (`t`) and have underscores, which seems unusual.

5. **Used a Debugger:**
   - Attempted to set breakpoints at `xnn_log_debug`, but the debugger cannot detect the symbol.
   - Stepped through the code and confirmed that logging functions are not being called.

6. **Modified `CMakeLists.txt`:**
   - Added `add_definitions(-DXNN_LOG_LEVEL=4)` to force the log level.
   - Modified `TARGET_COMPILE_DEFINITIONS` to always set `XNN_LOG_LEVEL=4`.
   - Rebuilt the library and application, but still no debug messages appear.

7. **Checked Output Streams and Buffering:**
[...]

PR #7270, opened 2024-10-15
fix:Enable debug logging in XNNPACK by setting XNN_LOG_LEVEL to 5 in Debug builds

wewe5215:
- Addresses issue #7206 (https://github.com/google/XNNPACK/issues/7206)


CMakeLists.txt
```diff
@@ -788,8 +788,8 @@ IF(XNNPACK_BUILD_LIBRARY)
   TARGET_LINK_LIBRARIES(operator-run PRIVATE xnnpack-base logging)
   TARGET_LINK_LIBRARIES(operator-utils PRIVATE xnnpack-base logging)
   TARGET_LINK_LIBRARIES(subgraph PRIVATE xnnpack-base allocator logging memory mutex operators operator-run)
-  TARGET_LINK_LIBRARIES(XNNPACK PRIVATE xnnpack-base allocator cache hardware-config indirection logging memory microkernel-utils microparams-init mutex normalization operators operator-run operator-utils packing microkernels-prod subgraph)
-  TARGET_LINK_LIBRARIES(XNNPACK PUBLIC pthreadpool)
+  TARGET_LINK_LIBRARIES(XNNPACK PRIVATE allocator cache hardware-config indirection memory microkernel-utils microparams-init mutex normalization operators operator-run operator-utils packing microkernels-prod subgraph)
+  TARGET_LINK_LIBRARIES(XNNPACK PUBLIC xnnpack-base logging)
```
gonnet, line 792: Does this also work if you do
```
TARGET_LINK_LIBRARIES(XNNPACK PUBLIC pthreadpool logging)
```
i.e. leave `xnnpack-base` in the `PRIVATE` linkeage statement above?

Asking because having `xnnpack-base` here was causing problems (and was recently removed in #7056) since it propagates all the associated compiler flags with it.

gonnet (changes requested):

CMakeLists.txt
wewe5215: Yes, that worked for me as well! I’ve made a new commit to reflect this fix. Thank you for the suggestion!

PR #7311, opened 2024-10-22
Enable AVXVNNI microkernels for mobile platforms

soniravi31:
AlderLake onwards x86 mobile platform supports avxvnni


soniravi31:
@fbarchard Please help to take a look.
Thanks.

PR #7321, opened 2024-10-23
Resolve the CMake EXECUTE_PROCESS Python Error on Windows

chenghuaWang:
When generating `microkernels.cmake` on Windows, Python3 is not found. We assume that on Windows, there is no conflict between different versions of Python, such as Python2 and Python3.


chenghuaWang:
@mcr229  https://github.com/google/XNNPACK/pull/7429 has solved this problem.

PR #7367, opened 2024-10-28
wrap get_mr and get_nr with #ifdef kleidi

mcr229:
Currently fails to build when XNN_ENABLE_KLEIDIAI=OFF. Need to wrap these helper methods with the #ifdef kleidiai


mcr229:
@gonnet @alankelly

alankelly (approved):

PR #7376, opened 2024-10-29
Fix build error

RahulSundarMCW:
- This fix addresses the build error 'No such file or directory' that occurs after generating the tests.
  - Removes  {xnncommon.xnnpack_src()} from tests += f'#include "{xnncommon.xnnpack_src()}/{folder}/{options.ukernel}.h"\n' from
 1. generate-vbinary-test.py
 2. generate-vunary-test.py
 3. generate-dwconv-unipass-test.py
 4. generate-dwconv-multipass-test.py

### DIFF
--- a/tools/generate-dwconv-multipass-test.py
+++ b/tools/generate-dwconv-multipass-test.py
@@ -399,7 +399,7 @@ def main(args):
       },
   ))
 
-  tests += f'#include "{xnncommon.xnnpack_src()}/{folder}/{options.ukernel}.h"\n'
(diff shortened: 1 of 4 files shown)


dsharletg:
I'm confused, if I run `generate-tests.sh`, I don't see any change to the code in the repo.

Can you build and run the tests in a clean version of the XNNPACK repo?

dsharletg:
Thanks for pointing out the problem. Unfortunately, these scripts interact with our internal version of the repo in a way that makes it hard to fix externally. I've sent https://github.com/google/XNNPACK/pull/7383 which applies the same fix slightly differently.

PR #7408, opened 2024-11-04
Replace vscaleextexp yaml with header

RahulSundarMCW:


scripts/generate-tests.sh
```diff
@@ -273,7 +273,7 @@ tools/generate-raddstoreexpminusmax-test.py --spec test/f16-raddstoreexpminusmax
 tools/generate-raddstoreexpminusmax-test.py --spec test/f32-raddstoreexpminusmax.yaml --output test/f32-raddstoreexpminusmax.cc &
 
 ### Tests for VScaleExtExp micro-kernels
-tools/generate-vscaleextexp-test.py --spec test/f32-vscaleextexp.yaml --output test/f32-vscaleextexp.cc &
+tools/generate-vscaleextexp-test.py --tester VScaleExtExpMicrokernelTester --ukernel f32-vscaleextexp --output test/f32-vscaleextexp.cc &
```
dsharlet, line 276: There's only one test using this, and now that it doesn't actually generate much code, we should just remove it. Let's also move the contents of vscaleextexp-microkernel-test.* to test/f32-vscaleextexp.cc as well.

test/f32-vscaleextexp.cc
```diff
@@ -4,7 +4,7 @@
 // LICENSE file in the root directory of this source tree.
 //
 // Auto-generated file. Do not edit!
```
dsharlet, line 6: When you remove the generator script, also remove this comment, because this file will not be generated code any more.

RahulSundarMCW:
Address the review comments and pushed the commit.

tools/generate-vscaleextexp-test.py
dsharletg, line 1: Delete this file

dsharlet (commented):
I just noticed something missing for all these yaml -> header PRs: there are declarations of the microkernels that should be converted to use the header tables, e.g.: https://github.com/google/XNNPACK/blob/060cece66687b44685c646c55856353698e67934/src/xnnpack/vscaleexpminusmax.h#L17.

This is important to help validate that the header tables are correct.

src/f32-vscaleextexp/f32-vscaleextexp.h
```diff
@@ -0,0 +1,56 @@
...
+#endif
+
+#ifndef XNN_UKERNEL
+#define XNN_UKERNEL(arch_flags, ukernel, element_tile,  datatype) \
+    XNN_UKERNEL_WITH_PARAMS(arch_flags, ukernel, element_tile,  datatype, void, /*init_params=*/nullptr)
+#define XNN_DEFINED_UKERNEL
+#endif
+
+#if XNN_ARCH_X86 || XNN_ARCH_X86_64
+XNN_UKERNEL_WITH_PARAMS(xnn_arch_x86_avx2, xnn_f32_vscaleextexp_ukernel__avx2_p5_u8, 8, float, struct xnn_f32_default_params, xnn_f32_vscaleextexp_ukernel_fn)
```
dsharlet, line 19: Why is `xnn_f32_vscaleextexp_ukernel_fn` here in the last parameter? I'd expect it to be NULL.

dsharlet (approved):

PR #7429, opened 2024-11-06
use Python_EXECUTABLE to generate microkernels.cmake

mcr229:
Help makes the CMake a little more robust by using the find_package python for build time python rather than straight python3.


mcr229:
@alankelly @gonnet @dsharlet

mcr229:
@fbarchard as well for review

alankelly (approved):

PR #7433, opened 2024-11-06
install microkernels-prod along with XNNPACK

mcr229:
Since microkernels prod has been changed into a static library instead of an object library, there are instances in external projects where libmicrokernels-prod still has to be explicitly linked along with libXNNPACK in order for the microkernels to be properly linked into the binary. As a result, it would be nice if XNNPACK could install microkernels-prod along with XNNPACK since they are both needed.


mcr229:
@alankelly @dsharlet @fbarchard @gonnet

dsharletg:
Should we/can we just undo the change motivating this? It does seem nicer to have one self contained installed target.

mcr229:
> Should we/can we just undo the change motivating this? It does seem nicer to have one self contained installed target.

I would say yes, but I don't know what the original motivation behind changing microkernels-prod to a static library was.

PR #7508, opened 2024-11-25
Fix suggestions for Hexagon to accommodate recent changes

ejparkqc:
- Fixed the error where Hexagon build assumes int32_t same as long data type.


dsharlet:
I think you need to get rid of all the `int` overloads of the functions you added, i.e. replace `int` with `int32_t` rather than add an overload.

ejparkqc:
I found my changes will break other target builds when they treat int same as int32_t (e.g., X86 and other archs). Working on it.

dsharlet (approved):

PR #7548, opened 2024-12-03
Fix c8 packw overread for wasmdot

yolanda15:
Also early load packed_b to avoid an unexpected load rescheduling by compiler and make it revectorizable by V8.


gonnet (approved):

fbarchard (approved):
Looks good with only 2 nits
remove OOB_READS
Schedule all the loads, followed by all the blends, if registers allow.   Allows them to co-issue/register rename

PR #7640, opened 2025-01-04
Prepare for LoongArch SX and ASX

MQ-mengqing:
1, Add an option named XNNPACK_ENABLE_CPUINFO so that avoid compiling cpuinfo for LoongArch (and Hexagon).
2, Add "&& XNN_ARCH_RISCV" in transpose-config for RISCV. 
3, Others are prepared for LoongArch SX and ASX.

Cc: @lixing-star @junchao-loongson


fbarchard (approved):

MQ-mengqing:
Gentle ping :smile:. Is there anything I need to do? @fbarchard

src/configs/transpose-config.c
```diff
@@ -197,7 +197,7 @@ static void init_transpose_config(void) {
       .const_size_ukernel = (xnn_transposec_ukernel_fn) xnn_x24_transposec_ukernel__1x2_scalar,
       .tile_size = 32,
     };
-    #if XNN_ENABLE_RISCV_VECTOR
+    #if XNN_ENABLE_RISCV_VECTOR && XNN_ARCH_RISCV
```
fbarchard, line 200: this change seems unrelated and unnecessary, since RISCV_VECTOR is only on RISCV?

MQ-mengqing: Because the XNNPACK_ENABLE_RISCV_VECTOR set default ON in CMakeLists.txt which csuse XNN_ENABLE_RISCV_VECTOR is true if we don't set it.
https://github.com/google/XNNPACK/blob/master/CMakeLists.txt#L138

fbarchard: Re riscv   got it.  I've gone ahead and submitted that, thanks

PR #7692, opened 2025-01-16
Remove trailing tab in JSON

ericsciple:


fbarchard (approved):
LG... if you prefer I can land it, but if you want to, theres a CLA to sign

fbarchard:
In your PR I noticed we have an old SDE and missing some new CPUs
So I made a PR that updates SDE and includes your trailing tabs
https://github.com/google/XNNPACK/pull/7744

Its my first attempt at updating this script... if that works, I'll add to the list of cpus

fbarchard:
I've landed this change, thanks

PR #7774, opened 2025-02-04
Spmm minmax

RahulSundarMCW:


fbarchard (approved):
okay looks right, but I wouldnt spend time on spmm.
We only implemented it for arm and wasmsimd and it hasnt been widely used, as it was an improvement over simple floats.  But now we favor QD8 to compress floats down to 4 or 8 bit achieving better performance and quality.

with 4 or 8 bit weights, I'm sure there is potential for something like spmm
technically you could apply the algorithm to 8 bit data, skipping zeros.  But to work well it needs training, and I would think theres a better compression scheme.

RahulSundarMCW:
> okay looks right, but I wouldnt spend time on spmm. We only implemented it for arm and wasmsimd and it hasnt been widely used, as it was an improvement over simple floats. But now we favor QD8 to compress floats down to 4 or 8 bit achieving better performance and quality.
> 
> with 4 or 8 bit weights, I'm sure there is potential for something like spmm technically you could apply the algorithm to 8 bit data, skipping zeros. But to work well it needs training, and I would think theres a better compression scheme.


Thanks, this PR was actually meant to refactor the existing yaml to table headers for spmm-minmax, but looks like we have issues with the CLA checks at the moment so we will rectify this and will reopen this PR.

Sorry for the inconvenience.

PR #7798, opened 2025-02-10
Add QNX to recognized systems

eleir9268:
We build XNNPACK as a dependency of pytorch mobile and tensorflow lite, see https://github.com/qnx-ports/build-files

For QNX SDP 8.0 it doesn't take many changes for the build to pass.
For QNX SDP 7.1 the build will only work with older versions of XNNPACK (because of missing fp16 / bf16 cpu features).


fbarchard (approved):

fbarchard:
Looks good.  You'll need to sign the CLA for us to take the change.
I expect there will be little code cleanup things to fix once you can attempt a build.

The cmake file checks for compilers/versions and disabled things like fp16 and bf16, so you may need to tweak the build if you're running into compiler issues

eleir9268:
> Looks good. You'll need to sign the CLA for us to take the change. I expect there will be little code cleanup things to fix once you can attempt a build.
> 
> The cmake file checks for compilers/versions and disabled things like fp16 and bf16, so you may need to tweak the build if you're running into compiler issues

Thus far I haven't managed to get it working. Is there somewhere I can go for support?

fbarchard:
You can report an issue on xnnpack github for the fp16/bf16 issues
For your QNX enable, its fine... I can land that on its own.

A quick search finds QNX runs on a variety of cpus, but I'm guessing more mobile series ARM?
In this CMakeLists.txt further down you see there are flags to disable parts of build which usually check the compiler version number.
You could add a check for QNX or find a better thing to check on, and disable those ISA that are unsupported by your target

eleir9268:
We support mostly embedded ARM and intel.

I'm still not having any luck on getting the ISA excluded from the build. I'll raise an issue when I have the time.

fbarchard:
for the missing fp16 / bf16 cpu features, we can disable the microkernels.
some specific build errors are needed and what version of compiler you're using

eleir9268:
Hi @fbarchard , I changed the author on the commit. Could you set off the workflow to see if it's fixed? Thanks.

PR #7828, opened 2025-02-18
Enable AVXVNNI microkernels for mobile platforms

soniravi31:
AlderLake onwards x86 mobile platform supports avxvnni

### DIFF
--- a/src/configs/gemm-config.c
+++ b/src/configs/gemm-config.c
@@ -2865,7 +2865,7 @@ static void init_qdu8_f16_qc8w_gemm_config(void) {
       } else
     #endif
     #if XNN_ENABLE_AVXVNNI
-      if (!XNN_PLATFORM_MOBILE && hardware_config->use_x86_avxvnni) {
+      if (hardware_config->use_x86_avxvnni) {
         // AVX VNNI checked before AVX512SKX as it performs better with VNNI microkernels
         qdu8_f16_qc8w_gemm_config.arch = xnn_arch_x86_avxvnni;
         qdu8_f16_qc8w_gemm_config.minmax.dqgemm[XNN_MR_TO_INDEX(1)] = xnn_init_hmp_dqgemm_ukernel((xnn_dqgemm_ukernel_fn) xnn_qd8_f16_qc8w_gemm_minmax_ukernel_1x8c8__avxvnni_prfm);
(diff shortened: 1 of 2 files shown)


src/configs/gemm-config.c
```diff
@@ -2865,7 +2865,7 @@ static void init_qdu8_f16_qc8w_gemm_config(void) {
       } else
     #endif
     #if XNN_ENABLE_AVXVNNI
-      if (!XNN_PLATFORM_MOBILE && hardware_config->use_x86_avxvnni) {
+      if (hardware_config->use_x86_avxvnni) {
```
fbarchard, line 2868: the meaning of 'XNN_PLATFORM_MOBILE' is really iphone and android (arm) devices using intel for emulating iphone when running on x86 mac, or x86 on linux for android studio, to mimic an arm device.

the mac will never have avxvnni... the last x86 mac was broadwell.
a linux or windows development machine could have avxvnni, but its not critical to use kernels for debugging
They are disabled for practical reasons - an old NDK / compiler wont have new instruction sets.

fbarchard: Alderlake should already be enabled for Windows, Linux, Chromebooks etc.
Its just not enabled for iOS or Android

Can you confirm?

fbarchard: oh cool.  For android it makes sense.  At the time it was because iOS and Android are really for ARM devices and the x86 builds were just for the 'Simulator' and Android Studio.
In the past there are old NDK and ios tools that dont have avx512 or modern x86 instruction sets.
You did avxvnni, which is relatively new;  newer than avx512 vnni.
So I agree we should try removing some MOBILE if statements, but proceed with caution

in src/xnnpack/common.h
// Mobile build x86 versions for debugging
#if XNN_PLATFORM_ANDROID || XNN_PLATFORM_IOS
#define XNN_PLATFORM_MOBILE 1
#else
#define XNN_PLATFORM_MOBILE 0
#endif

XNN_PLATFORM_MOBILE is set for iphone as well as android, and removing the if from avxvnni enables it for iphone, which is likely problematic, because the last x86 macos was Intel Broadwell based, and did not have avxvnni or avx512 vnni.

So instead, on any x86 kernels you want to try enabling for android, change 
XNN_PLATFORM_MOBILE to XNN_PLATFORM_IOS ?

The macro XNN_ENABLE_AVXVNNI can be disabled.  With bazel the option 
`-define=xnn_enable_avxvnni=false` 
for android if an old ndk doesnt build, avxvnni can be disabled, either with the bazel build flag, or the CMakeLists.txt check for compiler version.
The last old NDK we supported was 21, which has gcc 9 and clang 9, which do not have avxvnni.

soniravi31: Thanks for update. As per your suggestion I have updated XNN_PLATFORM_MOBILE to XNN_PLATFORM_IOS so that it remain disabled for iPhone.

soniravi31: I have checked that in CMakeList.txt there are already checks available for compiler version which does not support avxvnni / avxvnniint8.

Issue #3455, opened 2022-09-10
Add CMake support for Emscripten builds


dpar39:
It seems that it would be as simple as updating CMake to match the file list in `PROD_SCALAR_WASM_MICROKERNEL_SRCS` etc... from BUILD.bazel

Maratyszcza:
Emscripten's CMake toolchain doesn't support targeting WAsm Atomics / WAsm SIMD / WAsm Relaxed SIMD specifications used in XNNPack, which is the reason why we don't support WebAssembly builds with CMake. If you want to have it supported, I suggest to refactor Emscripten's CMake toolchain and add options to target these WAsm specifications.

dpar39:
Wasm SIMD can be enabled with a compiler flag from what I understand:

    IF(EMSCRIPTEN)
      IF(XNNPACK_ENABLE_WASMSIMD)
        SET_PROPERTY(SOURCE ${PROD_MICROKERNEL_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS " -msimd128 ")
        SET_PROPERTY(SOURCE ${ALL_MICROKERNEL_SRCS} APPEND_STRING PROPERTY COMPILE_FLAGS " -msimd128 ")
        SET_PROPERTY(TARGET XNNPACK APPEND_STRING PROPERTY COMPILE_FLAGS " -msimd128 ")
      ENDIF()
    ENDIF()

Relaxed SIMD and atomics I think are just at the proposal level, but even without simd128 I think it would be nice to support whatever can give a boost in WebAssembly. I currently have [this commit](https://github.com/dpar39/XNNPACK/commit/55fe9e2518dcf6d0143eefa6b87c5e00f0ba8040) on a branch from the reference of tflite 2.10.0 and it's compiling without issues, still haven't tested it though.

Maratyszcza:
WAsm extensions work differently than native extensions in that they "taint" the whole WAsm binary. E.g. if a WAsm binary contains even a single WAsm SIMD instruction, the whole binary will be rejected by WAsm engines which don't support WAsm SIMD. Thus, even though these flags should be set at toolchain level, not at the level of individual projects.

dpar39:
> the whole binary will be rejected by WAsm engines which don't support WAsm SIMD

aka Safari 🤣 

And I don't think wasm works that different to native binaries, all it takes is one shared object with an instruction not supported by the running CPU for the whole process to be killed by illegal instruction.

[...]

PR #7859, opened 2025-02-21
Enable support for Emscripten in CMake

aiusepsi:
Addresses #3455 

Takes the approach of detecting the use of Wasm SIMD by detecting if the requisite flags have been added to CMAKE_C_FLAGS or CMAKE_CXX_FLAGS. Adding the SIMD flags to these variables is the right way to enable Wasm SIMD globally for CMake builds, so this should do the right thing.

The use of `SET(XNNPACK_ENABLE_RISCV_VECTOR OFF)` is to suppress a compile error; without this RISC-specific code is compiled for Wasm; turning this option off here seemed the minimally invasive solution.


CMakeLists.txt
```diff
@@ -89,6 +89,15 @@ ELSEIF(CMAKE_GENERATOR MATCHES "^Visual Studio " AND CMAKE_GENERATOR_PLATFORM)
   ELSE()
     MESSAGE(FATAL_ERROR "Unsupported Visual Studio architecture \"${CMAKE_GENERATOR_PLATFORM}\"")
   ENDIF()
+ELSEIF(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
+  SET(XNNPACK_TARGET_PROCESSOR "wasm")
+  SET(XNNPACK_ENABLE_RISCV_VECTOR OFF)
```
fbarchard, line 94: RISCV in particular?

aiusepsi: Yeah, the `XNNPACK_ENABLE_RISCV_VECTOR` option defaults to ON, which causes the define `XNN_ENABLE_RISCV_VECTOR` to be set, which causes this line: https://github.com/google/XNNPACK/blob/4b857f39ce1ff8454f916e4e3cb0ab0f7f698c46/src/configs/transpose-config.c#L203 to be compiled, which (along with a few others) falls over:

```
XNNPACK/src/configs/transpose-config.c:203:28: error: no member named 'vlenb' in 'struct xnn_hardware_config'
  203 |       if (hardware_config->vlenb >= 128) {
      |           ~~~~~~~~~~~~~~~  ^
XNNPACK/src/configs/transpose-config.c:205:61: error: use of undeclared identifier 'xnn_x32_transposec_ukernel__32x8_rvv'
  205 |           .const_size_ukernel = (xnn_transposec_ukernel_fn) xnn_x32_transposec_ukernel__32x8_rvv,
      |                                                             ^
XNNPACK/src/configs/transpose-config.c:208:35: error: no member named 'vlenb' in 'struct xnn_hardware_config'
  208 |       } else if (hardware_config->vlenb == 64) {
      |                  ~~~~~~~~~~~~~~~  ^
XNNPACK/src/configs/transpose-config.c:210:61: error: use of undeclared identifier 'xnn_x32_transposec_ukernel__16x8_rvv'
  210 |           .const_size_ukernel = (xnn_transposec_ukernel_fn) xnn_x32_transposec_ukernel__16x8_rvv,
      |                                                             ^
XNNPACK/src/configs/transpose-config.c:213:35: error: no member named 'vlenb' in 'struct xnn_hardware_config'
  213 |       } else if (hardware_config->vlenb == 32) {
      |                  ~~~~~~~~~~~~~~~  ^
XNNPACK/src/configs/transpose-config.c:215:61: error: use of undeclared identifier 'xnn_x32_transposec_ukernel__8x8_rvv'
  215 |           .const_size_ukernel = (xnn_transposec_ukernel_fn) xnn_x32_transposec_ukernel__8x8_rvv,
      |                                                             ^
XNNPACK/src/configs/transpose-config.c:218:35: error: no member named 'vlenb' in 'struct xnn_hardware_config'
  218 |       } else if (hardware_config->vlenb == 16) {
      |                  ~~~~~~~~~~~~~~~  ^
XNNPACK/src/configs/transpose-config.c:220:61: error: use of undeclared identifier 'xnn_x32_transposec_ukernel__4x4_rvv'
  220 |           .const_size_ukernel = (xnn_transposec_ukernel_fn) xnn_x32_transposec_ukernel__4x4_rvv,
```
There's possible alternative fixes (e.g. fixing the ifdefs in transpose-config.c so that code isn't compiled, or defaulting the `XNNPACK_ENABLE_RISCV_VECTOR` to OFF except for RISCV builds, etc. Switching it off here in the Emscripten section was just the lowest risk of affecting anything else option I could see.

PR #7940, opened 2025-03-04
Implement a tile dispatch in FC layer

kylo5aby:
This PR implements the distribution of tiles to non-default dimensions for certain matrix sizes in the fully connected layer.

For some matrix sizes in the FC layer, the default 5x16 tile does not yield the best performance. Compared to the optimal tile , the performance of those ops may have a gap of around 20%. I conducted tests on Intel Lunar Lake and found that when **N is not a multiple of 16 but is a multiple of 8**, a 10x8 tile delivers the best performance, with an approximate 20% improvement. Below are the time consuming results on some matrixes:

| m    | k(no scaling) | n    | 5x16 tile | 10x8 tile | ratio |
| ---- | ------------- | ---- | --------- | --------- | ----- |
| 3136 | 96            | 24   | 0.134     | 0.105     | 78%   |
| 3136 | 144           | 24   | 0.203     | 0.153     | 75%   |
| 784  | 144           | 40   | 0.076     | 0.065     | 86%   |
| 784  | 240           | 40   | 0.132     | 0.106     | 80%   |
| 3136 | 48            | 104  | 0.062     | 0.058     | 94%   |
|      |               |      |           |           |       |

I tested on Chromium (commit ID: 29ad6d8) on Intel Lunar Lake by running `tflite_benchmark` and found that using a 10x8 tile for the aforementioned matrix sizes improves performance by approximately 2% on MobileNetV2 and EfficientNet. The exact improvement depends on how much time the model spends in the FC layer.

```shell
// command
 .\tflite_benchmark.exe  --warmup_min_secs=0 --num_runs=500 --num_threads=1 --enable_op_profiling=false --use_xnnpack=true --graph="path\to\efficientnet.tflite"
```

Additionally, when N does not satisfy the condition of "not a multiple of 16 but a multiple of 8," the 5x16 tile generally performs better overall, though the difference from the 10x8 tile is small. This behavior may be related to SIMD lane utilization, where the default tile introduces additional redundant computations. As N increases, the proportion of fully utilized computations becomes higher, gradually reducing the performance gap between the default tile and the 10x8 tile.
image


src/configs/gemm-config.c
```diff
@@ -1584,6 +1586,18 @@ static void init_f32_gemm_nr2_config(void) {
   assert(f32_gemm_nr2_config.mr <= XNN_MAX_MR);
 }
 
+static void init_f32_gemm_tile_10x8_config(void) {
```
dsharlet, line 1589: I don't think we should be adding a new config for this. We already have an array of kernels indexed by mr, we should add support for more than 2 of those to be non-null.

kylo5aby:
> Thanks for the interesting PR.
> 
> In general, we're aware that our system has a limitation in that we can't choose the optimal MR for particular cases. This PR seems like a way to address that in a particular case.
> 
> Currently, a config supports 2 kernels: MR = 1, and MR = MR. We should generalize that, so we can support multiple different MR values.
> 
> I think if we had that ability, that would enable what your PR is trying to do.

Thanks the feedback. 

I collected performance data on using different tile sizes for some models and found that when specifying NR, the optimal MR combinations are limited. For example, when NR = 8, the best MR values are 8, 9, and 10; when NR = 16, the best MR values are 4 and 5; and when NR = 24, MR = 3 often performs better. Overall, the 5x16, 10x8, and 8x8 tile sizes are good choices. I believe this is due to constraints such as register utilization and spilling overhead, or other limiting factors. In general, the choice of MR is limited on my test, and different NR values also impact performance. the fact that the 10x8 tile outperforms the 5x16 tile  is an example of this effect.

image

src/configs/gemm-config.c
kylo5aby: Here I just followed nr2_config. when the rule met, changed the default 5x16 tile to 10x8 tile, which has a better performance on those rules.

dsharlet:
To be clear, I definitely think that what your change is trying to do is good, and I think your data shows it is valuable.

However, I really want to avoid adding a new config. That adds a lot of issues to deal with, and I don't think it is necessary. I think what we need to do here is:

- Your example is basically exactly the use case of gemm_nr2. It looks like the existing gemm_config uses nr=16, and gemm_nr2_config uses nr = 8, so you could just add an MR=10 case to gemm_nr2_config.
- We just need to add support for more than one MR (other than mr=1) in all cases. We already do this for igemm: https://github.com/google/XNNPACK/blob/1a11ab7b3ef1cb3a477072ba165b19f78a48aa0f/src/xnnpack/operator-utils.h#L62 in some cases, but the equivalent for gemm is currently unused https://github.com/google/XNNPACK/blob/1a11ab7b3ef1cb3a477072ba165b19f78a48aa0f/src/xnnpack/operator-utils.h#L56
- We don't support gemm_nr2_config in all cases, we need to fix that.

I think if we fixed all of the above, it would enable the performance improvement you've identified, without adding a new kind of gemm config, and would be more generally useful (for more than just fully connected).

kylo5aby:
> To be clear, I definitely think that what your change is trying to do is good, and I think your data shows it is valuable.
> 
> However, I really want to avoid adding a new config. That adds a lot of issues to deal with, and I don't think it is necessary. I think what we need to do here is:
> 
> * Your example is basically exactly the use case of gemm_nr2. It looks like the existing gemm_config uses nr=16, and gemm_nr2_config uses nr = 8, so you could just add an MR=10 case to gemm_nr2_config.
> * We just need to add support for more than one MR (other than mr=1) in all cases. We already do this for igemm: https://github.com/google/XNNPACK/blob/1a11ab7b3ef1cb3a477072ba165b19f78a48aa0f/src/xnnpack/operator-utils.h#L62
>    in some cases, but the equivalent for gemm is currently unused https://github.com/google/XNNPACK/blob/1a11ab7b3ef1cb3a477072ba165b19f78a48aa0f/src/xnnpack/operator-utils.h#L56
> * We don't support gemm_nr2_config in all cases, we need to fix that.
> 
> I think if we fixed all of the above, it would enable the performance improvement you've identified, without adding a new kind of gemm config, and would be more generally useful (for more than just fully connected).

thanks. I think a general gemm or igemm configuration would be useful and hope it can be implemented ASAP, and appreciate that my PR would be helpful.
I have removed the new 10x8 gemm config and added  MR==10 to nr2_config. After a general gemm config is supported , maybe this change can be considered in  `xnn_get_heuristic_mr_gemm/igemm`

src/operators/fully-connected-nc.c
```diff
@@ -1673,6 +1673,20 @@ enum xnn_status xnn_create_fully_connected_nc_f32(
     }
   }
 
+// Only test on x86/x64 for now.
```
dsharlet, line 1676: Do you actually need this new logic? The check above is going to use `gemm_nr2_config` if `nr > output_channels`, so I would expect it to use `gemm_nr2_config` already?

kylo5aby: > Do you actually need this new logic? The check above is going to use `gemm_nr2_config` if `nr > output_channels`, so I would expect it to use `gemm_nr2_config` already?

in the experiments on the TFLite benchmark, when `nr <= output_channels`, the default configuration may still not be optimal, as it might select a configuration with a different nr value (e.g, nr=8). it can also be observed from the table above, where for `output_channels` of 40 and 104, the 10×8 tile configuration performs better.

so I believe the current NR routing rule (`gemm_config->nr > output_channels`) is not sufficient yet.

dsharlet: I see, I did some bad math.

I think this new logic is too much of a corner case, if there were a bug in it, our tests would have a hard time catching it. I think it would be better to just expand the existing condition. Maybe something like:
```
size_t nr = gemm_config->nr;
size_t nr2 = gemm_nr2_config->nr;
// There's probably a better way to compute this...
size_t nr_overcompute = (nr - output_channels % nr) % nr;
size_t nr2_overcompute = (nr2 - output_channels % nr2) % nr2;
if (nr2_overcompute < nr_overcompute && nr_overcompute * 100 > output_channels) {
  gemm_config = gemm_nr2_config;
}
```
The idea here is to use gemm_nr2_config if it would have less overcompute than gemm_nr_config, and the amount of overcompute is not negligible (the above considers <1% overcompute to be negligible).

kylo5aby: great! this is a better way to make gemm config routing more general.

src/operators/fully-connected-nc.c
```diff
@@ -1666,10 +1666,18 @@ enum xnn_status xnn_create_fully_connected_nc_f32(
...
-    if (gemm_nr2_config != NULL && gemm_nr2_config->minmax.gemm[gemm_nr2_config->mr-1].function[XNN_UARCH_DEFAULT] != NULL) {
-      gemm_config = gemm_nr2_config;
+
+  if (gemm_nr2_config != NULL) {
+    const size_t nr = gemm_config->nr;
+    const size_t nr2 = gemm_nr2_config->nr;
+    size_t nr_overcompute = (nr - output_channels % nr) % nr;
+    size_t nr2_overcompute = (nr2 - output_channels % nr2) % nr2;
+
+    if (nr > output_channels || (nr2_overcompute < nr_overcompute && nr_overcompute * 100 > output_channels)) {
```
dsharlet, line 1676: Add comment explaining this logic.

Did you test this and verify that it produces the same performance improvements that you were seeing with your original change?

kylo5aby: comment has been added. Yeah, I have tested and observed the performance of the ops on the tflite benchmark. it has the same performance. this logic will route to a new kernel when the default kernel has significant overcompute, for example, when N = 24 or 40  as shown in the picture above. it should be more general compared to the original change.

dsharlet (approved):

dsharlet:
You need to run `tools/update-microkernels.py` to fix the tests

dsharlet (approved):

kylo5aby:
> The fix for the current failures is to merge master and regenerate the kernels

Resolved. it can be compiled and passed `f32-igemm-test` and `f32-gemm-test` tests on my local machine. Also added tests for the new kernels

dsharlet (approved):

PR #8048, opened 2025-03-15
Fix hexagon build issue for unary test

ejparkqc:
We needed to use `int32_t` instead of `int` for hexagon build. 
If there is any better fix, please feel free to let me know!

### DIFF
--- a/src/xnnpack/simd/f32-hvx.h
+++ b/src/xnnpack/simd/f32-hvx.h
@@ -9,12 +9,13 @@
 
 #include <assert.h>
 #include <stddef.h>
+#include <string.h>  // for memcpy
 
 #include <hvx_hexagon_protos.h>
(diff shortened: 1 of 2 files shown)


src/xnnpack/simd/f32-hvx.h
```diff
@@ -146,7 +150,7 @@ static XNN_INLINE xnn_simd_f32_t xnn_loadu_f32(const float* ptr) {
 }
 
 static XNN_INLINE xnn_simd_f32_t xnn_load_f32(const float* ptr) {
-  return *((HVX_Vector*) ptr);
+  return *((HVX_UVector*) ptr);
```
dsharlet, line 153: This is supposed to be an aligned load, `xnn_loadu_f32` is above and is an unaligned load

dsharlet (approved):

Issue #8096, opened 2025-03-21
GEMM/IGEMM tests not working as expected for RVV?


ken-unger:
@dsharlet in PR#7932 in response to my noting that the unit test didn't catch my dumb rvv igemm bug, you said;

> Yes it does seem weird the unit tests would not catch this. I am pretty sure the unit tests catch far more subtle bugs than this. I suspect there's a problem specific to RVV, could you please take a look? There's a lot that is unique to the RVV tests due to the non-constant SIMD length.

I started looking into this a bit, and for sure there seems to be some problem.   I've noted that qd8-f32-qc8w-igemm-minmax-test and qu8-igemm-minmax-fp32-test (an upcoming RVV PR) do not actually execute *any* of the tests claimed to pass!  In other gemm & igemm tests I see very few RVV tests actually executed, and in some cases, notably qu8-igemm-minmax-fp32-test, even the scalar tests do not appear to be actually executed. 

So far just simplistic debugging, with a couple of printfs to see if it ever makes it here https://github.com/google/XNNPACK/blob/master/test/gemm-microkernel-tester.cc#L80 

I'll look at this further in the next few days but thought I'd log this issue to track,

ken-unger:
I'm probably not understanding something correctly, since this has been here for a while but shouldn't https://github.com/google/XNNPACK/blob/master/test/gemm-microkernel-tester.cc#L70

`
            for (size_t bl = params.loop_bl_.from; bl <= tester.k() / 2;
               bl = params.loop_bl_.next(bl)) {
`

be this instead?  Or at least this solves the issue.

`         for (size_t bl = params.loop_bl_.from; bl <= (tester.k() + 1) / 2;
               bl = params.loop_bl_.next(bl)) {
`

CNOCycle:
This issue has been reported in #7949. Some propoposed solutions were proposed and await further discussion.

fbarchard:
I'm told the / 2 in the tester is fixed, and a flurry of rvv bugs were exposed, but promptly disabled/fixed

PR #8105, opened 2025-03-23
[RVV] Add qu8-gemm/qu8-igemm kernels for rvv

ken-unger:
- Adding these QU8 kernels for completeness, although I recognize that QU8 will be deprecated at some point.
- Small change to qs8-gemm/rvv.in for qu8 support, however qs8/qd8 generated kernels remain unchanged. 
- Unit tests pass.  I used a local fix for https://github.com/google/XNNPACK/issues/8096, as described there, to properly execute the tests, however I decided not to include that fix here as that should really require re-execution of all gemm/igemm tests for all archs.
- Performance of 4x4v and 7x4v is roughly equivalent on the test platform (K1) at ~15 G/s and so I've used 1x4v and 4x4v for the production config.
- In future I'll take a look at why for RVV the generated gemm/igemm tests use the duplicate (mostly) CreateTests2 as its not clear to me why that is required. 

@dsharlet and @fbarchard  please review when you are able.  Thank you.


src/configs/gemm-config.c
```diff
@@ -5082,6 +5082,15 @@ static void init_qu8_gemm_config(void) {
...
     }
+  #elif XNN_ARCH_RISCV && XNN_ENABLE_RISCV_VECTOR
+    const struct xnn_hardware_config* hardware_config = xnn_init_hardware_config();
+    qu8_gemm_config.minmax.gemm[XNN_MR_TO_INDEX(1)] = xnn_init_hmp_gemm_ukernel((xnn_gemm_ukernel_fn) xnn_qu8_gemm_minmax_fp32_ukernel_1x4v__rvv);
+    qu8_gemm_config.minmax.gemm[XNN_MR_TO_INDEX(4)] = xnn_init_hmp_gemm_ukernel((xnn_gemm_ukernel_fn) xnn_qu8_gemm_minmax_fp32_ukernel_4x4v__rvv);
+    qu8_gemm_config.minmax.igemm[XNN_MR_TO_INDEX(1)] = xnn_init_hmp_igemm_ukernel((xnn_igemm_ukernel_fn) xnn_qu8_igemm_minmax_fp32_ukernel_1x4v__rvv);
+    qu8_gemm_config.minmax.igemm[XNN_MR_TO_INDEX(4)] = xnn_init_hmp_igemm_ukernel((xnn_igemm_ukernel_fn) xnn_qu8_igemm_minmax_fp32_ukernel_4x4v__rvv);
+    qu8_gemm_config.init.qu8 = xnn_init_qu8_conv_minmax_fp32_scalar_params;
+    qu8_gemm_config.mr = 4;
+    qu8_gemm_config.nr = 4 * hardware_config->vlenb / sizeof(int32_t);
```
fbarchard, line 5093: qu8_gemm_config.nr = hardware_config->vlenb / sizeof(uint8_t);
The 4v kernel is m1 so NR = vlenb

ken-unger: Yes I realized part way through this that I really should have changed the naming here to be 1v rather than 4v as you say.  But I also realized that I had done similar for the other Q gemm/igemm kernels ... so I probably should go back and fix those in my next pass.   Thanks for all the other notes above, I hope to get back to this in a couple of days.

PR #8212, opened 2025-04-04
Add support for Arm Branch Target Identification (BTI)

### LINKED ISSUE #8209: XNNPACK doesn't link with -Wl,-z,force-bti
I'm compiling XNNPACK as part of Chromium on Android (example GN args):
```
is_debug=false
symbol_level=1
target_cpu="arm64"
target_os="android"
is_high_end_android=true
build_tflite_with_xnnpack=true
```

I'm getting errors like this:
```
ld.lld: error: obj/third_party/xnnpack/f32-gemm_arch=armv8.2-a+fp16+dotprod_standalone/f32-gemm-1x8-minmax-asm-aarch64-neonfma-ld128-acc4.o: -z force-bti: file does not have GNU_PROPERTY_AARCH64_FEATURE_1_BTI property
ld.lld: error: obj/third_party/xnnpack/f32-gemm_arch=armv8.2-a+fp16+dotprod_standalone/f32-gemm-4x2-minmax-asm-aarch64-neonfma-ld128.o: -z force-bti: file does not have GNU_PROPERTY_AARCH64_FEATURE_1_BTI property
ld.lld: error: too many errors emitted, stopping now (use --error-limit=0 to see all errors)
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

The problem here is that Chrome requires all assembly code to support Arm's Branch Target Identification (BTI), which is a control flow integrity enforcement technology. BTI's purpose is to help ensure that indirect branch targets (i.e. those from a computed value in a register) can only reach the intended code. As a [security-critical library](https://source.chromium.org/chromium/chromium/src/+/main:third_party/xnnpack/README.chromium;l=1?q=xnnpack%2FREADME.&ss=chromium) that's (potentially) used for some features, it's important that XNNPACK adopts this feature on Android. It's only relevant for 64-bit Arm. Whilst it's an all-or-nothing flag, it should have a _very_ minimal performance impact and could be enabled by default.

To support it:

* Any assembly kernels which are called indirectly (e.g. via a function pointer) need a [`bti c`](https://developer.arm.com/documentation/100076/0100/A64-Instruction-Set-Reference/A64-General-Instructions/BTI) landing pad.
* Any branch targets which are called indirectly need a `bti j` landing pad.
* All assembly files need to be compiled with [`-mmark-bti-property`](https://reviews.llvm.org/D81930) flag. This emits a special `.note.gnu.property` section which indicates to the linker that the assembly file supports BTI. We'll need to do this for both GN and Bazel.

Once this is done and XNNPack is rolled, we can (perhaps) switch to using XNNPACK for inference on Android Chrome.

Sentimentron:
* -Wl,-z,force-bti is added to the linker flags on Android
* .cc/.c files have -mbranch-protection=bti set
* .s kernels get a `bti c` landing pad and `-mmark-bti-property` to add the note section

Verified by building all XNNPACK tests for Android via scripts/build-android-arm64.sh, running the binaries on a Pixel 9. Fixes #8209 for CMake builds.


fbarchard (approved):

Sentimentron:
OK, looks like we need one more maintainer to commit this

PR #8533, opened 2025-06-03
Explicit type arg in std::max for Hexagon build

ejparkqc:
`std::max` to `std::max<int32_t>` to avoid build error.


fbarchard (approved):

ejparkqc:
I didn't realize this PR was still sitting :D Just rebased then I see "merging is blocked", anything can be fixed from my side?

Issue #8488, opened 2025-05-23
Rename add_library(memory) to something else


kirklandsign:
If I add_subdirectory() both XNNPACK and abseil, library name `memory` can cause CMake conflict with absl 

https://github.com/google/XNNPACK/blob/master/CMakeLists.txt#L859-L865

https://github.com/abseil/abseil-cpp/blob/e4c43850ad008b362b53622cb3c88fd915d8f714/absl/memory/CMakeLists.txt#L17-L28

So I am wondering can we rename it, into something like memory-utils?

PR #8552, opened 2025-06-07
Rename libraries to include 'xnnpack-' prefix to avoid conflict

fs-eire:
try to fix https://github.com/google/XNNPACK/issues/8488


fs-eire:
try to fix #8488

gonnet:
I think you will also need to update the `CMakeLists.txt` files in the `bench/` and `test/` subdirectories as well.

fs-eire:
@gonnet please help to take a look again

dsharlet (approved):

dsharlet:
I tried various things that seem like they should work, but CMake seems to foil anything that might be considered any kind of namespace. We should just do this renaming.

PR #8562, opened 2025-06-10
Add QNX to recognized systems

eleir-contrib:
We build XNNPACK as a dependency of pytorch mobile and tensorflow lite, see https://github.com/qnx-ports/build-files

For QNX SDP 8.0 it doesn't take many changes for the build to pass.
For QNX SDP 7.1 the build will only work with older versions of XNNPACK (because of missing fp16 / bf16 cpu ISAs).


eleir-contrib:
Reopened: https://github.com/google/XNNPACK/pull/7798

@fbarchard

dsharlet (approved):

PR #8593, opened 2025-06-16
Rename CMake memory target to avoid name conflict

GregoryComer:
The CMake memory target can conflict with the target of the same name in Abseil. This has caused some issues with PyTorch-family builds (https://github.com/google/XNNPACK/issues/8488). This PR renames memory to memory-utils in CMake. As this is a private target, it should hopefully be a safe change as users should not directly depend on the target.

To validate the changes, I built XNNPACK on an M1 Mac. It configures, builds, and runs benchmarks successfully.

Edit: It looks like https://github.com/google/XNNPACK/pull/8552 is already up to try to resolve this. Either approach should solve the issue, though the other one is a little more comprehensive.


dsharlet:
I approved #8552 because it seems likely that our other libraries will collide with something in the future, not just `memory`. How do you see this PR as being more comprehensive than #8552?

GregoryComer:
> I approved #8552 because it seems likely that our other libraries will collide with something in the future, not just `memory`. How do you see this PR as being more comprehensive than #8552?

Ah, I actually meant yours is more comprehensive. If yours is merged, that solves the PyTorch issues so I'm happy to close this PR.

PR #8647, opened 2025-06-26
Fix the issue that Pad cannot be fused into conv2d

fujunwei:
The zero point should not be coverted to uint8 if the data type is int8.

### DIFF
--- a/test/subgraph/fusion.cc
+++ b/test/subgraph/fusion.cc
@@ -375,6 +375,63 @@ TEST(CONSTANT_PAD_THEN_CONVOLUTION, fusion) {
   ASSERT_EQ(unoptimized_output, optimized_output);
 }
 
+TEST(CONSTANT_PAD_THEN_CONVOLUTION, fusion_quantized_int8) {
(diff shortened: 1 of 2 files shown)


test/subgraph/fusion.cc
```diff
@@ -375,6 +375,63 @@ TEST(CONSTANT_PAD_THEN_CONVOLUTION, fusion) {
   ASSERT_EQ(unoptimized_output, optimized_output);
 }
 
+TEST(CONSTANT_PAD_THEN_CONVOLUTION, fusion_quantized_int8) {
```
gonnet, line 378: Please run `clang-format` on this file!

dsharlet:
Thanks for the fix!

gonnet (approved):

test/subgraph/fusion.cc
fujunwei: Sorry for missing the comment, i was on vacation last week and started working today.  Thanks for reviewing and merging it.

PR #8649, opened 2025-06-27
Add ohos macro in the CMakeLists.txt for cross compilation of OpenHarmony

SC404:
we are using XNNPACK lib in OpenHarmony, so we'd like to contribute the build guide, hope it can help developers.


CMakeLists.txt
```diff
@@ -94,7 +94,7 @@ ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "^i[3-7]86$")
   SET(XNNPACK_TARGET_PROCESSOR "x86")
 ELSEIF(CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64")
   SET(XNNPACK_TARGET_PROCESSOR "x86_64")
-ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "^armv[5-8]")
+ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "^(armv[5-8]|arm.*)$")
```
fbarchard, line 97: this is for aarch32 builds, which is supported up to armv8 but not armv9.  Can you clarify what 'arm' the processor is reporting?

CMakeLists.txt
```diff
@@ -940,7 +940,7 @@ IF(XNNPACK_TARGET_PROCESSOR STREQUAL "arm")
   # Related links:
   #   https://github.com/android/ndk/issues/910
   #   https://reviews.llvm.org/D58477
-  IF(ANDROID_NDK_MAJOR AND ANDROID_NDK_MAJOR LESS 21)
+  IF((ANDROID_NDK_MAJOR AND ANDROID_NDK_MAJOR LESS 21) OR OHOS)
```
fbarchard, line 943: is OHOS defined on its own?  Shouldnt this be CMAKE_SYSTEM_NAME MATCHES OHOS like above?
This section is meant for old ndk builds for armv7 where the abi needs to be specified.  Is this true for all harmony builds/cpus?

PR #8687, opened 2025-07-08
Prototype: Add support for fp16 iGEMM with SME2

gmiodice:
- Initial prototype to enable fp16 iGEMM with SME2 in conv2d


src/x16-pack-lh/x16-packlh-igemm-neonsme2.c
```diff
@@ -0,0 +1,58 @@
...
+#endif  // XNN_ENABLE_KLEIDIAI
+}
+
+size_t xnn_x16_pack_lh_offset__igemm_neonsme2(size_t m, size_t kc, size_t ks,
+                                             size_t mr_packed, size_t kr,
+                                             size_t sr) {
+#if XNN_ENABLE_KLEIDIAI
+  assert(kr == 2);
+
+  return kai_get_lhs_packed_offset_lhs_imatmul_pack_x16p2vlx2_x16p_sme(
```
fbarchard: This appears to be an SME kernel.  SME2 is not required?
If so we have emulator support for SME that could test this function with sme, but not SME2.
Is there a unittest we can run for x16-pack-lh microkernels?

src/x16-pack-lh/x16-packlh-igemm-neonsme2.c
```diff
@@ -0,0 +1,58 @@
...
+
+  return kai_get_lhs_packed_size_lhs_imatmul_pack_x16p2vlx2_x16p_sme(
+      m, ks, kc);
+#else
+  assert("Not compiled with XNN_ENABLE_KLEIDIAI" && 0);
+  return 0;
+#endif  // XNN_ENABLE_KLEIDIAI
+}
+
+size_t xnn_x16_pack_lh_offset__igemm_neonsme2(size_t m, size_t kc, size_t ks,
```
fbarchard, line 53: neon should be removed from these function/file isa names.  in this case, the function it calls is _sme so the cpuinfo should detect sme (1.0) and typically the compiler would need to support sme, but not sme2, to build microkernels.

src/reference/packing.cc
```diff
@@ -2080,6 +2081,15 @@ void transpose_weights_x8(const int8_t* in, int8_t* out, size_t height,
   }
 }
 
+void transpose_weights_x16(const uint16_t* in, uint16_t* out, size_t height,
```
fbarchard: is this 'x16' meant to be 16 bits per element?  if so it should be at the start of function/file name
x16_transpose_weights()
unless you're proposing a public api for transpose, this should be static or inlined into the calling function.

src/reference/packing.cc
```diff
@@ -2080,6 +2081,15 @@ void transpose_weights_x8(const int8_t* in, int8_t* out, size_t height,
...
+                          size_t width) {
+  for (size_t i = 0; i < height; ++i) {
+    for (size_t j = 0; j < width; ++j) {
+      out[j * height + i] = in[i * width + j];
```
fbarchard, line 2088: reverse the order - write rows, read columns

  for (size_t j = 0; j < width; ++j) {
    for (size_t i = 0; i < height; ++i) {
      out[j * height + i] = in[i * width + j];

src/subgraph.c
```diff
@@ -2453,6 +2462,7 @@ enum xnn_status xnn_subgraph_optimize_packed_lhs(xnn_subgraph_t subgraph,
 
       case xnn_node_type_convolution_2d:
       case xnn_node_type_deconvolution_2d: {
+
```
fbarchard: remove blank line

src/x16-pack-lh/x16-packlh-igemm-neonsme2.c
```diff
@@ -0,0 +1,58 @@
...
+                                             size_t mr_packed, size_t kr,
+                                             size_t sr) {
+#if XNN_ENABLE_KLEIDIAI
+  assert(kr == 2);
```
fbarchard, line 57: the kai function appears to be for x16 (NR=16)? 
I see kai struggles with the over use of 'x' just like XNNPack :-)
one of the x16 appears to be bits per element. In XNNPack that goes first in the name.
The x in pack functions is from MRxNR but MR is not applicable, so x16 is NR=16
We used to also have x2 means unrolled by 2, but renamed that to u2.
Anyway, the kai name implies a specific tiling.
In packw microkernels it is the same and the function/file are specific to a set of parameters.  NR, KR, SR
The function is specialized and expected to fail if called with anything else.
If kai is the same, the assert for KR=2 (ah x2.  We would call that c2).. the assert should check all applicable parameters, but especially NR.

src/configs/pack-lh-config.c
```diff
@@ -232,4 +234,29 @@ const struct xnn_pack_lh_config* xnn_init_x8_igemm_pack_lh_config() {
...
+    assert(hardware_config != NULL);
+    if ((hardware_config->arch_flags & xnn_arch_arm_sme2)) {
+      x16_igemm_pack_lh_config.pack_lh_for_igemm_fn = (xnn_pack_lh_igemm_ukernel_fn) xnn_x16_pack_lh_ukernel__igemm_neonsme2;
+      x16_igemm_pack_lh_config.size_for_igemm_fn = (xnn_pack_lh_igemm_size_fn) xnn_x16_pack_lh_size__igemm_neonsme2;
+      x16_igemm_pack_lh_config.offset_for_igemm_fn = (xnn_pack_lh_igemm_offset_fn) xnn_x16_pack_lh_offset__igemm_neonsme2;
+    }
+  #endif  // XNN_ENABLE_ARM_SME2
+  #endif  // XNN_ARCH_ARM64 && XNN_ENABLE_KLEIDIAI
+  x16_igemm_pack_lh_config.log2_input_element_size = 0;
+  x16_igemm_pack_lh_config.log2_packed_element_size = 0;
```
gonnet: These values should be `1` since `fp16` is `1 << 1` bytes.

src/operators/convolution-nhwc.c
```diff
@@ -1468,6 +1468,142 @@ enum xnn_status xnn_create_convolution2d_nhwc_pqs8_qs8_qs8(
   return status;
 }
 
+enum xnn_status xnn_create_convolution2d_nhwc_pf16(
```
gonnet: This function and `xnn_create_convolution2d_nhwc_f16` (below it) differ only in the `gemm_config` and `operator_type`.

Could you create a single `create_convolution2d_nhwc_f16` that takes the `gemm_config` and `operator_type` as an argument, and have this function and `xnn_create_convolution2d_nhwc_f16` call it?

src/pf16-f16-f16-igemm/pf16-f16-f16-igemm-32x32c2-minmax-neonsme2.c
```diff
@@ -0,0 +1,54 @@
...
+      cm_stride, sizeof(xnn_float16),
+      xnn_float16_to_float(params->scalar.min),
+      xnn_float16_to_float(params->scalar.max));
+#else
+  assert(
+      "Calling wrapped KleidiAI function, but XNNPACK was compiled without "
+      "`XNN_ENABLE_KLEIDIAI`." &&
+      0);
+#endif  // XNN_ENABLE_KLEIDIAI
+}
```
gonnet: Please add a final newline.

src/pf16-f16-f16-igemm/pf16-f16-f16-igemm-32x32c2-minmax-neonsme2.c
```diff
@@ -0,0 +1,54 @@
+#include <stddef.h>
```
gonnet, line 6: Copyright header?

src/reference/packing.cc
gonnet: I think this function is meant to be internal-only.

src/reference/packing.cc
```diff
@@ -2465,6 +2475,50 @@ void xnn_pack_kai_qb4_weights_and_biases(
...
+  uint16_t* tmp_data =
+      (uint16_t*)xnn_allocate_memory(nc * ks * kc * sizeof(uint16_t));
+  const size_t rhs_row_stride = nc * sizeof(uint16_t);
+  const size_t packed_rhs_size =
+      kai_get_rhs_packed_size_rhs_imatmul_pack_kxn_x16p2vlx2b_x16_x16_sme(
+          nc, ks, kc);
+
+  for (size_t g_idx = 0; g_idx < g; ++g_idx) {
+    transpose_weights_x16(k, tmp_data, nc, ks * kc);
+    kai_run_rhs_imatmul_pack_kxn_x16p2vlx2b_x16_x16_sme(
```
gonnet, line 2517: Is there any chance we can rewrite this function, or provide an analogue, that does not require

1. The empty `bias` pointer, and
2. The manual transpose?

src/subgraph.c
```diff
@@ -2368,6 +2368,15 @@ enum xnn_status xnn_subgraph_optimize_packed_lhs(xnn_subgraph_t subgraph,
             if (assumed_datatype == xnn_datatype_qdint8) {
               // If the input is already `qdint8`, don't do anything different.
               continue;
+            } else if (assumed_datatype == xnn_datatype_pfp16) {
```
gonnet: I think this is in the wrong place. As per lines `2295`--`2296` above, this is only for `xnn_node_type_batch_matrix_multiply` and `xnn_node_type_fully_connected`.

The `if (assumed_datatype == xnn_datatype_qdint8)` here, and the `else if (assumed_datatype == xnn_datatype_qpint8)` just below, are to handle `qd8` LHS "packing", which is slightly different as a conversion node already exists and thus a new `pack-lh` node does not need to be added.

src/subgraph.c
```diff
@@ -2477,6 +2487,16 @@ enum xnn_status xnn_subgraph_optimize_packed_lhs(xnn_subgraph_t subgraph,
           node->packed_input_datatype = xnn_datatype_pqint8;
           node->flags |= XNN_FLAG_INLINE_LHS_PACKING;
         }
+        if (input_datatype == xnn_datatype_fp16 &&
```
gonnet, line 2479: This should be something like
```c
          } else if (input_datatype == xnn_datatype_qint8 &&
            kernel_datatype == xnn_datatype_fp16 &&
            output_datatype == xnn_datatype_fp16 &&
            xnn_init_pf16_gemm_config() != NULL
            !(optimization_flags & XNN_FLAG_NO_INLINED_LHS_PACKING)) {
```
i.e. check the types of `input`/`kernel`/`output`, and make sure that LHS packing is not disabled.

src/subgraph.c
gonnet: Please run `clang-format` using this project's config on all modified files.

src/subgraph/convolution-2d.c
```diff
@@ -309,6 +309,31 @@ static enum xnn_status create_convolution_operator(
         switch (filter_datatype) {
           case xnn_datatype_fp16:
           case xnn_datatype_fp32:
+            if(input_datatype == xnn_datatype_pfp16) {
+
```
gonnet: Please format correctly.

src/subgraph/convolution-2d.c
```diff
@@ -309,6 +309,31 @@ static enum xnn_status create_convolution_operator(
         switch (filter_datatype) {
           case xnn_datatype_fp16:
           case xnn_datatype_fp32:
+            if(input_datatype == xnn_datatype_pfp16) {
```
gonnet: Can this handle `fp32` filters correctly?

src/x16-pack-lh/x16-pack-lh-igemm.inc
```diff
@@ -0,0 +1,7 @@
+// arch_flags, igemm_ukernel, igemm_size_fn, igemm_packed_offset_fn
```
gonnet, line 7: Missing copyright header.

src/x16-pack-lh/x16-packlh-igemm-neonsme2.c
```diff
@@ -0,0 +1,58 @@
+#include <arm_neon.h>
```
gonnet, line 7: Missing copyright header.

src/x16-pack-lh/x16-packlh-igemm-neonsme2.c
gonnet: This is OK as long as it follows the naming convention for the other `pack-lh` kernels, which it does.

src/x16-pack-lh/x16-packlh-igemm-neonsme2.c
gonnet: KleidiAI has a very extensive documentation of their type naming conventions on their GitLab/GitHub pages.

src/x16-pack-lh/x16-packlh-igemm-neonsme2.c
gonnet: This is the KleidiAI naming convention, so not something that needs to be addressed here.

src/xnnpack/gemm.h
```diff
@@ -340,6 +340,19 @@ DECLARE_PQS8_QC8W_GEMM_MINMAX_UKERNEL_FUNCTION(
 DECLARE_PQS8_QC8W_PACKED_IGEMM_MINMAX_UKERNEL_FUNCTION(
     xnn_pqs8_qc8w_igemm_minmax_fp32_ukernel_32x32c4__neonsme2)
 
+
+#define DECLARE_PF16_F16_PACKED_IGEMM_MINMAX_UKERNEL_FUNCTION(fn_name)       \
```
gonnet: Please fix formatting.

src/xnnpack/pack.h
```diff
@@ -491,6 +491,21 @@ XNN_INTERNAL size_t xnn_packed_stride_kai_qb4_weights_and_biases(
     size_t k_stride,                            //
     size_t extra_bytes);
 
+XNN_INTERNAL void xnn_pack_kai_f16_conv_goki_w_sme2(
```
gonnet: Please fix formatting.

test/gemm-microkernel-tester.cc
```diff
@@ -2802,6 +2802,185 @@ void GemmMicrokernelTester::Test_PQS8(
...
+  xnnpack::Buffer<xnn_float16, XNN_ALLOCATION_ALIGNMENT> b_packed_f16(ks() * packed_n() * packed_k() +
+      packed_n() * (sizeof(int32_t) + sizeof(float)) / sizeof(int8_t), /*extra_bytes=*/{0},
+                                         "weights_packed_f16");
+  xnnpack::Buffer<xnn_float16> bias(n(), 0.0f);
+  xnnpack::Buffer<xnn_float16> c((m() - 1) * cm_stride() + n(),
+                                 /*extra_bytes=*/{0}, "c");
+  xnnpack::Buffer<xnn_float16> c_ref(m() * n(), 0, /*extra_bytes=*/{0},
+                                     "c_ref");
+
+#if 0
```
gonnet: Does this test pass?

test/gemm-microkernel-tester.h
```diff
@@ -354,6 +354,12 @@ class GemmMicrokernelTester {
                  xnn_pack_weights_and_biases_fn pack,
                  xnn_packed_stride_weights_and_biases_fn packed_stride);
 
+void Test_PF16(xnn_packed_f16_lhs_igemm_ukernel_fn packed_igemm,
```
gonnet: For this function to be called, you need to add your new kernel to the corresponding `.yaml` files and likely also add a few lines to the test generator script (see e.g. `XNNPACK/test/qp8-f32-qc8w-gemm-minmax.cc`).

src/subgraph.c
JonathanC-ARM: clang format was ran against all the files in the latest push

src/configs/pack-lh-config.c
JonathanC-ARM: Updated to the correct value

src/operators/convolution-nhwc.c
JonathanC-ARM: Single common creation function create_convolution2d_nhwc_f16 which takes the gemm config and op type as you have suggested

src/pf16-f16-f16-igemm/pf16-f16-f16-igemm-32x32c2-minmax-neonsme2.c
JonathanC-ARM: Added headers where required

src/pf16-f16-f16-igemm/pf16-f16-f16-igemm-32x32c2-minmax-neonsme2.c
JonathanC-ARM: New line added in latest push

test/gemm-microkernel-tester.h
JonathanC-ARM: Added the yml files and updated tests, they should now run and pass

src/subgraph/convolution-2d.c
JonathanC-ARM: As far as I am aware yes, XNN_FLAG_FP32_STATIC_WEIGHTS is set when the filter is fp32, and the PF16 creation path switches its packing kernels to convert fp32 weights to fp16 at pack time.

src/subgraph.c
JonathanC-ARM: Updated to have the correct check

src/subgraph.c
JonathanC-ARM: Removed the check as its not needed

src/reference/packing.cc
JonathanC-ARM: Yeah seemed that was the intention, so have made it static

src/reference/packing.cc
JonathanC-ARM: Order swapped as suggested

src/reference/packing.cc
JonathanC-ARM: The answer to both really would be to have a new function variant which would require the kleidi team to implement new function versions and provide them in a release. The majority of their function signatures are the similar in terms of the bias. So I suspect they would hesitate to remove the bias. 

But as for the transpose this variant of the pack kernel is kxn if we had a nxk version we could remove the manual transpose, as for right now though we will need it.

dsharlet:
@gmiodice It looks like this needs conflicts resolved

@gonnet @fbarchard can you please follow up? Your comments have been responded to.

test/BUILD.bazel
```diff
@@ -380,6 +380,17 @@ xnnpack_unit_test(
     ],
 )
 
+xnnpack_unit_test(
+    name = "pf16_f16_igemm_minmax_test",
+    srcs = [
+        "pf16-f16-igemm-minmax.cc",
```
gonnet, line 386: This also needs to be added to the corresponding `CMakeLists.txt` file.

src/reference/packing.cc
```diff
@@ -2465,6 +2475,73 @@ void xnn_pack_kai_qb4_weights_and_biases(
...
+
+  xnn_release_memory(tmp_data);
+
+  if (tmp_bias != NULL) {
+    xnn_release_memory(tmp_bias);
+  }
+}
+
+size_t xnn_packed_size_kai_f16_conv_goki_w(size_t nc, size_t ks, size_t kc) {
+#if XNN_ENABLE_KLEIDIAI
```
gonnet, line 2534: I think you are already inside an `#if XNN_ENABLE_KLEIDIAI`, so the `#else` below is not needed.

src/reference/packing.cc
gonnet: > The answer to both really would be to have a new function variant which would require the kleidi team to implement new function versions and provide them in a release. The majority of their function signatures are the similar in terms of the bias. So I suspect they would hesitate to remove the bias.

Would it be sufficient for the function to handle `b == NULL` as if it were full of zeros? That would not require a change in the API?

> But as for the transpose this variant of the pack kernel is kxn if we had a nxk version we could remove the manual transpose, as for right now though we will need it.

OK, could you add a `TODO` to remind us? Thanks!

src/reference/packing.cc
JonathanC-ARM: True, removed the additional #if

PR #8695, opened 2025-07-11
Add missing wasm revectorize cmake option

yolanda15:
We've added "XNN_ENABLE_WASM_REVECTORIZE" macro in bazel for wasm simd build, but not for cmake.


CMakeLists.txt
```diff
@@ -299,6 +299,10 @@ IF(XNNPACK_TARGET_PROCESSOR STREQUAL "arm64" AND XNNPACK_ENABLE_ARM_I8MM AND NOT
 ELSE()
   SET(XNNPACK_ENABLE_KLEIDIAI OFF)
 ENDIF()
+OPTION(XNNPACK_ENABLE_WASM_REVECTORIZE "Build XNNPACK with WASM SIMD AVX-256 revectorization micro-kernels" ON)
+IF(NOT CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
+  SET(XNNPACK_ENABLE_WASM_REVECTORIZE OFF)
```
dsharlet, line 304: It looks like the bazel build doesn't bother with this, do we need this? It looks like if the build system isn't Emscripten, this option doesn't matter anyways.

yolanda15: We are working on restore the XNNPACK EP for wasm in ONNX runtime, which will use the cmake to build wasm. See issue https://github.com/microsoft/onnxruntime/issues/24766. It provides a patch like this  https://github.com/microsoft/onnxruntime/blob/main/cmake/patches/xnnpack/AddEmscriptenAndIosSupport.patch.

dsharlet: To clarify, I'm fine with adding the option, I just meant the logic to conditionally disable it when not targeting Emscripten seems unnecessary.

yolanda15: Sorry for the misunderstanding. Yeah, that also makes sense if it's acceptable :)

PR #8768, opened 2025-08-07
Support WASM Kernels in CMake With Emscripten Tool Chain

mcr229:
Support building with Emscripten Tool chain in XNNPACK's CMake. Also support leveraging some of the wasm kernels.


mcr229:
cc. @dsharlet

PR #8812, opened 2025-08-24
Fix incorrect declarations

cyyever:


src/xnnpack/internal.h
```diff
@@ -103,7 +103,7 @@ enum xnn_status xnn_setup_batch_matrix_multiply_nc_qp8_f32_qc8w(
     xnn_operator_t batch_matrix_multiply_op,  //
     void* workspace,                          //
     const int8_t* input_a,                    //
-    const int8_t* input_b,                    //
+    const float* input_b,                     //
```
fbarchard: this doesnt look right.  the input_b is qc8w which is int8_t

cyyever: But the function definition in `src/operators/batch-matrix-multiply-nc.c` uses `const float* input_b`. 
PS: it looks like I should change the definition.

src/xnnpack/internal.h
```diff
@@ -476,7 +476,7 @@ enum xnn_status xnn_reshape_batch_matrix_multiply_nc_qdu8_f32_qc8w(
 
 enum xnn_status xnn_setup_batch_matrix_multiply_nc_qdu8_f32_qc8w(
     xnn_operator_t batch_matrix_multiply_op, void* workspace,
-    const int8_t* input_a, const int8_t* input_b,
+    const int8_t* input_a, const float* input_b,
```
fbarchard, line 479: input_b is byte, not float.
qdu8_f32_qc8w
qdu8 is input_a is uint8.  but apparently declared as int8.
f32 is output is float
qc8w is weights are int8

fbarchard (changes requested):

fbarchard:
The prototype mismatch bug is landed.  If there is another issue with lto, report an issue, preferably with details on how to reproduce it.

PR #8915, opened 2025-09-25
Allow disabling of SME in bazel builds

SuhailMunshi:
- Fixes typo in bazel build defs allowing disabling of SME/SME2


dsharlet (approved):
Thanks for the fix!

fbarchard (changes requested):
also occurs in 
build_defs.bzl.opensource:        ["XNN_ENABLE_SRM_SME=0"],

PR #8930, opened 2025-10-01
Updating GEMMBenchmark constructor to use ConstantOrFunction

Colm-in-Arm:
When building XNNPACK with XNNPACK_ENABLE_ARM_SME2=ON

mkdir build && cd build
cmake .. -DXNNPACK_ENABLE_ARM_SME2=ON
cmake --build . --config Release

The build fails with a method signature mismatch in bench/gemm-benchmark.cc. This patch corrects the mismatch.


dsharlet (approved):
Thank you for the fix!

PR #9208, opened 2025-11-26
SME1 support for fp16 GEMM and IGEMM

qti-vaiskv:


qukhan:
This is a flaky test. Re-running the tests works.

Issue #8087, opened 2025-03-20
vrnd kernels for rvv are incorrect


dsharlet:
The vrnd kernels for rvv work by converting a float to an int and back. This doesn't work in a few cases:
- For floats that are outside the range of an int32
- It loses infinity/NaN

I tried to partially fix the latter in issue in #7971.

@KaustubhIMG @ken-unger

ken-unger:
Thanks for the note.  

re: // TODO: We should also preserve NaN

FWIW this is something that Eigen also needs to do, and there is an ongoing PR where min/max with NaN propagation could perhaps be used for reference.  
https://gitlab.com/libeigen/eigen/-/blob/8ac76af2902bedd9a075c32a67ec9f55d2e6b325/Eigen/src/Core/arch/RVV10/PacketMath.h#L1223

fbarchard:
I see this same behavior (cast to int) in sse2, 0armv7, wasmsimd, and hexagon.
Also hexagon qfloats do not have a formal NaN or inf.
You can find implementations for arm and intel in wasmsimd.
The basic idea is compare if the exp is large enough that the value is already an integer.  (exp >= 23?)
This includes exp = 255 for NaN and inf.
Meanwhile proceed with casting to int (with rounding modes for ties).  The basic one is away from zero for ties.
2.5 rounds to 3.0
-2.5 rounds to -3.0
3.5 rounds to 4.0
-3.5 rounds to -4.0
And convert back to float
Then use the comparison of exponent to select the original float or the float that was convert to/from int. 
Most cpus have a way to select between 2 vectors using a mask.
On ARM Neon bitselect
On SSE4 vblend_ps

See Also
|https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/main/doc/rvv-intrinsic-spec.adoc#explicit-fp-rounding-mode-intrinsics

PR #9379, opened 2026-01-10
[RVV] f32-vrnd fix for issue #8087

ken-unger:
Fix to address #8087
- Avoid NaN and int conversion for large numbers (no remainder) for f32-vrnd RVV kernels. 
- Remove test workaround for this issue.
- Tested with f32-vrnd tests and subgraph/unary-test which now all pass


fbarchard (approved):
Looks good thanks.  But the test work around looks like something needed for other cpus that are not IEEE complient.

dsharlet (approved):

PR #9516, opened 2026-02-09
[RVV] Add F16 vunary and vbinary rvv kernels

ken-unger:
- Add vunary and vbinary f16 rvv kernels  
- Add config for existing f32-f16-vcvt and f16-f32-vcvt rvv kernels

### DIFF
--- a/src/configs/binary-elementwise-config.c
+++ b/src/configs/binary-elementwise-config.c
@@ -119,7 +119,7 @@ static void init_f16_vadd_config(void) {
     f16_vadd_config.op_ukernel = XNN_INIT_BINARY_UKERNEL(xnn_f16_vadd_ukernel__rvvfp16arith_u8v);
     f16_vadd_config.opc_ukernel = XNN_INIT_BINARY_UKERNEL(xnn_f16_vaddc_ukernel__rvvfp16arith_u8v);
     f16_vadd_config.ropc_ukernel = XNN_INIT_BINARY_UKERNEL(xnn_f16_vaddc_ukernel__rvvfp16arith_u8v);
-    f16_vadd_config.element_tile = 8 * hardware_config->vlenb / sizeof(_Float16);
+    f16_vadd_config.element_tile = 8 * hardware_config->vlenb / sizeof(xnn_float16);
   #endif
 }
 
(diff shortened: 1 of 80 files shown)


ken-unger:
I followed the existing naming convention using the "fp16arith" suffix, which follows the arm usage, although alternatively for rvv this could be a "zvfh" suffix which would be the pattern for adding other or future extension support (eg zvbdot).  

Support for cpuinfo_has_riscv_zvfh() not yet provided by cpuinfo, and so this is currently enabled if vector and f16 compiled.

ken-unger:
@dsharlet I have been running on a BPI-F3, rather than using QEMU, due to speed and benchmarking accuracy.  I strongly suspect that this failure (on qemu) is for any rvv fp16 instruction, and that the fix is likely the addition of "zvfh=true" to the qemu cpu options given at https://github.com/google/XNNPACK/blob/master/cmake/riscv64.toolchain#L44

ken-unger:
@dsharlet just fyi, I was able to run this on qemu-riscv64 version 8.2.2  by adding cpu options "zfh=true, x-zvfh=true".  Note the "x-" on zvfh which may not be required on a newer version.  Run time detection of zvfh requires some updates to cpuinfo (hwprobe), however, cpuinfo is disabled for qemu builds if I understand correctly.

dsharlet:
I think that we're going to need hardware-config.c support to detect and disable this if it is missing. We are using someone else's emulator configuration and I'd rather not change it. Besides, we need this anyways if RVV is going to be a real target we can support in production.

Is this something you can implement?

ken-unger:
> Is this something you can implement?

Yes I plan to update this in the near future.  I'm preparing a pytorch/cpuinfo PR for real target support (with a subsequent xnnpack hardware-config.c PR), and will determine what to do to keep qemu happy.

ken-unger:
> This kernel shouldn't be using any unsupported instructions, right?

That is odd.  I thought for a moment that auto-vectorization might be emitting vector-half instructions for this scalar kernel but I see that those flags would only be on rvv kernels (and verified in the build log).  Also objdump of this kernel doesn't show anything suspicious.  Just for clarity on that, what compiler version is your CT using?

dsharlet:
We're using something pretty close to clang trunk, maybe delayed by a few weeks.

PR #9531, opened 2026-02-10
Optimize ARM SME2 Dot-Product Kernel

kasper0406:
# Optimize ARM SME2 Dot-Product Kernels with 4-way ZA Tiling and x4 Loads

I had a go at getting Gemini to code the proposed improvements from this paper: [Demystifying ARM SME to Optimize General Matrix Multiplications](https://arxiv.org/pdf/2512.21473)
The code has achieved the following results on my Apple M4 Pro with ARM64 SME2 hardware, tested with a python script wrapping `bazel-bin/ynnpack/kernels/dot/bench` running 10 iterations. Mean and stddev are reported along with the computed speedup.

```
(py: venv_bench) ➜  XNNPACK git:(kn/sme2-dot) ✗ uv run python bench_sme2_all_types.py
[-] Building benchmark binary... Done.

Type   | Shape           | Orig (ms)       | Opt (ms)        | Speedup
-------------------------------------------------------------------------------------
fp32   | 240x240x240     | 0.024 ±0.00     | 0.017 ±0.00     | 1.400x ±0.086
fp32   | 4096x4096x4096  | 360.991 ±3.89   | 332.242 ±1.16   | 1.087x ±0.012
fp32   | 1x4096x4096     | 1.442 ±0.02     | 1.285 ±0.00     | 1.123x ±0.013
fp32   | 128x1024x1024   | 0.183 ±0.01     | 0.165 ±0.01     | 1.110x ±0.062
fp32   | 16x16x16384     | 0.018 ±0.00     | 0.019 ±0.00     | 0.979x ±0.047
bf16   | 240x240x240     | 0.023 ±0.00     | 0.017 ±0.00     | 1.375x ±0.048
bf16   | 4096x4096x4096  | 156.802 ±1.01   | 149.963 ±0.44   | 1.046x ±0.007
bf16   | 1x4096x4096     | 0.591 ±0.00     | 0.568 ±0.00     | 1.041x ±0.007
bf16   | 128x1024x1024   | 0.154 ±0.00     | 0.151 ±0.01     | 1.021x ±0.044
bf16   | 16x16x16384     | 0.018 ±0.00     | 0.018 ±0.00     | 1.000x ±0.001
fp16   | 240x240x240     | 0.023 ±0.00     | 0.017 ±0.00     | 1.378x ±0.042
fp16   | 4096x4096x4096  | 154.456 ±0.71   | 149.419 ±0.45   | 1.034x ±0.006
fp16   | 1x4096x4096     | 0.599 ±0.00     | 0.575 ±0.00     | 1.042x ±0.006
fp16   | 128x1024x1024   | 0.153 ±0.01     | 0.147 ±0.00     | 1.037x ±0.048
fp16   | 16x16x16384     | 0.018 ±0.00     | 0.018 ±0.00     | 1.000x ±0.001
int8   | 240x240x240     | 0.009 ±0.00     | 0.008 ±0.00     | 1.139x ±0.075
int8   | 4096x4096x4096  | 58.342 ±0.25    | 57.809 ±0.68    | 1.009x ±0.013
int8   | 1x4096x4096     | 0.223 ±0.00     | 0.221 ±0.00     | 1.012x ±0.003
int8   | 128x1024x1024   | 0.073 ±0.00     | 0.072 ±0.00     | 1.007x ±0.004
int8   | 16x16x16384     | 0.005 ±0.00     | 0.005 ±0.00     | 0.992x ±0.062
```

Correctness has been verified using `//ynnpack/kernels/dot:test --gtest_filter=*sme2_opt*`

I haven't been through in reviewing the generated code yet. Before spending time on it, I would be interested to know if this is something you are even interested in pursuing?


ynnpack/kernels/dot/arm64_sme2_opt.cc
```diff
@@ -0,0 +1,352 @@
...
+    size_t A_stride_k3, size_t A_stride_k2, const void* A, size_t B_stride_k3,
+    size_t B_stride_k2, size_t B_stride_k1, const void* B, size_t C_in_stride_m,
+    const void* C_in, size_t C_out_stride_m, void* C_out) {
+  assert(M > 0);
+  assert(N > 0);
+  assert(K3 > 0);
+  assert(K2 > 0);
+  assert(K1 > 0);
+
+  A = __builtin_assume_aligned(A, 64);
```
dsharlet, line 37: I am not sure we can guarantee these assumptions:
- If we're processing fp32, tile_k = 1, so we could process unpacked B if we don't think it is profitable to pack B, which would mean we are taking arbitrary data for B that may not be aligned.
- Similarly, if A is already transposed, we might not pack A either.
- I am not sure we can ever guarantee C is aligned. But hopefully this doesn't matter much, since we aren't loading/storing it often?

How much difference does this make to performance? If the impact is small, I think we should just remove it to be safe. If there's a lot of benefit to doing this, we might need to instantiate the kernel twice and add some flags to allow the dot kernel selection to dispatch to the aligned one when possible.

ynnpack/kernels/dot/arm64_sme_internal.h
```diff
@@ -20,6 +20,9 @@ namespace ynn {
 // Helpers to call the right SME intrinsics from overloads based on the type.
 inline size_t svcnt(float) { return svcntsw(); }
 inline size_t svcnt(int32_t) { return svcntsw(); }
+inline size_t svcnt(bfloat16_t) { return svcntsh(); }
```
dsharlet: Do you use these anywhere? I didn't see them.

ynnpack/kernels/dot/arm64_sme_internal.h
```diff
@@ -79,6 +82,26 @@ YNN_ALWAYS_INLINE svbool_t svptrue(int8_t) YNN_SME_HELPER {
   return svptrue_b8();
 }
 
+YNN_ALWAYS_INLINE bool svany(svbool_t p) YNN_SME_HELPER {
+  return svptest_any(svptrue_b8(), p);
+}
+
+YNN_ALWAYS_INLINE void svprefetch(const float* ptr) YNN_SME_HELPER {
```
dsharlet: These also look unused.

ynnpack/kernels/dot/arm64_sme_internal.h
```diff
@@ -95,6 +118,63 @@ YNN_ALWAYS_INLINE svcount_t svctrue(int8_t) YNN_SME_HELPER {
   return svptrue_c8();
 }
 
+// Helpers for ZA loads/stores
+template <uint32_t tile>
+YNN_ALWAYS_INLINE void svld1_hor_za(uint32_t slice, svbool_t p, const float* ptr) YNN_SME_HELPER {
```
dsharlet: Unused?

ynnpack/kernels/dot/arm64_sme_internal.h
```diff
@@ -95,6 +118,63 @@ YNN_ALWAYS_INLINE svcount_t svctrue(int8_t) YNN_SME_HELPER {
...
+YNN_ALWAYS_INLINE void svst1_ver_za(uint32_t slice, svbool_t p, float16_t* ptr) YNN_SME_HELPER {
+  svst1_ver_za16(tile, slice, p, ptr);
+}
+template <uint32_t tile>
+YNN_ALWAYS_INLINE void svst1_ver_za(uint32_t slice, svbool_t p, int8_t* ptr) YNN_SME_HELPER {
+  svst1_ver_za8(tile, slice, p, ptr);
+}
+
+// Helpers to call SME2 multi-vector loads and creations via overloading.
+YNN_ALWAYS_INLINE svfloat32x4_t svld1_x4_impl(svcount_t c, const float* p) YNN_SME_HELPER {
```
dsharlet: I don't think you need these wrappers, you can just use `svld1_x4`. arm_sve.h already provides the overloads we need. We only need to provide overloaded wrappers for the intrinsics that don't do this.

ynnpack/kernels/dot/bench.cc
```diff
@@ -82,29 +81,62 @@ void dot(benchmark::State& state, uint64_t arch_flags, dot_kernel_fn kernel,
   c.fill(0);
   b = b.crop_padding({0, 0}, {b.extent(0) - k, b.extent(1) - n});
 
+  // Panel Packing for A if transpose_a is requested.
+  // This simulates the optimal memory layout where A is stored in vertical strips of width `tile_m`.
+  // For widening kernels (tile_k > 1), we must interleave tile_k elements of K.
+  std::vector<TA> a_packed_buffer;
   if (transpose_a) {
-    // This mangles the data, but we don't care here.
-    a = a.reshape({k / tile_k, m * tile_k});
```
dsharlet, line 87: I wonder if https://github.com/google/XNNPACK/blob/8023ac4e79ab8899029b2d4653c9681015766b00/ynnpack/kernels/dot/pack_test_tensor.h#L61 is implementing this logic? I think it might be.

dsharlet (commented):
Thank you for the very interesting work. The results look very nice. We are definitely interested to see this land if possible. Some high level comments:
- If there is no significant downside/tradeoff to this kernel, I'd rather we replace the existing SME2 kernel instead of adding a new one.
- It looks like there are some unrelated changes in the PR. Some of them maybe should be separate PRs?
- I'd rather not submit the benchmark script to the repo.

For reference, I dabbled in trying to pipeline this kernel myself: https://github.com/google/XNNPACK/pull/8979, but it seems like you have gone much further with it and had more success than I did.

kasper0406:
Cool, I will try to split out the tail handling in a separate PR.

While digging a bit deeper, I noticed the proposed kernel only achieves a performance gain when the input matrix is packed (the old kernel did not seem to gain an advantage of the packing from the micro-benchmark). This does not seem to be the case when using the current high-level unpack api. I will try to look into that too. It may take a bit of time before I get to file a proper PR.

dsharlet:
I hope you don't mind, I got impatient and wanted to try this: #9609. I also managed to mess up the PR number, I udpated it but our tooling might take a minute to see that.

kasper0406:
@dsharlet thanks for picking this up. I’ve had a few other things happening - sorry about not getting around to this

PR #9594, opened 2026-02-25
add rvv fp16 kernels for f16-gemm, f16-igemm, f16-dwconv, f16-spmm

ken-unger:
Add RVV FP16 kernels for f16-gemm, f16-igemm, f16-dwconv, f16-spmm
- Relevant tests pass and benchmarks executed.  

Fixed definition of XNN_HAVE_FLOAT16 for RVV in src/xnnpack/math.h

Note that since there is no scalar fallback for FP16 kernels, not all the operator and subgraph tests will execute for RVV FP16 until the remaining kernels are implemented/integrated.  I'll continue with qd8-f16 and other RVV kernels in subsequent PRs.


src/xnnpack/math.h
```diff
@@ -526,7 +526,7 @@ XNN_INLINE static int32_t math_round_f32_to_s32(float x) {
 #define XNN_HAVE_FLOAT16 1
 #endif
 
-#if defined(__riscv) && defined(__riscv_zvfh) && __clang__ >= 1600
+#if defined(__riscv) && defined(__riscv_zvfh)
 #define XNN_HAVE_FLOAT16 1
```
ken-unger, line 530: @dsharlet and @fbarchard it appears this was incorrect, or perhaps the original author intended to use __clang_version__.   I've chosen to remove since there is no preexisting rvv fp16 implementation in xnnpack, and anyone using RVV should be at clang 19 or later (imo).

PR #9935, opened 2026-04-10
Security: fix output loop bound and double-free in subgraph copy

wangw-1991:
This PR fixs two bugs in static subgraph copy:
1. Fix output copy loop using wrong bound: The loop that copies node outputs was iterating with src_node->num_inputs instead of src_node->num_outputs. This could cause out-of-bounds access when num_outputs > num_inputs, or silently skip copying outputs when num_outputs < num_inputs.
2. Fix potential double-free of value data: When copying a value to the destination subgraph, the XNN_VALUE_FLAG_NEEDS_CLEANUP flag was carried over from the source. Since both source and destination values pointed to the same data, destroying both subgraphs would free the same pointer twice. Clear the flag on the copy so only the source subgraph retains ownership.


wangw-1991:
@fbarchard Can you help review this or find a suitable reviewer?

dsharlet (approved):

PR #9940, opened 2026-04-10
Add WASM RELAXED-SIMD F32 RMINMAX microkernels

yolanda15:
Add the missing WASM Relaxed simd f32 rminmax microkernels and update the reduce config to point to it.


fbarchard (approved):

dsharlet (approved):

src/f32-rminmax/f32-rminmax.inc
```diff
@@ -43,7 +43,7 @@ XNN_UKERNEL(xnn_arch_x86_avx512f, xnn_f32_rminmax_ukernel__avx512f_u64_acc2, 64,
 XNN_UKERNEL(xnn_arch_x86_avx512f, xnn_f32_rminmax_ukernel__avx512f_u64_acc4, 64, false, float, float, struct xnn_f32_default_params, ((struct xnn_init_f32_default_params_fn) NULL))
 #endif  // XNN_ENABLE_AVX512F && (XNN_ARCH_X86 || XNN_ARCH_X86_64)
 
-#if XNN_ARCH_WASMSIMD || XNN_ARCH_WASMRELAXEDSIMD
+#if XNN_ARCH_WASMSIMD
```
dsharlet, line 46: This change breaks the wasmdimd build here because when XNN_ARCH_WASMRELAXEDSIMD is enabled, this kernel is not declared: https://github.com/google/XNNPACK/blob/a4673cf1cdc872f47305f61c78456078e9d8fe50/src/configs/reduce-config.c#L1043-L1044

yolanda15: Thanks for the comment! Do you think I should update reduce-config in this PR, or create a seperate PR to update it?

dsharlet: I think it depends on what you update it to. If you want the relaxed simd config to use these new kernels, I would just update it in this PR.

yolanda15: Got it. Thanks! I have uploaded the new change. Please note that it rebases with the latest change on wasm simd template from https://github.com/google/XNNPACK/pull/9970.

fbarchard (approved):

dsharlet (approved):

PR #9974, opened 2026-04-14
[RVV] adding f16-mulcaddc

keaganchern:
Add RVV microkernel for f16-vmulcaddc. A new xngen template (rvvfp16arith.c.in) generates c4v and c8v channel tile variants with row tile 2, consistent with the f32-vmulcaddc RVV implementation.


src/f16-vmulcaddc/gen/f16-vmulcaddc-c8v-minmax-rvvfp16arith-2x.c
```diff
@@ -0,0 +1,86 @@
...
+
+void xnn_f16_vmulcaddc_minmax_ukernel_c8v__rvvfp16arith_2x(
+    size_t rows,
+    size_t channels,
+    const xnn_float16* restrict input,
+    size_t input_stride,
+    const xnn_float16* restrict weights,
+    xnn_float16* restrict output,
+    size_t output_stride,
+    const struct xnn_f16_minmax_params* restrict params) XNN_OOB_READS
```
dsharlet: Does this actually need XNN_OOB_READS?

keaganchern: No, the RVV kernel doesn't need XNN_OOB_READS. thanks for the catch!

dsharlet (approved):

PR #9987, opened 2026-04-17
Add overflow detection to shape multiplication functions

TristanInSec:
## Summary

- Add overflow detection to all five `xnn_shape_multiply_*` variants in `tensor.c`
- Use `__builtin_mul_overflow` to detect when dimension products exceed `SIZE_MAX`
- Return `SIZE_MAX` on overflow instead of a silently wrapped value
- Prevents undersized buffer allocations on platforms where `size_t` is 32-bit (ARM32, WebAssembly)

## Test plan

- [x] Verified overflow is detected for shapes with dimension products exceeding 2^32 on 32-bit
- [x] Verified normal (non-overflowing) shapes return correct results
- [x] All five variants use the same checked multiplication pattern

### DIFF
--- a/src/tensor.c
+++ b/src/tensor.c
@@ -637,12 +637,20 @@ enum xnn_status xnn_define_blockwise_quantized_tensor_value(
                                                         block_size, dims, data, external_id, flags, xnn_datatype_bf16, id_out);
 }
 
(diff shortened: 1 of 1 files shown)


TristanInSec:
Hi, friendly ping on this PR (and the related #9988). Let me know if any changes are needed. Thanks!

PR #10001, opened 2026-04-17
Fix xnn_get_rounded_size to always guarantee XNN_EXTRA_BYTES padding

### LINKED ISSUE #9982: buffer overflow issue after enabling MTE
I encountered the crash issue when executing a model using the benchmark tool.

- Test env : Galaxy S26 with MTE(Memory Tagging Extension) on 
- ML Framework : Tensorfow Lite
- Reproduction ratio : 100%

====================================================================
Cmdline: ./benchmark_model_dbg --graph=model.tflite
pid: 16173, ppid: 16167, tid: 16173, name: benchmark_model  >>> ./benchmark_model_dbg <<<
uid: 0
tagged_addr_ctrl: 000000000007fff7 (PR_TAGGED_ADDR_ENABLE, PR_MTE_TCF_SYNC, PR_MTE_TCF_ASYNC, mask 0xfffe)
pac_enabled_keys: 000000000000000f (PR_PAC_APIAKEY, PR_PAC_APIBKEY, PR_PAC_APDAKEY, PR_PAC_APDBKEY)
esr: 0000000092000011 (Data Abort Exception 0x24)
signal 11 (SIGSEGV), code 9 (SEGV_MTESERR), fault addr 0x0c0000774460d0c0 (read)
Cause: [MTE]: Buffer Overflow, 13 bytes right of a 5523-byte allocation at 0x774460bb20
    x0  020000771460b36c  x1  0c0000774460d0bc  x2  0000000000000014  x3  09000075a4623990
    x4  0c0000774460d0d0  x5  020000771460b380  x6  0000000000000000  x7  0000000000000000
    x8  020000771460b36c  x9  09000075a4623960  x10 0000000000000000  x11 0000000000000003
    x12 0000000000000000  x13 0000000000000000  x14 00000008ba7b352c  x15 0000000034155555
    x16 0000005df4a8e5d0  x17 000000783b16ea40  x18 000000783f8c4000  x19 0000005df3aef060
    x20 0000000000000002  x21 0000000000000002  x22 0000007ff9d52c28  x23 0000007ff9d52c38
    x24 0000000000000000  x25 0000000000000000  x26 0000000000000000  x27 0000000000000000
    x28 0000000000000000  x29 0000007ff9d4fef0
    lr  0000005df47bcccc  sp  0000007ff9d4fed0  pc  000000783b16ea64  pst 0000000020001000
    esr 0000000092000011  vg  0000000000000002

20 total frames
backtrace:
      #00 pc 00000000000a5a64  /apex/com.android.runtime/lib64/bionic/libc.so (__memmove_aarch64_nt+36) (BuildId: 3c40a3f379b085efaaa6d495d43b79ac)
      #01 pc 00000000013c0cc8  /data/local/tmp/benchmark_model_dbg (xnn_xx_copy_ukernel__scalar_memcpy+184) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #02 pc 00000000013271b0  /data/local/tmp/benchmark_model_dbg (xnn_compute_univector_strided+140) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #03 pc 00000000015ef920  /data/local/tmp/benchmark_model_dbg (pthreadpool_parallelize_1d_tile_1d_dynamic+148) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #04 pc 0000000001329b20  /data/local/tmp/benchmark_model_dbg (xnn_run_operator_with_index+1136) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #05 pc 00000000012eed94  /data/local/tmp/benchmark_model_dbg (xnn_invoke_runtime+224) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #06 pc 000000000124e250  /data/local/tmp/benchmark_model_dbg (tflite::xnnpack::(anonymous namespace)::Subgraph::Invoke(TfLiteContext*, bool, tflite::xnnpack::(anonymous namespace)::Delegate*)+1928) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #07 pc 0000000001244e24  /data/local/tmp/benchmark_model_dbg (tflite::xnnpack::(anonymous namespace)::SubgraphInvoke(TfLiteContext*, TfLiteNode*)+120) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #08 pc 000000000128792c  /data/local/tmp/benchmark_model_dbg (tflite::Subgraph::OpInvoke(TfLiteRegistration const&, TfLiteNode*)+348) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #09 pc 00000000012885c4  /data/local/tmp/benchmark_model_dbg (tflite::Subgraph::InvokeImpl()+1300) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #10 pc 0000000001288080  /data/local/tmp/benchmark_model_dbg (tflite::Subgraph::Invoke()+24) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #11 pc 000000000119d1b8  /data/local/tmp/benchmark_model_dbg (tflite::impl::Interpreter::Invoke()+124) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #12 pc 00000000006f5820  /data/local/tmp/benchmark_model_dbg (tflite::benchmark::BenchmarkInterpreterRunner::Invoke()+76) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #13 pc 00000000006ff034  /data/local/tmp/benchmark_model_dbg (tflite::benchmark::BenchmarkTfLiteModel::RunImpl()+28) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #14 pc 0000000000730268  /data/local/tmp/benchmark_model_dbg (tflite::benchmark::BenchmarkModel::Run(int, float, float, tflite::benchmark::RunType, TfLiteStatus*)+604) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #15 pc 0000000000730be8  /data/local/tmp/benchmark_model_dbg (tflite::benchmark::BenchmarkModel::Run()+1176) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #16 pc 00000000007306fc  /data/local/tmp/benchmark_model_dbg (tflite::benchmark::BenchmarkModel::Run(int, char**)+96) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #17 pc 00000000006f2c0c  /data/local/tmp/benchmark_model_dbg (tflite::benchmark::Main(int, char**)+140) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #18 pc 00000000006f3080  /data/local/tmp/benchmark_model_dbg (main+32) (BuildId: f70fa405d8e9f47545267648ab2f5657)
      #19 pc 000000000006e984  /apex/com.android.runtime/lib64/bionic/libc.so (__libc_init+124) (BuildId: 3c40a3f379b085efaaa6d495d43b79ac)

Note: To display stack pointer information, use the pbtombstone tool:
        pbtombstone --display-sp tombstone_XX.pb

Memory tags around the fault address (0xc0000774460d0c0), one tag per 16 bytes:
      0x774460c800: c  c  c  c  c  c  c  c  c  c  c  c  c  c  c  c
      0x774460c900: c  c  c  c  c  c  c  c  c  c  c  c  c  c  c  c
      0x774460ca00: c  c  c  c  c  c  c  c  c  c  c  c  c  c  c  c
      0x774460cb00: c  c  c  c  c  c  c  c  c  c  c  c  c  c  c  c
      0x774460cc00: c  c  c  c  c  c  c  c  c  c  c  c  c  c  c  c
      0x774460cd00: c  c  c  c  c  c  c  c  c  c  c  c  c  c  c  c
      0x774460ce00: c  c  c  c  c  c  c  c  c  c  c  c  c  c  c  c
      0x774460cf00: c  c  c  c  c  c  c  c  c  c  c  c  c  c  c  c
    =>0x774460d000: c  c  c  c  c  c  c  c  c  c  c  c [0] 0  0  0
      0x774460d100: 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
      0x774460d200: 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
      0x774460d300: 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
      0x774460d400: 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
      0x774460d500: 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
      0x774460d600: 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  3
      0x774460d700: 3  3  3  3  3  3  3  3  3  3  3  3  3  3  3  3

====================================================================

Could you let me know if there's know issue regarding this one?
And please let me know how I can share the tflite model for your reference.

Thank you,
SangJun

ssam18:
The old formula `round_up_po2(round_up_po2(size, XNN_EXTRA_BYTES), alignment)` silently provided zero over-read headroom whenever a tensor's byte size was already a multiple of `XNN_EXTRA_BYTES`, which causes `SEGV_MTESERR` on MTE-enabled ARM64 devices when a kernel reads past the end of one tensor into a differently-tagged adjacent allocation. Changing the formula to `round_up_po2(size + XNN_EXTRA_BYTES, alignment)` guarantees at least `XNN_EXTRA_BYTES` of padding after every XNNPACK-managed tensor on all architectures. Verified against ARM64, x86-64, Hexagon, and WASM alignment configs — no case regresses in allocated size. Fixes #9982

### DIFF
--- a/src/xnnpack/subgraph.h
+++ b/src/xnnpack/subgraph.h
@@ -554,11 +554,11 @@ size_t xnn_tensor_get_size(const struct xnn_value* value);
 size_t xnn_runtime_tensor_get_size(const struct xnn_runtime_value* value);
 
 XNN_INLINE static size_t xnn_get_rounded_size(size_t size) {
-  // We round it to XNN_EXTRA_BYTES to ensure that we can read more than the
-  // actual size of the tensor, and round it to allocation alignment to ensure
-  // that all tensors and operator workspaces are aligned correctly.
-  return round_up_po2(round_up_po2(size, XNN_EXTRA_BYTES),
-                      XNN_ALLOCATION_ALIGNMENT);
+  // Add XNN_EXTRA_BYTES of padding after the data so that XNNPACK kernels can
+  // over-read by up to XNN_EXTRA_BYTES without triggering memory-safety faults
+  // (e.g. MTE on ARM64). Merely rounding up to a multiple of XNN_EXTRA_BYTES
+  // provides no padding for sizes that are already aligned.
+  return round_up_po2(size + XNN_EXTRA_BYTES, XNN_ALLOCATION_ALIGNMENT);
 }
 
 // Returns the size of tensor rounded to appropriate extra bytes and allocation


dsharlet:
Thanks for the PR. 

I think this is not the right fix for the issue. We already add 2x `XNN_EXTRA_BYTES` in the workspace:  https://github.com/google/XNNPACK/blob/1c7554a8813d400472ce36ed173fca9db2f65408/src/runtime.c#L242-L243

We should not need to add `XNN_EXTRA_BYTES` to every tensor, because we only need this extra memory for reading.

Do you know more about the issue? Maybe you can share the model? It looks like it is possible to attach files to a comment here.

ssam18:
> Thanks for the PR.
> 
> I think this is not the right fix for the issue. We already add 2x `XNN_EXTRA_BYTES` in the workspace:
> 
> https://github.com/google/XNNPACK/blob/1c7554a8813d400472ce36ed173fca9db2f65408/src/runtime.c#L242-L243
> 
> We should not need to add `XNN_EXTRA_BYTES` to every tensor, because we only need this extra memory for reading.
> 
> Do you know more about the issue? Maybe you can share the model? It looks like it is possible to attach files to a comment here.

Thanks for the pushback. You're right! I traced through the formulas with XNN_EXTRA_BYTES=128 for the specific 5523-byte case in #9982 and the old round_up_po2(round_up_po2(size, 128), 64) already gives 109 bytes of slack, identical to what my version produces, so this PR doesn't actually change that case. And since the workspace is one allocation with one MTE tag, over-reads inside it shouldn't trigger SEGV_MTESERR anyway.

The crash report shows the read crossed into a differently-tagged region (c c ... [0] 0 0), which means the offending allocation is external (TFLite-allocated, one tag per tensor) rather than a workspace tensor. So xnn_get_rounded_size isn't really on the path here.

I don't have the failing model, only the report from #9982. Could you ping the issue author for it? Once we have it I can verify whether the fix should be at the copy ukernel (do not over-read past externally-supplied tensors) or at the TFLite delegate side. Happy to close this PR in the meantime if that is cleaner.

PR #10101, opened 2026-04-28
Reject qpint8 in xnn_define_dynamically_quantized_tensor_value

### LINKED ISSUE #10099: NULL deref in `get_tensor_size` when defining a `qpint8` tensor
`xnn_define_dynamically_quantized_tensor_value` accepts `xnn_datatype_qpint8` in its switch at `src/tensor.c:262`, then calls `xnn_tensor_get_size`, which routes into the `qpint8` branch of `get_tensor_size`:

```c
// src/tensor.c:769-781
if (datatype == xnn_datatype_qpint8) {
  assert(gemm_config != NULL);
  size_t num_groups = xnn_shape_multiply_batch_dims(shape, 2);
  size_t m = shape->dim[shape->num_dims - 2];
  const size_t k = shape->dim[shape->num_dims - 1];
  ...
  return num_groups *
         xnn_x8_packq_f32qp8_gemm_packed_size(gemm_config, m, k);
}
```

Nothing on the dynamically-quantized define path sets `value->gemm_config`, so it is the zero-initialized NULL. Debug builds abort on the assert; Release builds dereference NULL inside `xnn_x8_packq_f32qp8_gemm_packed_size` (`src/xnnpack/packq.h:62`):

```c
const uint32_t mr_packed = m == 1                   ? 1
                           : gemm_config->mr_packed ? gemm_config->mr_packed
                                                    : gemm_config->mr;
```

`qpint8` works through other code paths that resolve and assign the gemm config first; the dynamically-quantized define path is not one of them.

## Repro

Three lines of public API on top of `xnn_initialize` / `xnn_create_subgraph` are enough:

```c
size_t dims[2] = {4, 8};
uint32_t id = XNN_INVALID_VALUE_ID;
xnn_define_dynamically_quantized_tensor_value(
    sg, xnn_datatype_qpint8, /*num_dims=*/2, /*num_nonbatch=*/2,
    dims, XNN_INVALID_VALUE_ID, /*flags=*/0, &id);
```

Debug build under ASan/UBSan:

```
Assertion failed: (gemm_config != NULL), function get_tensor_size, file tensor.c, line 771.
```

Reproduces against `master` at `172b7f4a01`.

## Fix

Easiest is to reject `qpint8` in the `xnn_define_dynamically_quantized_tensor_value` switch and return `xnn_status_unsupported_parameter`. Happy to open a PR for fix.

MarkLee131:
Fix https://github.com/google/XNNPACK/issues/10099:
Calling xnn_define_dynamically_quantized_tensor_value with datatype=xnn_datatype_qpint8 lands in get_tensor_size at tensor.c:770, which asserts gemm_config != NULL. Nothing on the dynamically-quantized define path sets value->gemm_config, so the assert fires in Debug; in Release the NULL is dereferenced inside
xnn_x8_packq_f32qp8_gemm_packed_size (packq.h:62), which reads gemm_config->mr_packed and friends.

qpint8 only works through other code paths that wire up gemm_config first. The dynamically-quantized define path is not one of them, so reject it here with xnn_status_unsupported_parameter rather than letting it crash later.

### DIFF
--- a/src/tensor.c
+++ b/src/tensor.c
@@ -261,11 +261,10 @@ enum xnn_status xnn_define_dynamically_quantized_tensor_value(
     case xnn_datatype_qdint8:
       break;
     case xnn_datatype_qpint8:
-      // qpint8 reaches get_tensor_size() at tensor.c:770, which dereferences
-      // value->gemm_config (left NULL by this define path). Reject until the
-      // gemm-config wiring is added.
+      // qpint8 is an internal datatype produced downstream of public types
+      // like qdint8; it is not valid as an input to the public define API.
       xnn_log_error("failed to create Dynamically Quantized Dense Tensor value: "
-        "datatype %s (%d) is not supported on this code path",
+        "datatype %s (%d) is an internal type and cannot be used through the public API",
         xnn_datatype_to_string(datatype), datatype);
       return xnn_status_unsupported_parameter;
     default:
--- a/src/tensor.c
+++ b/src/tensor.c
@@ -259,8 +259,15 @@ enum xnn_status xnn_define_dynamically_quantized_tensor_value(
 
   switch (datatype) {
     case xnn_datatype_qdint8:
-    case xnn_datatype_qpint8:
       break;
+    case xnn_datatype_qpint8:
+      // qpint8 reaches get_tensor_size() at tensor.c:770, which dereferences
+      // value->gemm_config (left NULL by this define path). Reject until the
+      // gemm-config wiring is added.
+      xnn_log_error("failed to create Dynamically Quantized Dense Tensor value: "
+        "datatype %s (%d) is not supported on this code path",
+        xnn_datatype_to_string(datatype), datatype);
+      return xnn_status_unsupported_parameter;
     default:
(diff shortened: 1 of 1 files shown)


Aelphy (approved):

src/tensor.c
```diff
@@ -259,8 +259,15 @@ enum xnn_status xnn_define_dynamically_quantized_tensor_value(
...
   switch (datatype) {
     case xnn_datatype_qdint8:
-    case xnn_datatype_qpint8:
       break;
+    case xnn_datatype_qpint8:
+      // qpint8 reaches get_tensor_size() at tensor.c:770, which dereferences
+      // value->gemm_config (left NULL by this define path). Reject until the
+      // gemm-config wiring is added.
+      xnn_log_error("failed to create Dynamically Quantized Dense Tensor value: "
+        "datatype %s (%d) is not supported on this code path",
```
dsharlet: I think the comment and error message should be adjusted. I think qpint8 simply is not a valid public API type at all. It should only be used internally, it is a downstream result of using qdint8 or other public API types.

MarkLee131: Updated the comment and the error message to say qpint8 is an internal type that can't be used through the public define API.

dsharlet (approved):

PR #10102, opened 2026-04-28
Detect size_t overflow in get_tensor_size and reject the tensor

### LINKED ISSUE #10100: Integer overflow in `get_tensor_size` undersizes the workspace arena
`xnn_shape_multiply_all_dims` and the `size_bits *= ...` step in `get_tensor_size` use unchecked `size_t` arithmetic. With shapes that fit inside `XNN_MAX_TENSOR_DIMS`, the product can wrap to a small non-zero value, which then becomes `value->size`.

```c
// src/tensor.c:640-647
size_t xnn_shape_multiply_all_dims(const struct xnn_shape* shape) {
  size_t batch_size = 1;
  for (size_t i = 0; i < shape->num_dims; i++) {
    batch_size *= shape->dim[i];           // unchecked
  }
  return batch_size;
}

// src/tensor.c:783-790
uint64_t size_bits = xnn_datatype_size_bits(datatype);
size_bits *= xnn_shape_multiply_all_dims(shape);   // unchecked
return (size_bits + 7) >> 3;
```

`xnn_plan_memory` at `src/runtime.c:826` uses `value->size` (via `xnn_tensor_get_rounded_size`) to size the workspace arena for internal tensors, so the arena ends up undersized while the kernel still iterates the real dim count. Any caller that uses `xnn_tensor_get_size` to size its own external buffers gets the same mis-sizing.

#9860 already adds `xnn_safe_mul` / `xnn_safe_add` helpers and applies them in several operator reshape paths, but it does not touch `src/tensor.c`, so this overflow chain is still live on master.

## Repro

`(2^59 + 1)` element 1-D fp32 tensor wraps the byte count to 4:

```c
size_t dims[1] = { ((size_t)1 << 59) + 1 };
xnn_define_tensor_value(sg, xnn_datatype_fp32, 1, dims, NULL,
                        0, XNN_VALUE_FLAG_EXTERNAL_INPUT, &in);
// subgraph->values[in].size == 4   (true byte count: 2305843009213693956)
```

Chain it through a unary op and the kernel goes past the buffer:

```
==58557==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60200000017f
READ of size 16 at 0x60200000017f thread T0
    #0 xnn_f32_vabs_ukernel__neon_u8 f32-vabs-neon.c:66
    #1 pthreadpool_parallelize_1d_tile_1d_dynamic portable-api.c:3653
    #2 xnn_run_operator_with_index operator-run.c:1963
    #3 xnn_invoke_runtime runtime.c:1138
0x60200000017f is located 11 bytes after 4-byte region
```

Same wrap shows up in the internal arena: a graph that routes the tensor through an internal value gets `created workspace of size 48` for a tensor whose real byte count is ~2 EiB. Reproduces against `master` at `172b7f4a01`.

## Fix

Build on top of #9860's helpers:

- Switch all five `xnn_shape_multiply_*_dims` helpers to `xnn_safe_mul` and return `SIZE_MAX` on overflow.
- Same treatment for the `size_bits` step in `get_tensor_size`.
- Reject the tensor at all five `xnn_define_*` call sites and at `xnn_reshape_external_value` when the computed size is `SIZE_MAX`.

Happy to open a PR for fixing

MarkLee131:
Fix https://github.com/google/XNNPACK/issues/10100:

- Add xnn_safe_mul / xnn_safe_add helpers in math.h. Same shape as the helpers proposed in #9860, so the two PRs can share them once one of them lands.
- Use xnn_safe_mul in all five xnn_shape_multiply_*_dims helpers and in the size_bits step of get_tensor_size; return SIZE_MAX on overflow.
- Reject the tensor with xnn_status_unsupported_parameter at the five xnn_define_* call sites and at xnn_reshape_external_value when the computed size is SIZE_MAX.

This stacks cleanly on top of #9860; happy to rebase or fold once that lands.

### DIFF
--- a/src/tensor.c
+++ b/src/tensor.c
@@ -816,24 +816,15 @@ size_t get_tensor_size(const struct xnn_gemm_config* gemm_config, enum xnn_value
   // Returns SIZE_MAX on overflow so the caller can detect and reject the tensor
   // before the wrapped value is treated as an allocation size.
   const size_t num_elements = xnn_shape_multiply_all_dims(shape);
-  if (num_elements == SIZE_MAX) {
-    return SIZE_MAX;
-  }
-  uint64_t size_bits = xnn_datatype_size_bits(datatype);
-  if (size_bits != 0 && num_elements > UINT64_MAX / size_bits) {
-    return SIZE_MAX;
-  }
-  size_bits *= (uint64_t) num_elements;
-
-  // Round size up to the nearest byte. (size_bits + 7) cannot wrap because we
-  // already bounded size_bits via the check above.
-  const uint64_t size_bytes = (size_bits + 7) >> 3;
-  if (size_bytes > (uint64_t) SIZE_MAX) {
+  size_t size_bits;
+  if (num_elements == SIZE_MAX ||
+      !xnn_safe_mul(num_elements, xnn_datatype_size_bits(datatype), &size_bits) ||
+      !xnn_safe_add(size_bits, 7, &size_bits)) {
     return SIZE_MAX;
   }
   // TODO: We should not be using this helper for non-byte-addressable types,
   // perhaps we should just assert here.
-  return (size_t) size_bytes;
+  return size_bits >> 3;
 }
 
 size_t xnn_runtime_tensor_get_size(const struct xnn_runtime_value* value)
--- a/src/tensor.c
+++ b/src/tensor.c
@@ -153,6 +153,10 @@ enum xnn_status xnn_define_tensor_value(
   value->datatype = datatype;
   set_shape(value, num_dims, dims);
   value->size = xnn_tensor_get_size(value);
+  if (value->size == SIZE_MAX) {
+    xnn_log_error("failed to create Dense Tensor value: size overflow for the given shape");
+    return xnn_status_unsupported_parameter;
+  }
   value->flags = flags;
   value->data = (void*) (uintptr_t) data;
   set_allocation_type(value);
@@ -209,6 +213,10 @@ enum xnn_status xnn_define_quantized_tensor_value(
   value->quantization.scale = scale;
   set_shape(value, num_dims, dims);
   value->size = xnn_tensor_get_size(value);
+  if (value->size == SIZE_MAX) {
+    xnn_log_error("failed to create Quantized Dense Tensor value: size overflow for the given shape");
+    return xnn_status_unsupported_parameter;
+  }
   value->flags = flags;
   value->data = (void*) (uintptr_t) data;
   set_allocation_type(value);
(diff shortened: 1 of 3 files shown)


src/tensor.c
```diff
@@ -780,14 +813,27 @@ size_t get_tensor_size(const struct xnn_gemm_config* gemm_config, enum xnn_value
...
+  // before the wrapped value is treated as an allocation size.
+  const size_t num_elements = xnn_shape_multiply_all_dims(shape);
+  if (num_elements == SIZE_MAX) {
+    return SIZE_MAX;
+  }
   uint64_t size_bits = xnn_datatype_size_bits(datatype);
+  if (size_bits != 0 && num_elements > UINT64_MAX / size_bits) {
+    return SIZE_MAX;
+  }
+  size_bits *= (uint64_t) num_elements;
```
dsharlet: How about just using `xnn_safe_mul` here instead of the extra logic above?

dsharlet (commented):
Thanks for the PR.

src/tensor.c
MarkLee131: Done, swapped the open-coded uint64 check for xnn_safe_mul on num_elements * size_bits and xnn_safe_add for the +7 round-up. Falls through to SIZE_MAX on either overflow.

src/tensor.c
```diff
@@ -780,14 +813,18 @@ size_t get_tensor_size(const struct xnn_gemm_config* gemm_config, enum xnn_value
...
-  size_bits *= xnn_shape_multiply_all_dims(shape);
-
-  // Round size up to the nearest byte.
+  // Returns SIZE_MAX on overflow so the caller can detect and reject the tensor
+  // before the wrapped value is treated as an allocation size.
+  const size_t num_elements = xnn_shape_multiply_all_dims(shape);
+  size_t size_bits;
+  if (num_elements == SIZE_MAX ||
+      !xnn_safe_mul(num_elements, xnn_datatype_size_bits(datatype), &size_bits) ||
+      !xnn_safe_add(size_bits, 7, &size_bits)) {
```
dsharlet: I might be being paranoid here but IIRC some compilers (IIRC MSVC 32-bit) might miscompile this because they would pass the value of `size_bits` from before `xnn_safe_mul` to `xnn_safe_add`. So I think we should break this into two separate satements:
```
if (!xnn_safe_mul(num_elements, xnn_datatype_size_bits(datatype), &size_bits)) {
  return SIZE_MAX;
}
if (!xnn_safe_add(size_bits, 7, &size_bits)) {
  return SIZE_MAX;
}
```

MarkLee131: OK, split into two separate ifs so each size_bits write is sequenced before the next read.

dsharlet (approved):

PR #10189, opened 2026-05-06
Clarify explicit-false build option alias behavior

mike-lmctl:
This clarifies the intentionally counterintuitive `define_build_option` alias behavior in `ynnpack/build_defs.bzl`.

The public alias is used as a positive condition in downstream `select()` calls. I tested the suggested one-line change against `//ynnpack/base:base` and it would make `--define ynn_enable_cpuinfo=false` still select the CPUINFO path. The existing mapping keeps that condition false.

Verified with Docker + Bazel 9.1.0:
- original mapping, `--define ynn_enable_cpuinfo=true`: `YNN_ENABLE_CPUINFO` present
- original mapping, `--define ynn_enable_cpuinfo=false`: `YNN_ENABLE_CPUINFO` absent
- changed mapping, `--define ynn_enable_cpuinfo=false`: `YNN_ENABLE_CPUINFO` present

So this PR is now documentation only: it preserves the current behavior and explains why the explicit-false branch intentionally routes to the explicit-true config setting.

### DIFF
--- a/ynnpack/build_defs.bzl
+++ b/ynnpack/build_defs.bzl
@@ -39,6 +39,9 @@ def define_build_option(name, default_all = [], default_any = []):
         name = name,
         actual = select({
             explicit_true: ":" + explicit_true,
+            # This alias is used as a positive condition in downstream select()
+            # calls. When the option is explicitly false, route the alias to the
+            # explicit-true config_setting so the public condition does not match.
             explicit_false: ":" + explicit_true,
             "//conditions:default": ":" + default,
         }),


mike-lmctl:
Reproduced the issue by inspecting `define_build_option`: the `explicit_false` select branch aliased to `:<name>_explicit_true`, so an explicit `--define <name>=false` could select the true target. The fix changes that branch to alias to `:<name>_explicit_false`. I verified the exact bad mapping is gone, the corrected mapping is present, and the diff has no whitespace errors. Bazel analysis/tests were not run because Bazel/Bazelisk are unavailable in this environment.

dsharlet:
I know this is weird but I think it's a correct bazel-ism. The way bazel select works means that if the key "matches", then take the value.

So what this means is that if the explicit_false flag is set, then the setting is considered true if the explicit_true value is also set.

Does this change actually fix a problem for you?

mike-lmctl:
You were right. I went back and reproduced this with Bazel instead of relying on static inspection, and the original semantic change was wrong.

The key point is that this alias is used as a positive condition in downstream `select()` calls. With the current mapping, `--define ynn_enable_cpuinfo=false` keeps `//ynnpack:ynn_enable_cpuinfo` from matching, so `YNN_ENABLE_CPUINFO` is not set. With my original change to route explicit false to the explicit-false config setting, that positive condition matches and incorrectly enables `YNN_ENABLE_CPUINFO`.

I verified this two ways with Docker + Bazel 9.1.0:

- on XNNPACK `//ynnpack/base:base`: current mapping keeps `YNN_ENABLE_CPUINFO` absent under `--define ynn_enable_cpuinfo=false`; the changed mapping makes it present
- with a minimal alias/select repro: under `--define opt=false`, the current mapping selects the false/default branch, while the changed mapping selects the true branch

I force-pushed this PR so it no longer changes behavior. It only adds a short comment documenting why the counterintuitive mapping is intentional.

PR #10199, opened 2026-05-08
Use safe tail loads for inputs in maxpool kernels

wangw-1991:
The maxpool kernel uses `xnn_load_tail_impl` for input reads in the tail section. On SSE2 for example, `xnn_load_tail_u8` performs a full 16-byte `_mm_loadu_si128` even when fewer bytes are valid. This out-of-bounds read is annotated with `XNN_OOB_READS` to suppress sanitizer warnings, but it can cause a real SEGV when the input tensor's last valid bytes are near a page boundary and the overread crosses into unmapped memory.

This PR fixes the issue by switching the tail section input loads from `xnn_load_tail_impl` to `xnn_load_tail_safe_impl`. This change affects both the template (maxpool.c.in) and all generated maxpool kernels (f16, f32, s8, u8) across SSE2, SSE4.1, AVX2, NEON, WASM SIMD, and HVX backends.

Chromium issue: https://issues.chromium.org/issues/506960160

### DIFF
--- a/src/f16-maxpool/gen/f16-maxpool-9p-minmax-avx2-u16.c
+++ b/src/f16-maxpool/gen/f16-maxpool-9p-minmax-avx2-u16.c
@@ -16,7 +16,6 @@
 // On some architectures, we have max(u8, u8) but not max(s8, s8). We can emulate max(s8, s8) on these architectures by
 // xoring with the sign bit mask.
 #define xnn_load_impl(x) xnn_signcomplement_s16(xnn_loadu_s16(x))
-#define xnn_load_tail_impl(x, c) xnn_signcomplement_s16(xnn_load_tail_s16(x, c))
 #define xnn_load_tail_safe_impl(x, c) xnn_signcomplement_s16(xnn_load_tail_safe_s16(x, c))
 #define xnn_pre_store_impl(x) xnn_signcomplement_s16(x)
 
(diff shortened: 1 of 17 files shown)


wangw-1991:
@dsharlet Can you help review this?

dsharlet:
This shouldn't be necessary. XNNPACK requires allocations to have `XNN_EXTRA_BYTES` extra bytes allocated, which makes these over-reads safe. So this change is a workaround for a bug where those extra bytes are not present. Do you have a reproducer for how to hit this crash?

wangw-1991:
> This shouldn't be necessary. XNNPACK requires allocations to have `XNN_EXTRA_BYTES` extra bytes allocated, which makes these over-reads safe. So this change is a workaround for a bug where those extra bytes are not present. Do you have a reproducer for how to hit this crash?

Yes, you can find the reproducer in the Chromium issue: https://issues.chromium.org/issues/506960160

wangw-1991:
Hi, @dsharlet Any update about this?

wangw-1991:
> This shouldn't be necessary. XNNPACK requires allocations to have `XNN_EXTRA_BYTES` extra bytes allocated, which makes these over-reads safe. So this change is a workaround for a bug where those extra bytes are not present. Do you have a reproducer for how to hit this crash?

hi, @dsharlet. I think the root cause is WebNN does not include XNN_EXTRA_BYTES padding in tensor buffer sizes when passing to LiteRT/TFLite. I have a CL to fix in chromium: https://chromium-review.googlesource.com/c/chromium/src/+/7848797. I can close this PR when that is merged, thanks.

PR #10260, opened 2026-05-16
fix: check for size_t overflow in constant-pad output dimension calculation

adilburaksen:
## Summary

`reshape_constant_pad_nd()` in `src/operators/constant-pad-nd.c` computes the output shape for each dimension as:

```c
normalized_output_shape[d] = pre_padding + input_dim + post_padding;
```

When `pre_padding` or `post_padding` is near `SIZE_MAX`, the unchecked `size_t` addition wraps around silently.

## Vulnerable input

```
num_dims = 1, input_shape[0] = 1
pre_paddings[0] = SIZE_MAX, post_paddings[0] = 1
```

Result: `SIZE_MAX + 1 + 1 = 1` (wraps). The operator stores `SIZE_MAX` as `pre_paddings[0]` but the output buffer is allocated for only 1 element. The pad microkernel then issues a write of `SIZE_MAX` bytes past the end of the output buffer — a heap OOB write.

## ASan repro

```c
// Compile: gcc -fsanitize=address -O0 -o /tmp/poc /tmp/poc.c
// Wrap: pre_paddings[0]=SIZE_MAX, post_paddings[0]=1 → output_shape wraps to 1
// malloc(1) → memset(output, 0x00, SIZE_MAX) → ASan: WRITE of size N at 0 bytes after 1-byte region
```

## Fix

Use the existing `xnn_safe_add()` helper (wraps `__builtin_add_overflow`) for both additions and return `xnn_status_invalid_parameter` if either overflows. The include for `src/xnnpack/math.h` is added to expose `xnn_safe_add` in this translation unit.

### DIFF
--- a/src/operators/constant-pad-nd.c
+++ b/src/operators/constant-pad-nd.c
@@ -184,9 +184,17 @@ static enum xnn_status reshape_constant_pad_nd(
 
     const bool is_current_dim_padded = (pre_padding | post_padding) != 0;
     if (is_current_dim_padded || is_previous_dim_padded) {
-      size_t pre_input_sum, output_dim;
-      if (!xnn_safe_add(pre_padding, input_dim, &pre_input_sum) ||
-          !xnn_safe_add(pre_input_sum, post_padding, &output_dim)) {
+      size_t pre_input_sum;
+      if (!xnn_safe_add(pre_padding, input_dim, &pre_input_sum)) {
+        xnn_log_error(
+            "failed to setup %s operator: output dimension overflow in "
+            "dimension %zu (pre_padding=%zu + input_dim=%zu overflows size_t)",
+            xnn_operator_type_to_string_v2(constant_pad_op),
+            num_dims - 1 - i, pre_padding, input_dim);
+        return xnn_status_invalid_parameter;
+      }
+      size_t output_dim;
+      if (!xnn_safe_add(pre_input_sum, post_padding, &output_dim)) {
         xnn_log_error(
             "failed to setup %s operator: output dimension overflow in "
             "dimension %zu (pre_padding=%zu + input_dim=%zu + "
--- a/src/operators/constant-pad-nd.c
+++ b/src/operators/constant-pad-nd.c
@@ -184,9 +184,9 @@ static enum xnn_status reshape_constant_pad_nd(
 
     const bool is_current_dim_padded = (pre_padding | post_padding) != 0;
     if (is_current_dim_padded || is_previous_dim_padded) {
-      size_t output_dim;
-      if (!xnn_safe_add(pre_padding, input_dim, &output_dim) ||
-          !xnn_safe_add(output_dim, post_padding, &output_dim)) {
+      size_t pre_input_sum, output_dim;
+      if (!xnn_safe_add(pre_padding, input_dim, &pre_input_sum) ||
+          !xnn_safe_add(pre_input_sum, post_padding, &output_dim)) {
         xnn_log_error(
             "failed to setup %s operator: output dimension overflow in "
             "dimension %zu (pre_padding=%zu + input_dim=%zu + "
--- a/src/operators/constant-pad-nd.c
+++ b/src/operators/constant-pad-nd.c
@@ -16,6 +16,7 @@
 #include "src/xnnpack/config-types.h"
 #include "src/xnnpack/config.h"
 #include "src/xnnpack/log.h"
+#include "src/xnnpack/math.h"
 #include "src/xnnpack/operator-type.h"
 #include "src/xnnpack/operator-utils.h"
 #include "src/xnnpack/operator.h"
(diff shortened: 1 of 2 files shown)


src/operators/constant-pad-nd.c
```diff
@@ -183,9 +184,20 @@ static enum xnn_status reshape_constant_pad_nd(
 
     const bool is_current_dim_padded = (pre_padding | post_padding) != 0;
     if (is_current_dim_padded || is_previous_dim_padded) {
+      size_t output_dim;
+      if (!xnn_safe_add(pre_padding, input_dim, &output_dim) ||
```
dsharlet: I've seen some old windows compilers that don't handle this correctly (output parameter in one statement is used as an input parameter in the same statement to another call). I think it would be best to split this into two separate statements.

adilburaksen: Good catch — fixed by splitting into two separate statements using a distinct intermediate variable `pre_input_sum`. No variable now appears as both input and output in the same call.

dsharlet: I don't see any difference in the change you uploaded.

adilburaksen (commented):
The difference is in the overflow handling. With the original code, `pre_padding + input_dim + post_padding` is unsigned `size_t` arithmetic — on overflow it silently wraps to a small value. For example:

```
pre_padding  = SIZE_MAX - 1
input_dim    = 2
post_padding = 0

pre_padding + input_dim = SIZE_MAX - 1 + 2 → wraps to 0
output_dim  = 0
```

A 0-byte buffer gets allocated, but the microkernel then tries to write `pre_padding` (≈ 2^64) bytes into it — a heap out-of-bounds write. These values can come from model-supplied padding arrays and are not range-checked before reaching `reshape_constant_pad_nd`.

`xnn_safe_add` uses `__builtin_add_overflow` to detect the overflow and return `xnn_status_invalid_parameter` before any allocation occurs.

Added a test case in the latest commit (`7fd2d09`) that confirms `xnn_reshape_constant_pad_nd_x32` returns `xnn_status_invalid_parameter` for `pre_padding = SIZE_MAX - 1`.

adilburaksen:
The difference is in the overflow handling. With the original code, `pre_padding + input_dim + post_padding` is unsigned `size_t` arithmetic — on overflow it silently wraps to a small value. For example:

```
pre_padding  = SIZE_MAX - 1
input_dim    = 2
post_padding = 0

pre_padding + input_dim = SIZE_MAX - 1 + 2  →  wraps to 0
output_dim  = 0
```

A 0-byte buffer gets allocated, but the microkernel then tries to write `pre_padding` (≈ 2^64) bytes into it — a heap out-of-bounds write. These values can come from model-supplied padding arrays and are not range-checked before reaching `reshape_constant_pad_nd`.

`xnn_safe_add` uses `__builtin_add_overflow` to detect the overflow and return `xnn_status_invalid_parameter` before any allocation occurs.

Added a test case in the latest commit (7fd2d09) that confirms `xnn_reshape_constant_pad_nd_x32` returns `xnn_status_invalid_parameter` for `pre_padding = SIZE_MAX - 1`.

src/operators/constant-pad-nd.c
adilburaksen: The difference is in the overflow handling. With the original code, pre_padding + input_dim + post_padding is unsigned size_t arithmetic — on overflow it silently wraps to a small value. For example:

pre_padding  = SIZE_MAX - 1
input_dim    = 2
post_padding = 0

pre_padding + input_dim = SIZE_MAX - 1 + 2 → wraps to 0
output_dim  = 0
A 0-byte buffer gets allocated, but the microkernel then tries to write pre_padding (≈ 2^64) bytes into it — a heap out-of-bounds write. These values can come from model-supplied padding arrays and are not range-checked before reaching reshape_constant_pad_nd.

xnn_safe_add uses __builtin_add_overflow to detect the overflow and return xnn_status_invalid_parameter before any allocation occurs.

Added a test case in the latest commit (7fd2d09) that confirms xnn_reshape_constant_pad_nd_x32 returns xnn_status_invalid_parameter for pre_padding = SIZE_MAX - 1.

dsharlet: I meant specifically this feedback:

> I've seen some old windows compilers that don't handle this correctly (output parameter in one statement is used as an input parameter in the same statement to another call). I think it would be best to split this into two separate statements.

There was no change after this comment. The code still uses `f(&a) || g(a)`

src/operators/constant-pad-nd.c
```diff
@@ -16,6 +16,7 @@
 #include "src/xnnpack/config-types.h"
```
adilburaksen, line 16: Thanks for the clarification. The previous commit missed that —  was still being written and read within the same `||` expression (`f(&a) || g(a)`).

Fixed in d8b8d32: split into two completely independent `if` blocks so `pre_input_sum` is fully committed before being used as an input to the second call. The test case added in 7fd2d09 still passes.

adilburaksen:
@dsharlet friendly ping on this one — the aliased `xnn_safe_add` concern you raised is addressed in d8b8d32: the two `xnn_safe_add` calls are now in separate `if` blocks, so no output parameter is read as an input within the same statement (`pre_input_sum` is fully committed before it's passed to the second call). 7fd2d09 also adds an overflow regression test that confirms `xnn_reshape_constant_pad_nd_x32` returns `xnn_status_invalid_parameter` for `pre_padding = SIZE_MAX - 1`.

Is there anything else you'd like changed before this can be merged? Happy to rebase if the copybara import needs a refresh. Thanks for the reviews.

adilburaksen:
Heads-up: it looks like an equivalent fix landed on `master` in 8d8fa1fe ("reject overflowing output dimensions in constant pad reshape", 2026-05-31). It applies the same `xnn_safe_add` overflow guard in `reshape_constant_pad_nd`, and additionally covers the subgraph path in `static-constant-pad.c` (which this PR didn't touch) — so the overflow this PR targets is now handled on `master`, which is also why the branch now shows a conflict in `constant-pad-nd.c`.

I'm happy to either close this as superseded or rebase it down to anything not already covered — whichever you prefer. Either way the original issue (a `SIZE_MAX` padding wrapping the output-dim allocation in `reshape_constant_pad_nd`) is now fixed on `master`. Thanks for the reviews @dsharlet.

adilburaksen:
Closing this — the same overflow guard has since landed on `master` in commit `8d8fa1fe` ("reject overflowing output dimensions in constant pad reshape", 2026-05-31), which applies `xnn_safe_add` to `reshape_constant_pad_nd` (the same approach and location proposed here, opened 2026-05-23) and additionally covers the subgraph path in `static-constant-pad.c`.

Since the fix is already in, this PR is redundant. Thanks to @aizu-m for landing it. The issue is resolved on master.

PR #10280, opened 2026-05-19
Add Hexagon support on top of YNNPACK CMake build changes

ejparkqc:
Added Hexagon-specific changes on top of the CMake build from: https://github.com/google/XNNPACK/pull/9803


ejparkqc:
@dsharlet I see that the 'cmake' folder is currently auto-generated by the build. Should the new Hexagon recipe also be auto-generated, or should we maintain separate source directories for scripts and recipes? I am currently mirroring the XNNPACK structure (scripts/ for shell scripts and cmake/ for cmake configuration called by shell scripts).

ejparkqc:
New PR will be opened if needed.

PR #10291, opened 2026-05-20
Fix Bazel Windows ARM64 MSVC config

snnn:
## Summary
The problem is:

Bazel’s Windows C++ toolchain path here uses arm64_windows. TensorFlow/XLA’s Windows toolchain definitions also key Windows ARM64 on arm64_windows. But XNNPACK currently keys its Windows ARM64 config on aarch64_windows.

## Reproduction
On Windows ARM64 with MSVC, upstream XNNPACK currently falls through to GCC std C flags for `--cpu=arm64_windows` and fails with:

```text
cl : Command line error D8021 : invalid numeric argument '/Wimplicit-fallthrough'
```

### DIFF
--- a/build_config/BUILD.bazel
+++ b/build_config/BUILD.bazel
@@ -142,9 +142,9 @@ config_setting(
 )
 
 config_setting(
-    name = "windows_aarch64",
+    name = "windows_arm64",
     values = {
-         "cpu": "aarch64_windows",
+        "cpu": "arm64_windows",
     },
 )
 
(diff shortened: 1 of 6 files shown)


build_config/BUILD.bazel
```diff
@@ -351,6 +359,14 @@ selects.config_setting_group(
     ],
 )
 
+selects.config_setting_group(
+    name = "arm_or_windows_aarch64",
+    match_any = [
+        ":arm",
+        ":windows_aarch64",
```
dsharlet: Can we add `:windows_aarch64` to the config_setting_group `:arm` and `:aarch64` above, instead of making a new one?

Then you might not need to change `build_params.bzl` (and anywhere else we might have missed).

dsharlet (approved):

PR #10307, opened 2026-05-22
Per-microkernel symbol-name and wall-time profiling

luhenry:
XNNPACK exposes per-operator wall time but not the C symbol of the microkernel that dispatched nor the wall time spent inside it. This adds both to the public profiling API.

**Public API (`include/xnnpack.h`):**
```
enum xnn_profile_info {
  [...]
  xnn_profile_info_microkernel_name    -- NUL-separated C symbol names,
                                          parallel to operator_name.
  xnn_profile_info_microkernel_timing  -- uint64_t[] of microkernel wall
                                          time per operator (microseconds).
}
```

**Opt-in build flag:**
| Flag | Description |
| ---- | ---- |
| `-DXNN_ENABLE_UKERNEL_TIMING=1` (default off) | Bracket each ukernel call in `xnn_compute_*` with `xnn_read_timer` and fold the elapsed time into a per-context accumulator. Changes the layout of gemm/igemm/dwconv/transpose/reduce/elementwise_binary/univector_{contiguous,strided} contexts; must be set consistently across all XNNPACK TUs. |

Symbol-name registry
--------------------
Each `XNN_INIT_*_UKERNEL` macro in `src/configs/*-config.c` calls `XNN_REGISTER_UKERNEL_NAME(ukernel)`, which stores `(fn_ptr, "#ukernel")` in an open-addressed hash table (`src/microkernel-name-registry.{h,c}`).

Operator-side resolution
------------------------
`xnn_get_runtime_profiling_info` walks each operator's ukernel slot based on `op->ukernel.type` (and `op->type` for the "default" type used by vbinary/vunary), looks up the function pointer in the registry, and returns the matching symbol.

Per-microkernel timing
----------------------
`src/xnnpack/ukernel-timing.h` defines four macros driving the instrumentation; the storage they touch is an implicit `__ukernel_timing_local` accumulator plus per-context elapsed_us slots.

The looped form:
```
  XNN_UKERNEL_TIMING_LOCAL;
  for (...) {
    XNN_UKERNEL_TIMING_RUN(fn, args, ...);
  }
  XNN_UKERNEL_TIMING_COMMIT(target);
```

Or the single-call form:
```
  XNN_UKERNEL_TIMING_RUN_AND_COMMIT(target, fn, args);
```

The local accumulator sums inside the loop; one `__atomic_fetch_add` per tile-task publishes it to the per-context slot. Workers race the slot under RELAXED ordering; the reader runs after the pthreadpool join, which is the acquire barrier.

Each compute context (gemm, igemm, dwconv, transpose, reduce, elementwise_binary, univector_{contiguous,strided}) gains a single ukernel_elapsed_us slot under `#if XNN_ENABLE_UKERNEL_TIMING`. The gemm and igemm union variants (dq/dq_qc2w/qp8) alias the same union slot and only one is live per op, so one accumulator covers all of them.

`xnn_read_timer` / `xnn_get_elapsed_time` move out of `src/runtime.c` into `src/xnnpack/timer.h` so `operator-run.c` can share the platform shims. The header deliberately does not pull in `log.h`, which would require `XNN_LOG_LEVEL` set on every TU that transitively includes `compute.h`.

`XNN_ENABLE_UKERNEL_TIMING` adds a u64 per context plus a `read_timer` pair per ukernel call; the static-lib delta is in the low-percent range.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -604,6 +604,7 @@ SET(XNNPACK_SRCS
   src/configs/xx-pad-config.c
   src/configs/x8-lut-config.c
   src/init.c
+  src/microkernel-name-registry.c
   src/params.c
   "${PROJECT_BINARY_DIR}/build_identifier.c")
 
--- a/include/xnnpack.h
+++ b/include/xnnpack.h
@@ -2414,6 +2414,13 @@ enum xnn_profile_info {
   xnn_profile_info_operator_name,
   /// Returns a uint64_t[] with the runtimes of all operators in the same order as xnn_profile_info_operator_name.
   xnn_profile_info_operator_timing,
+  /// Returns a char[] of null-separated C symbol names for the microkernel dispatched by each operator, in
+  /// the same order as xnn_profile_info_operator_name. Operators that don't go through the XNN_INIT_*_UKERNEL
+  /// macros contribute an empty string.
+  xnn_profile_info_microkernel_name,
+  /// Returns a uint64_t[] of accumulated microkernel wall time in microseconds, in the same order as
+  /// xnn_profile_info_microkernel_name. Zeroes when XNNPACK was built without -DXNN_ENABLE_UKERNEL_TIMING=1.
+  xnn_profile_info_microkernel_timing,
 };
 
 /// Return profile information for all operators.
--- a/src/configs/argmaxpool-config.c
+++ b/src/configs/argmaxpool-config.c
@@ -14,15 +14,17 @@
 #include "src/xnnpack/init-once.h"
 #include "src/xnnpack/log.h"
 #include "src/xnnpack/microfnptr.h"
+#include "src/xnnpack/microkernel-name-registry.h"
 
 static struct xnn_argmaxpool_config f32_argmaxpool_config = {0};
 
 XNN_INIT_ONCE_GUARD(f32_argmaxpool);
 
 // Macros to log the microkernel names if and when they are registered.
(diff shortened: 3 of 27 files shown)


luhenry:
You can find a sample of using the information at [1]. It's generally motivated to have a better understanding of which XNNPACK kernels are used for the given operators, and understand how much wall-time is spent in each of these kernels. Right now, this information is gathered by enabling logging at XNN_LOG_LEVEL=4, parsing the log output, and having some heuristic on matching microkernel name -> operator based on regexes and XNNPACK-internal knowledge [2].

Having this data readily available in XNNPACK Profiler makes it actually reliable, and makes information available that's hard to impossible to gather otherwise. This data is then key to understanding _what_ kernels need optimizations for a specific workloads.

If needed, this work can also be split into 2:
- capture the microkernel naming only; that's much lighter weight and gives us the exact operator -> microkernel mapping we are missing today and otherwise need to rely on heuristics for.
- capture the microkernel timing; that's much more impactful on performance and we can have an approximation already with the operators timing information. We are missing some of the head vs tail wall-time, as well as capturing the pthreadpool overhead with the operator timing.

[1]:
https://github.com/riseproject-dev/executorch/compare/riscv-testing-rvv...riseproject-dev:executorch:xnnpack-ukernel-profiler
[2]:
https://github.com/pytorch/executorch/pull/19707/changes/7eba60a0d4d28ba95b7c6d818f39d20e17a3bb04#diff-039550a86dd34eeb7e5adaa6a40dd57786c3ffc23bc6f3cae127143871a8add6

PR #10455, opened 2026-06-11
resize-bilinear: clamp source coordinate in legacy/align-corners mode (OOB read)

evilgensec:
### Problem

In the resize-bilinear indirection initializers (`src/indirection.c`, e.g. `xnn_indirection_init_resize_bilinear2d_hwc_f32`), the `tensorflow_legacy || align_corners` branch computes the source coordinate and casts it to the top index **without clamping**:

```c
const float input_y = (float) (int32_t) output_y * height_scale;
assert(input_y < (float) input_height);                 // compiled out in release
const uint32_t input_y_top = (uint32_t) (int32_t) input_y;   // NOT clamped to input_y_max
const uint32_t input_y_bottom = math_min_u32(input_y_top + 1, input_y_max);  // only this is clamped
```

The code relies on `input_y` staying `< input_height`, but `height_scale` is float32 and for a large `output_height` the rounded product `(output_height - 1) * height_scale` can round up to exactly `input_height`. `input_y_top` is then `input_height`, and the indirection buffer references input row `input_height` — one full row past the input tensor:

```c
indirection_buffer[0] = input + (input_y_top * input_width + input_x_left) * input_pixel_stride;
```

The `ibilinear` ukernel dereferences it, reading out of bounds (CWE-125). The default (half-pixel) branch does not have this problem because it clamps the coordinate first (`input_y = math_min_f32(math_max_f32(input_y, 0.0f), (float) input_y_max);`). The same unclamped pattern is in all five resize initializers (f16/f32/qs8/qu8 NHWC and the CHW variant), for both the height and width coordinate.

Reproduced deterministically: an exact-size input tensor (`2x2x8`) placed at the end of a mapped page with an unmapped guard page immediately after, resized to `output_height = 13103685, output_width = 2` in `XNN_FLAG_TENSORFLOW_LEGACY_MODE`. On the last output row, `input_y` rounds to `2.0` so `input_y_top == input_height == 2`, and the over-read of input row 2 faults on the guard page (`EXC_BAD_ACCESS`). `output_height < 2**24`, so the shape is accepted by `xnn_create_resize_bilinear2d_nhwc`. A normal heap buffer hides the fault but silently reads adjacent heap into the interpolated output (information disclosure).

### Fix

Clamp `input_y` and `input_x` with `math_min_f32`/`math_max_f32` in the legacy/align-corners branch, exactly as the default branch already does, in all five resize indirection initializers. For valid shapes the coordinate is already `< input_height`, so the clamp is a no-op and the output is unchanged.

### Testing

- The guard-page reproducer above faults before the fix and runs cleanly after it.
- A normal legacy-mode resize (e.g. `4x4 -> 8x8`) still runs and produces the same output (the clamp is a no-op for in-range coordinates; corner values are preserved).

### DIFF
--- a/src/indirection.c
+++ b/src/indirection.c
@@ -496,17 +496,20 @@ void xnn_indirection_init_resize_bilinear2d_hwc_f16(
   const uint32_t input_x_max = (uint32_t) input_width - 1;
   if (tensorflow_legacy || align_corners) {
     for (size_t output_y = output_y_start; output_y < output_y_end; output_y++) {
-      const float input_y = (float) (int32_t) output_y * height_scale;
-      assert(input_y >= 0.0f);
-      assert(input_y < (float) input_height);
+      float input_y = (float) (int32_t) output_y * height_scale;
+      // Clamp the source coordinate as the half-pixel branch does. The legacy
+      // and align-corners formulas rely on input_y staying < input_height, but
+      // height_scale is float32 and for large output_height the rounded product
+      // can reach input_height, putting input_y_top one full row past the input
+      // (out-of-bounds read in the bilinear ukernel).
+      input_y = math_min_f32(math_max_f32(input_y, 0.0f), (float) input_y_max);
 
       const uint32_t input_y_top = (uint32_t) (int32_t) input_y;
       const uint32_t input_y_bottom = math_min_u32(input_y_top + 1, input_y_max);
       const float alpha_y = input_y - (float) input_y_top;
       for (size_t output_x = 0; output_x < output_width; output_x++) {
-        const float input_x = (float) (int32_t) output_x * width_scale;
-        assert(input_x >= 0.0f);
-        assert(input_x < (float) input_width);
+        float input_x = (float) (int32_t) output_x * width_scale;
+        input_x = math_min_f32(math_max_f32(input_x, 0.0f), (float) input_x_max);
 
         const uint32_t input_x_left = (uint32_t) (int32_t) input_x;
         const uint32_t input_x_right = math_min_u32(input_x_left + 1, input_x_max);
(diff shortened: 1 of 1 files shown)


Aelphy (commented):
Sorry for late review. Please sync to resolve a merge conflict.

evilgensec:
Reopening — closing this was premature on my part. This PR (opened 2026-06-11) proposes the coordinate clamp that later landed in master via 14663846eb (2026-06-23).

evilgensec:
Status update so this does not stall on the sync request:

The same coordinate clamp already landed on master in 14663846eb (2026-06-23), which also added regression tests in test/indirection.cc. Rebasing this branch onto current master now produces an empty diff, so there is no code left to merge on top of that commit.

Background: I reported this out-of-bounds read to the Google OSS VRP on 2026-06-11 (issue 522593840) and opened this PR the same day as the proposed fix. The report was initially closed as out of scope because the OSS VRP requires an OT1 memory-corruption finding to have a committed and merged patch, which did not exist at that point. With 14663846eb now merged, that condition is satisfied.

To reopen the VRP case, proof of issue initiation is required: evidence that the merged fix traces back to the original report. Would you confirm on this thread that 14663846eb addresses the issue reported here (this PR and report 522593840, both dated 2026-06-11)? A short acknowledgment, or merging or referencing this PR from the fix, is enough to meet that requirement.

Aelphy:
Both changes seem to solve same problem.

evilgensec:
Thanks for confirming both changes address the same problem. That coordinate clamp already landed on master in 14663846eb, so rebasing this branch produces an empty diff and there is nothing left to merge here. Happy to close this PR since the fix is in, unless you would prefer to keep it linked for reference.

Aelphy:
Please close.

PR #10473, opened 2026-06-15
bound split axis to input rank in split_dims reshape

aizu-m:
Tracing the static reshape helpers in copy.c, split_dims is the one that never checks its split axis against the live input rank:

    split_dims, rank-2 input {4,4}, axis=3:
      input elements = 16
      output buffer  = 0   (axis past the rank pulls a stale dim into the count)
      copy writes 16 elements into a 0-element buffer -> out-of-bounds write

axis comes from the node and is only bounded against XNN_MAX_TENSOR_DIMS at define time, never against the input tensor. The runtime rank can be smaller, or change later through xnn_reshape_external_value. With axis at or past the rank, resize_split_dims_output_tensor reads input_shape->dim[axis] from beyond the live rank and emits an output element count that no longer matches the input. reshape_copy_operator then copies xnn_shape_multiply_all_dims(input) elements into that undersized output.

fuse_dims already rejects this (input_shape->num_dims < first_dim + num_dims), and even-split and concatenate both reject an axis past the rank. split_dims was the gap, so it gets the same guard. Added a regression test that fails on the old tree and passes with the check.

### DIFF
--- a/src/subgraph/copy.c
+++ b/src/subgraph/copy.c
@@ -263,6 +263,16 @@ static enum xnn_status resize_split_dims_output_tensor(
   const struct xnn_shape* input_shape = &input->shape;
   struct xnn_shape* output_shape = &output->shape;
 
+  if (axis >= input_shape->num_dims) {
+    xnn_log_error("failed to split dims in %s operator with input ID #%" PRIu32
+                  " and output ID #%" PRIu32
+                  ": split axis, %zu, is out of range for an input with %zu "
+                  "dimensions",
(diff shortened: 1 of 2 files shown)


Aelphy:
Please fix merge conflicts.

aizu-m:
Rebased onto master, conflicts resolved. The only clash was in split-fuse.cc where the new large_axis_rejected tests landed at the same spot as this one; kept both, mine now sits after the #ifndef XNNPACK_USE_YNNPACK block. copy.c merged cleanly.

PR #10521, opened 2026-06-17
Add f32 x16 gemm microkernels for WASM Relaxed SIMD

yolanda15:
Add the f32 x16 gemm microkernels that is similar to avx2 default config for wasm relaxed simd.
The 5x16 wasm relaxed simd fma loadsplat kernel is ~20% slower than the corresponding 4x8 microkernel, but with the unrolled version, the gap can be reduced to within 5%.

With revec enabled, the unrolled version is also 18 - 30% faster than the default 4x8 microkernels.


fbarchard (approved):
Looks good.  I wonder if we should be adding 'revec' to the naming to clarify kernel choices if find a way to detect revec is enabled.. or 512 bit in the future
Thanks for the change!

PR #10523, opened 2026-06-17
Add Hexagon V75 CMake build and adapt subgraph to newer slinky API

ejparkqc:
- These are additions to the branch 'test_890235712' (https://github.com/google/XNNPACK/pull/9803)
- Add a Hexagon V75 cross-compile path for YNNPACK: toolchain file, build driver, pthreadpool fetch, and a hexagon-only slinky patch for `int` -->  `slinky::expr` overload disambiguation. Switch gtest discovery to PRE_TEST so it doesn't run binaries on hexagon-sim at build time.
 - Update `subgraph/` for slinky API changes: `mutable_dim` --> `dim`, drop `in_bounds{}` wrapper, fix `enqueue` signature, and replace removed `store_outputs_innermost()` with per-output `store_at` calls.


ejparkqc:
@dsharlet @fbarchard Created a PR again. Works for V75.

PR #10528, opened 2026-06-18
ci: add cmake-windows-arm64-native job using windows-11-arm runner

chinazhangchao:
Add a native arm64-on-arm64 CI job and the matching build script.

The existing cmake-windows-arm64 job runs an x64 -> arm64 cross from windows-2022-8core and therefore cannot exercise ctest on the produced binaries. The new cmake-windows-arm64-native job runs on the windows-11-arm runner (publicly available since April 2025) with the native arm64 v143 toolset that ships in the image, and runs ctest when the caller workflow sets inputs.run-tests.

Changes:

scripts/build-windows-arm64-native.cmd (new): mirrors the cl.exe feature set of scripts/build-windows-arm64.cmd, but does not pass a CMAKE_TOOLCHAIN_FILE so CMAKE_CROSSCOMPILING stays FALSE and ctest can launch the binaries.
.github/workflows/build.yml: new cmake-windows-arm64-native job inserted next to the existing cmake-windows-arm64 cross job. Same style as cmake-windows-x64 (vswhere + vcvarsall + ninja + ctest).
The new job is additive; the existing cross job is unchanged so users building on x64 hosts keep working.

Tested locally on Snapdragon X Plus (X1P-64-100, Surface Laptop 7) by running scripts/build-windows-arm64-native.cmd in a fresh repo clone under VS 2022 17.14 Developer Prompt. Build completes; ctest run launches without CMAKE_CROSSCOMPILING errors.


.github/workflows/build.yml
dsharlet, line 188: Should we delete this one now? It seems redundant, and we struggle in general with the number of presubmits we have.

dsharlet (approved):

.github/workflows/build.yml
```diff
@@ -184,34 +184,26 @@ jobs:
           key: ccache-riscv64-${{ runner.os }}-${{ github.job }}-${{ github.run_id }}-${{ github.run_attempt }}
       - name: Print ccache stats
         run: ccache -s
-
-  cmake-windows-arm64:
-    runs-on: windows-2022-8core
+  cmake-windows-arm64-native:
```
dsharlet: We have some internal infrastructure that uses this name, just leaving it alone (not changing it) would be easiest.

.github/workflows/build.yml
```diff
@@ -184,34 +184,26 @@ jobs:
           key: ccache-riscv64-${{ runner.os }}-${{ github.job }}-${{ github.run_id }}-${{ github.run_attempt }}
       - name: Print ccache stats
         run: ccache -s
-
```
dsharlet, line 187: I think we should keep this newline. I don't know if YAML requires it or not, but it's also inconsistent with the rest.

PR #10531, opened 2026-06-18
Hexagon ISA version detection

ejparkqc:
- Addition on Cmake build PR (https://github.com/google/XNNPACK/pull/9803 and https://github.com/google/XNNPACK/pull/10523)
- Runtime detection via qurt_sysenv_get_arch_version, exposed as arch_flag::hvx_v79 / hvx_v81 (cumulative; V81 implies V79).
- Per-file -mv79 / -mv81 compile flags in kernels/arch_copts.cmake so individual kernels can opt into newer ISA while the build floor stays at V75 (YNN_ENABLE_HVX_V79 / _V81 default to YNN_ENABLE_HVX).
- Fix YNN_ARCH_HVX -> YNN_ARCH_HEXAGON_HVX in base/CMakeLists.txt so the existing HVX kernel .inc registrations actually compile in.


ejparkqc:
@dsharlet @fbarchard 

I still need to add test code for v79 kernel to evaluate further.

ynnpack/kernels/dot/consistent_arithmetic_test.cc
```diff
@@ -139,8 +139,8 @@ void TestMatMul(AT, BT, CT, size_t k) {
       int finite = 0;
       for (size_t i = 0; i < m; ++i) {
         for (size_t j = 0; j < n; ++j) {
-          bool c_finite = std::isfinite(c(i, j));
-          bool kernel_c_finite = std::isfinite(kernel_c(i, j));
+          bool c_finite = std::isfinite(static_cast<float>(c(i, j)));
```
dsharlet: This should be fixed at head, can you please rebase?

ejparkqc: Rebasing the entire branch onto master seems to bring in all changes between test_890235712 and master (few hundreds of commits showed up in this PR, mainly not mine), so I only brought this one file from master, I brought over just that file instead. Do you think I should still rebase the entire branch onto master?

PR #10538, opened 2026-06-19
validate axis in ynn_define_broadcast against input rank

aizu-m:
`ynn_define_broadcast` converts each caller axis with `axis_to_slinky_dim` and sets the bit in `axes_set`, a `std::bitset<YNN_MAX_TENSOR_RANK + 2>`:

    axis_to_slinky_dim(rank=3, axis=-100) -> 99   // axes_set holds 10 bits
    axes_set[99] = true;                          // operator[] out of range

`axis_to_slinky_dim` returns `-(axis + 1)` for a negative axis, so any axis below `-rank` (`-100`, or `INT32_MIN` -> `2147483647`) lands past the end. `std::bitset::operator[]` does not range check; on libstdc++ `bitset<10>` that is `_M_w[99 / 64] = _M_w[1]`, one word past the single-word stack array, an out-of-bounds write.

Spotted while auditing which axis-taking define functions validate their axes after the recent gather index fix. `broadcast_like`, `slice_like` and `reduce` already bound the converted dim against the rank, and `concatenate`, `even_split`, `gather`, `stack` and the rest call `validate_axis` up front. broadcast was the one that skipped it, so the same guard goes in.

Regression test in `errors.cc` rejects an out-of-range axis (fails before the guard, passes after).

### DIFF
--- a/ynnpack/subgraph/test/errors.cc
+++ b/ynnpack/subgraph/test/errors.cc
@@ -49,21 +49,4 @@ TEST(Errors, bad_dot_shape) {
   ASSERT_EQ(runtime.ReshapeRuntime().Status(), ynn_status_error);
 }
 
-TEST(Errors, broadcast_axis_out_of_bounds) {
-  const uint32_t in_id = 0;
-  SubgraphBuilder subgraph(1);
-  subgraph.AddInput(ynn_type_fp32, 3, in_id);
-
-  // An axis far outside [-rank, rank) maps to a slinky dim past the input's
-  // dimensions. Such a dimension is an implicit broadcast, so broadcasting it
-  // is a no-op: it must be skipped rather than indexing the axes_set bitset out
-  // of range.
-  const int32_t axes[] = {-100};
-  uint32_t output_id = YNN_INVALID_VALUE_ID;
-  EXPECT_EQ(ynn_define_broadcast(subgraph.GetSubgraph(), /*num_axes=*/1, axes,
-                                 in_id, &output_id, /*flags=*/0),
-            ynn_status_success);
-  EXPECT_EQ(output_id, in_id);
-}
-
 }  // namespace ynn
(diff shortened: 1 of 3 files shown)


dsharlet:
Thanks for the fix, but in YNNPACK, internally we consider tensors to have infinite dimensions, where the extra dimensions are broadcasts. A broadcast of a broadcast dimension is a no-op, so we should just skip assigning the bitset, and move on without producing an error.

dsharlet:
See e.g. broadcast_like: https://github.com/google/XNNPACK/blob/4bfa3bb8391555b78e4b299297f6ccc29e47163c/ynnpack/subgraph/broadcast_like.cc#L44-L46

aizu-m:
Makes sense, the infinite-dimensions framing wasn't on my radar. Switched it to match broadcast_like: convert the axis, and only set the bit when it lands inside the input's rank, otherwise skip it as a no-op rather than erroring. With axis -100 on a rank-3 input the define now returns success and aliases the input, and the out-of-range axes_set index is gone. Updated the errors.cc test to assert that.

ynnpack/subgraph/test/errors.cc
```diff
@@ -49,4 +49,21 @@ TEST(Errors, bad_dot_shape) {
   ASSERT_EQ(runtime.ReshapeRuntime().Status(), ynn_status_error);
 }
 
+TEST(Errors, broadcast_axis_out_of_bounds) {
```
dsharlet: This test isn't an error any more. I think if we are going to cover this behavior, we should modify the existing broadcast test to include out of bounds broadcasts in its behavior.

aizu-m: Agreed, it returns success now so it doesn't belong in errors.cc. Dropped it there and put a behavioural check in static_broadcast.cc instead: broadcast with axis -100 on a rank-3 input, run it, then confirm the output matches the input since the out-of-range axis is skipped as a no-op. Had to give it its own suite name (BroadcastAxes) because the Broadcast suite in that file is a TEST_P and gtest won't let a plain TEST reuse the name. That suite only drives ynn_define_static_broadcast, so there wasn't an existing axes-based broadcast test to fold into. Happy to move it if you had a different file in mind.

dsharlet:
Thanks for the PR.

This has conflicts, and I think it was done in #10564, please re-open if not and the conflicts are resolved.

PR #10543, opened 2026-06-20
Fix benchmark minmax parameter order

fallintoplace:
## Summary
- Fix reversed no-clamp min/max parameters in the f16-f32acc IGEMM benchmark.
- Fix the same reversed half-precision min/max constants in the f16 HWC-to-CHW convolution benchmark.
- Add a lightweight CMake source check for obvious reversed `init_params` min/max argument pairs.

## Testing
- `python3 tools/check-minmax-init-params.py`
- `ctest --test-dir build/minmax-check -R ^minmax-init-params-test$ --output-on-failure`
- `cmake --build build/minmax-check --target f16-f32acc-igemm-bench f16-conv-hwc2chw-bench --parallel 8`

### DIFF
--- a/bench/f16-conv-hwc2chw.cc
+++ b/bench/f16-conv-hwc2chw.cc
@@ -87,7 +87,8 @@ static void f16_conv_hwc2chw(
   xnnpack::Buffer<xnn_float16> output(output_elements * num_buffers);
 
   xnn_f16_minmax_params params;
-  init_params(&params, 0x7C00 /* inf */, 0xFC00 /* -inf */);
+  init_params(&params, static_cast<xnn_float16>(-INFINITY),
+              static_cast<xnn_float16>(INFINITY));
 
   size_t buffer_index = 0;
   for (auto _ : state) {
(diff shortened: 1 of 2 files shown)


bench/f16-conv-hwc2chw.cc
```diff
@@ -87,7 +87,7 @@ static void f16_conv_hwc2chw(
   xnnpack::Buffer<xnn_float16> output(output_elements * num_buffers);
 
   xnn_f16_minmax_params params;
-  init_params(&params, 0x7C00 /* inf */, 0xFC00 /* -inf */);
+  init_params(&params, 0xFC00 /* -inf */, 0x7C00 /* inf */);
```
dsharlet: This is wrong for two reasons: it swapped +/-inf, but it also is passing an integer to an xnn_float16, so this is converting the integer to a float, and then converting that float to a float16. It should be 
```
  init_params(&params, static_cast<xnn_float16>(-INFINITY),
              static_cast<xnn_float16>(INFINITY));
```

Like the other file in this PR.

tools/check-minmax-init-params.py
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env python3
```
dsharlet: I don't think we need this test. The affected code was just benchmarks, it was unlikely to affect performance anyways. If it affected correctness, we would have known about it earlier.

PR #10544, opened 2026-06-21
ignore fuse of last dimension in ynn_define_fuse_dims

aizu-m:
ASan, YNNPACK built with `-fsanitize=address`:

    ERROR: AddressSanitizer: heap-buffer-overflow, READ of size 8
      #0 ynn_define_fuse_dims copy.cc:542
    0x... located 0 bytes after 24-byte region
      allocated by vector<slinky::expr>::__assign ... copy.cc:539

`ynn_define_fuse_dims` fuses each axis with the dimension after it: it sets `op.axes[axis_to_slinky_dim(rank, axes[i] + 1)]`, then propagates the shape with `output.extents[i] = extent(i) * extent(i + 1)` and `erase(begin() + i + 1)`.

For the last dimension (`axes[i] == rank-1`, or `-1`) the successor is dimension `rank`, which is out of bounds, so the loop wrote `output.extents[rank]` and erased one past the end of the rank-sized extents vector.

The successor of the last dimension is an implicit broadcast dimension, so fusing with it is a no-op. Skip any axis whose successor is out of bounds instead of touching the extents. Out of bounds axes are broadcast dimensions too, so they are now ignored rather than rejected.

### DIFF
--- a/ynnpack/subgraph/copy.cc
+++ b/ynnpack/subgraph/copy.cc
@@ -528,11 +528,16 @@ ynn_status ynn_define_fuse_dims(ynn_subgraph_t subgraph, size_t num_axes,
 
   ynn_node::fuse_dims op;
   for (size_t i = 0; i < num_axes; ++i) {
-    YNN_RETURN_IF_ERROR(
-        validate_axis("fuse_dims", "input", input.rank(), axes[i]));
     // Since we are reversing the axes, the first dimension to fuse is actually
     // the next dimension.
-    op.axes[axis_to_slinky_dim(input.rank(), axes[i] + 1)] = true;
+    const int dim = axis_to_slinky_dim(input.rank(), axes[i] + 1);
+    // Fusing the last (or an out of bounds) dimension fuses it with an implicit
+    // broadcast dimension, which is a no-op, so ignore it. This also keeps the
+    // shape propagation below from indexing past the end of the extents.
+    if (dim + 1 >= input.rank()) {
+      continue;
+    }
+    op.axes[dim] = true;
   }
 
   // Propagate shape.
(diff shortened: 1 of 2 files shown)


dsharlet:
Most of these and the existing checks in split/fuse should probably be removed, and replaced with code that just ignores them. Splitting or fusing a broadcast dimension (which is what out of bounds dimensions are) is a no-op and should be ignored, not an error.

aizu-m:
Makes sense. Reworked fuse_dims to skip an axis whose successor is out of bounds rather than error, since fusing with a broadcast dimension is a no-op. Dropped the validate_axis call too, so out of bounds axes are ignored rather than rejected. The test now checks the fuse passes through unchanged for the last axis, the negative index, and an out of range axis.

The sibling split/fuse defines still validate up front. Happy to give them the same pass if you want it, here or as a follow-up.

dsharlet:
Thanks for this PR, and all of your other PRs, it has been great to have your contributions!

Sorry to collide in this case but I decided to just fix a few of these myself: https://github.com/google/XNNPACK/pull/10564

dsharlet:
I'm going to close this one, please feel free to re-open if I missed something after resolving the conflicts.

aizu-m:
No worries, #10564 covers it. Pulled the merged diff and traced it: fuse_dims now skips the axis when axis_to_slinky_dim lands at or past the rank, so for the last dim (slot rank) op.axes never gets set and the propagation loop never writes output.extents[rank] or erases past the end. The ASan overflow is gone. Good to see fuse_dim and broadcast got the same pass in one go, that was the follow-up I had in mind. Nothing missed, happy to leave this closed.

PR #10545, opened 2026-06-21
reject input rank mismatch in transpose reshape

aizu-m:
Was poking at transpose with mismatched runtime shapes. A rank-3 transpose fed a rank-2 external input slips straight through reshape instead of being rejected:

    Transpose.reshape_rejects_input_rank_mismatch
      subgraph.Status()              Which is: 0   (xnn_status_success)
      xnn_status_invalid_parameter   Which is: 2

reshape_transpose_operator takes its working rank from the permutation length, which is fixed at define time, and only ties it to the live input rank with an assert. xnn_reshape_external_value lets an external input be reshaped to any rank at run time, so under NDEBUG the assert is gone and the mismatched input is accepted. The op then transposes with a stale dimension count: it reads input->shape.dim[perm[i]] for slots past the live rank (stale values left over from the larger declared shape) and ends up walking the input buffer past what it actually holds.

The permutation length and the input rank must agree. Swapped the assert for a real check, same shape as the softmax and fully-connected runtime rank guards. The regression test reshapes a rank-3 transpose input down to rank 2 and expects invalid_parameter.

### DIFF
--- a/test/subgraph/static-transpose.cc
+++ b/test/subgraph/static-transpose.cc
@@ -85,4 +85,26 @@ INSTANTIATE_TEST_SUITE_P(Transpose, TransposeQU8, rank_params);
 INSTANTIATE_TEST_SUITE_P(Transpose, TransposeF16, rank_params);
 INSTANTIATE_TEST_SUITE_P(Transpose, TransposeF32, rank_params);
 
+TEST(Transpose, reshape_rejects_input_rank_mismatch) {
+  ASSERT_EQ(xnn_status_success, xnn_initialize(nullptr /* allocator */));
+
+  // The permutation fixes the operator rank at define time. An external input
+  // may be reshaped to a different rank at runtime; the reshape must reject the
+  // mismatch instead of transposing with a stale dimension count and reading
+  // past the input buffer.
+  const std::vector<size_t> perm = {2, 0, 1};
+  SubgraphTester subgraph(2);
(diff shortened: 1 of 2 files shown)


test/subgraph/static-transpose.cc
```diff
@@ -85,4 +85,26 @@ INSTANTIATE_TEST_SUITE_P(Transpose, TransposeQU8, rank_params);
 INSTANTIATE_TEST_SUITE_P(Transpose, TransposeF16, rank_params);
 INSTANTIATE_TEST_SUITE_P(Transpose, TransposeF32, rank_params);
 
+TEST(Transpose, reshape_rejects_input_rank_mismatch) {
```
dsharlet, line 89: Add `#ifndef XNNPACK_USE_YNNPACK` around this test. In YNNPACK, transpose can add or remove dimensions, so this check is overly conservative.

dsharlet (commented):
Thank you for (another) fix!

test/subgraph/static-transpose.cc
aizu-m: Done, wrapped it in #ifndef XNNPACK_USE_YNNPACK with an #else note, same shape as the softmax test. The reshape check itself lives in src/subgraph/static-transpose.c, which YNNPACK doesn't route through (it has its own static_transpose), so only the test needed the guard.

PR #10576, opened 2026-06-24
Apply XNN_NO_SANITIZE_FUNCTION to remaining type-erased call sites

wangw-1991:
`XNN_NO_SANITIZE_FUNCTION` was introduced to suppress `-fsanitize=function` false positives where XNNPACK calls microkernels/helpers through type-erased function pointers, but it was not applied everywhere it should have been.

This PR fills in the gaps across `src/operator-run.c` and `src/operators/*.c`, annotating the remaining functions that directly invoke such pointers. The macro is a no-op unless built with Clang `-fsanitize=function`, so there is no effect on code generation or runtime behavior.

Chromium issue: https://issues.chromium.org/issues/520663347

### DIFF
--- a/src/operator-run.c
+++ b/src/operator-run.c
@@ -103,9 +103,9 @@ XNN_NO_SANITIZE_FUNCTION void xnn_compute_transposec_5d(struct transpose_context
   context->const_size_ukernel(x, y, ld_input, ld_output, tile_l, tile_m);
 }
 
-void xnn_compute_transposec_6d(struct transpose_context* restrict context,
-                               size_t i, size_t j, size_t k, size_t l, size_t m,
-                               size_t n, size_t tile_m, size_t tile_n) {
+XNN_NO_SANITIZE_FUNCTION void xnn_compute_transposec_6d(
+    struct transpose_context* restrict context, size_t i, size_t j, size_t k,
+    size_t l, size_t m, size_t n, size_t tile_m, size_t tile_n) {
   const size_t ld_input = context->input_stride[5];
   const size_t ld_output = context->output_stride[4];
   const void* x =
@@ -124,9 +124,9 @@ void xnn_compute_transposec_6d(struct transpose_context* restrict context,
   context->const_size_ukernel(x, y, ld_input, ld_output, tile_m, tile_n);
 }
 
-void xnn_compute_transposev_2d(struct transpose_context* restrict context,
-                               size_t i, size_t j, size_t tile_i,
-                               size_t tile_j) {
+XNN_NO_SANITIZE_FUNCTION void xnn_compute_transposev_2d(
+    struct transpose_context* restrict context, size_t i, size_t j,
+    size_t tile_i, size_t tile_j) {
   const size_t element_size = context->output_stride[1];
   const size_t ld_input = context->input_stride[1];
   const size_t ld_output = context->output_stride[0];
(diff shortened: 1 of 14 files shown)


wangw-1991:
Hi, @alexander-shaposhnikov. Can you help review this? Thanks.

wangw-1991:
@alexander-shaposhnikov Friendly ping. Or @dsharlet Can you help review this?

alexander-shaposhnikov:
iirc , one could apply the attribute to callers rather than callees, so one doesn't have to apply the attribute throughout the whole codebase. E.g. it's redundant to annotate leaf functions etc. 
Do you happen to have a link to new reports ? I can easily imagine that the codebase could have evolved or xnnpack has more uncovered code paths / smth was missed.

wangw-1991:
Thanks. I think -fsanitize=function checks the callee's prologue type metadata at the call site. Marking a function no_sanitize("function") makes Clang omit that prologue signature, so the (instrumented) call site reads "no type info" and skips the report. That's why the attribute has to sit on the callee — it's the only place that removes the metadata the check trips over. Annotating callers wouldn't help even if we could reach them.

What do you mean by the "new report"? I only have the link to the Chromium issue that detects this issue by fuzzer: https://issues.chromium.org/issues/520663347

alexander-shaposhnikov:
Let me clarify a few things.
Consider the following example:
```
#include <cstdio>

#ifdef CALLEE
__attribute__((no_sanitize("function")))
#endif
void greet(void*) {
    std::printf("Hello, world!\n");
}

#ifdef CALLER
__attribute__((no_sanitize("function")))
#endif
int main() {
    // --- Correct call: type matches, no error ---
    void (*fp)(void*) = greet;
    fp(nullptr);

    // and call through it. The call itself is UB and gets caught.
    void (*bad)(int*) = reinterpret_cast<void (*)(int*)>(reinterpret_cast<void*>(greet));
    bad(nullptr);   // <-- runtime error reported here

    return 0;
}
```
and try to compile 
```clang -DCALLER -fsanitize=function ...``` (https://godbolt.org/z/h71zsjx56) 
```clang -DCALLEE -fsanitize=function ...``` (https://godbolt.org/z/93ccsnzvq)

In particular one needs to pay attention to (for -DCALLEE)
```
        .type   greet(void*),@function
        .long   3238382334                      # 0xc105cafe
        .long   1413295203                      # 0x543d2c63
greet(void*):                             # @greet(void*)
```
But the mechanisms by which they work are different - the former removes the check entirely while the latter disables the signature.

Regarding https://issues.chromium.org/issues/520663347 - unfortunately i don't seem to have access, 
but just wanted to double-check if it uses the up-to-date version of XNNPACK or not.

wangw-1991:
Thanks for the detailed explanation. If I understand correctly, this attribute does not cascade downwards. If `main` calls another normal function `A()`, and `A()` executes the exact same `bad_ptr(nullptr)` call, the runtime error will still trigger because `A()'s` internal compiler checks remain active.  This means you must decorate every single function that could potentially initiate the mismatched indirect call, right? 

I have CC'd you in the issue, you should be able to access it using your Google account now.

alexander-shaposhnikov (approved):
I've checked a few examples (xnn_compute_transposev_4d, xnn_compute_transposec_6d) - the changes make sense to me.
I'd  probably double check if there are any redundant annotations, 
but otherwise looks good to me.

wangw-1991:
@alexander-shaposhnikov thanks, looking forward to your feedback. Please help merge this if no other issue.

PR #10588, opened 2026-06-25
security: zero sensitive weight buffers before release in xnn_destroy_operator() and FP16/BF16 conversion functions

vijaypatilraje-hue:
## Summary

This patch adds `explicit_bzero()` calls before every `xnn_release_memory()` /
`xnn_release_simd_memory()` call on buffers that hold caller-supplied model
weight data. The goal is to reduce the window during which those values remain
readable in freed heap memory.

## Motivation

Standard allocators (`free()`, `posix_memalign`) do not scrub freed memory.
On Linux, a freed heap page stays mapped in the process address space until the
OS reclaims it. In multi-model inference services or multi-tenant deployments
where several models share a process, weight data from a destroyed operator can
remain accessible in freed memory and be read by a later allocation landing on
the same freed page.

`memset()` is not used here because compilers are permitted to elide dead-store
`memset()` calls under optimization. `explicit_bzero()` is the POSIX primitive
specifically designed to survive optimization for this purpose — the same
approach used by OpenSSL, libsodium, and the Linux kernel for sensitive data.

## Changes

**`src/operator-utils.c` — `xnn_destroy_operator()`**
Zero `op->packed_weights.pointer` (the repacked weight matrix, held for the
full operator lifetime) before `xnn_release_simd_memory()`.

**`src/operators/fully-connected-nc.c`**
- Zero transient `fp32_kernel_buffer` and `fp32_bias_buffer` in
  `xnn_create_fully_connected_nc_f32_f16` and
  `xnn_create_fully_connected_nc_pf32_f16` before `xnn_release_memory()`.
- Zero `bf16_scale_buffer` at 4 callsites before `xnn_release_memory()`.

**`src/operators/convolution-nhwc.c`**
Zero transient `fp32_kernel_buffer` and `fp32_bias_buffer` in
`xnn_create_convolution2d_nhwc_f32_f16` and
`xnn_create_convolution2d_nhwc_pf32_f16` before release.

**`src/operators/convolution-nchw.c`**
Zero transient `fp32_kernel_buffer` and `fp32_bias_buffer` in
`xnn_create_convolution2d_nchw_f32_f16` before release.

**`src/operators/deconvolution-nhwc.c`**
Zero transient `fp32_kernel_buffer` and `fp32_bias_buffer` in
`xnn_create_deconvolution2d_nhwc_f32_f16` before release.

## Compatibility

`explicit_bzero()` is available on:
- glibc 2.25+ (all modern Linux distributions)
- Bionic (Android NDK r23+)
- Apple platforms (macOS 10.12+, iOS 10+)

The `_GNU_SOURCE` guard already present in several XNNPACK source files
(e.g. `src/memory.c` lines 24–25) is added to each patched file to expose
`explicit_bzero` from `<string.h>` on Linux. A follow-up can address
`SecureZeroMemory()` for `XNN_PLATFORM_WINDOWS` paths if desired.

## Impact on correctness and performance

This patch adds no new code paths and changes no return values or API behavior.
The `explicit_bzero()` calls add a one-time linear write over each buffer
immediately before it is freed — this occurs only at operator teardown and
creation, not during inference.

5 files changed, 71 insertions(+)

### DIFF
--- a/src/operator-utils.c
+++ b/src/operator-utils.c
@@ -3,6 +3,9 @@
 // This source code is licensed under the BSD-style license found in the
 // LICENSE file in the root directory of this source tree.
 
+#if !defined(_GNU_SOURCE)
+#define _GNU_SOURCE
+#endif
 #include "src/xnnpack/operator-utils.h"
 
 #include <assert.h>
@@ -194,6 +197,16 @@ enum xnn_status xnn_destroy_operator(xnn_operator_t op)
     xnn_release_memory(op->convolution_op);
   }
   if (op->weights_cache == NULL) {
+    // CWE-226: Zero packed weight buffer before release so model weights do not
+    // remain recoverable from freed heap memory. The buffer size is
+    // group_output_channels * weights_stride, matching the allocation in
+    // xnn_get_pointer_to_write_weights().
+    if (op->packed_weights.pointer != NULL && op->convolution_op != NULL &&
+        op->weights_stride > 0) {
+      explicit_bzero(op->packed_weights.pointer,
+                     op->convolution_op->group_output_channels *
+                         op->weights_stride);
+    }
     xnn_release_simd_memory(op->packed_weights.pointer);
   }
   xnn_release_simd_memory(op->zero_buffer);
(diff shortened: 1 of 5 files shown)


dsharlet:
I think if you want this behavior, it can be achieved by customzing the behavior of `xnn_allocator`: https://github.com/google/XNNPACK/blob/6ca28886403bc52351a1398055ba01a1139884a9/include/xnnpack.h#L168-L173

I think this change is not something we should do as a built-in behavior of XNNPACK

PR #9960, opened 2026-04-13
Add qd8-qb4w-bf16 scalar kernels

GregoryComer:
Add scalar qd8-qb4w-bf16 kernels. This is mainly just an output conversion, so kernels are very similar to the existing qd8-qb4w-f16/f32. Performance is with ~5% of scalar fp16. The scalar rounding logic is strictly correct for NaNs/INFs and has correct round to even support, so it's a little expensive. The M4 Max f16 kernels beat bf16 slightly because it uses native scalar fp16 arithmetic in the default build config.

### M4 Max
  | Kernel | BL | 128/128/1024 | 128/4096/1024 | 128/11008/4096 |
  |---|---|---|---|---|
  | scalar 1x4 bf16 | 32 | 39.3 | 39.4 | 39.6 |
  | scalar 1x4 f16 | 32 | 43.1 | 43.0 | 43.4 |
  | scalar 1x4 bf16 | 256 | 48.9 | 49.4 | 50.0 |
  | scalar 1x4 f16 | 256 | 50.3 | 50.0 | 50.7 |
  | scalar 4x4 bf16 | 32 | 28.0 | 28.1 | 28.2 |
  | scalar 4x4 f16 | 32 | 29.7 | 29.4 | 29.8 |
  | scalar 4x4 bf16 | 256 | 31.8 | 31.8 | 32.0 |
  | scalar 4x4 f16 | 256 | 32.0 | 32.1 | 32.1 |

### Cortex-X1 (Pixel 7 Pro)
| Kernel | BL | 128/128/1024 | 128/4096/1024 | 128/11008/4096 |
  |---|---|---|---|---|
  | scalar 1x4 bf16 | 32 | 19.3 | 19.1 | 19.4 |
  | scalar 1x4 f16 | 32 | 19.9 | 19.7 | 20.2 |
  | scalar 1x4 bf16 | 256 | 25.1 | 25.1 | 25.3 |
  | scalar 1x4 f16 | 256 | 23.3 | 23.3 | 23.8 |
  | scalar 4x4 bf16 | 32 | 11.4 | 11.3 | 11.3 |
  | scalar 4x4 f16 | 32 | 12.0 | 12.0 | 12.0 |
  | scalar 4x4 bf16 | 256 | 13.4 | 13.3 | 13.4 |
  | scalar 4x4 f16 | 256 | 12.9 | 12.8 | 12.9 |

### Broadwell
| Kernel | BL | 128/128/1024 | 128/4096/1024 | 128/11008/4096 |
[...]

PR #10602, opened 2026-06-26
Add support for neon dot qd8_bf16_qb4w

JakeStevens:
Add NEON dot-product (sdot) qd8_bf16_qb4w GEMM microkernels. Same sdot
path as the qd8_f16_qb4w neondot kernels; fp32 accumulation is rounded
to bf16 on store via vectorized round-to-nearest-even.

Test plan:
  cmake --build build-native --target qd8-bf16-qb4w-gemm-minmax-test
  ./build-native/test/qd8-bf16-qb4w-gemm-minmax-test  # 114/114 pass

Benchmark (Apple M4 Max, bench/models/llm_gemm.txt, BL=128),
4x4__scalar vs 4x16c4__neondot:

  M    N      K     scalar     neondot     speedup
  128  16     1024  32.1 GOPS  352.3 GOPS  ~11.0x
  128  128    1024  32.3 GOPS  370.2 GOPS  ~11.5x
  128  4096   1024  32.3 GOPS  366.2 GOPS  ~11.3x
  128  11008  4096  32.1 GOPS  368.5 GOPS  ~11.5x
  128  32000  4096  32.4 GOPS  366.9 GOPS  ~11.3x


JakeStevens:
Also fix the bf16 GEMM tester tolerance in gemm-microkernel-tester.cc
(max_abs_product 16 -> 256) so the bf16 output-rounding tolerance is
sized to the real output range, matching the f16/f32 qb4w paths.


This is dependent on #9960 which adds the scaffolding and scalar impl, the changes of importance here are just the dot changes. 

Once that PR lands, will rebase appropriately to make it easier to review at that point, but publishing early for feedback from @GregoryComer and to start the CLA process.

PR #10629, opened 2026-07-01
Add fp32->bf16 rewrite pass

GregoryComer:
Support rewriting bf16 subgraphs to run with excess precision (in f32) when native kernels aren't available. This generalizes the pattern recently introduced for f16 to also support bf16.

Relatively few bf16 ops are currently in-tree, so the initial implementation rewrites everything except reductions + transparent ops (copies/permutes/etc.). GEMMs will be wired up once the bf16 GEMM PRs merge.

## Test Plan
I verified that bazel + cmake build and tests pass on Mac. I've also verified this stack works e2e through executorch running gemma3-1b in bf16.


src/subgraph/rewrites/cvt_to_fp32.cc
```diff
@@ -43,10 +44,10 @@ bool ReplaceInSet(uint32_t* arr, uint32_t size, uint32_t old_value,
 
 enum class OpAction {
   kNone,             // Don't do anything, skip the node.
-  kRewrite,          // Force outputs to fp32, insert converts from fp16 input
-                     // to fp32
-  kNeedsFP16Inputs,  // The inputs must be converted back to fp16 if they were
-                     // rewritten.
+  kRewrite,          // Force outputs to fp32, insert converts from the source
+                     // 16-bit type to fp32
+  kNeedsFP16Inputs,  // The inputs must be converted back to the source 16-bit
```
Aelphy: kNeedsFP16Inputs -> kNeedsXF16Inputs?

Aelphy (commented):
Overall looks good, but please fix the constant name.

src/subgraph/rewrites/cvt_to_fp32.cc
```diff
@@ -320,23 +433,26 @@ CloneValueRet CloneValue(xnn_subgraph_t subgraph, xnn_value*& value) {
...
                                            XNN_INVALID_VALUE_ID);
-  // Maps fp32 value ids to the corresponding fp16 value id if a conversion
+  // Maps fp32 value ids to the corresponding `from_dt` value id if a conversion
   // has been inserted.
-  std::vector<uint32_t> fp32_id_to_fp16_id(subgraph->num_values,
+  std::vector<uint32_t> fp32_id_to_from_id(subgraph->num_values,
                                            XNN_INVALID_VALUE_ID);
 
-  xnn_log_debug("Running fp16 analysis and falling back to fp32.");
+  xnn_log_debug("Running 16-bit analysis and falling back to fp32.");
```
qukhan: Everywhere you are using the `"16-bit"` string, you can use `xnn_datatype_to_string` instead to make the message clearer.

qukhan (changes requested):

src/subgraph/rewrites/cvt_to_fp32.cc
GregoryComer: Updated, thanks.

src/subgraph/rewrites/cvt_to_fp32.cc
GregoryComer: Updated, thanks.

Aelphy (approved):

qukhan (approved):

qukhan:
Ah, the GN build hasn't been updated with the file changes.

If you want to test it locally, you can follow the GN actions' docker setup. Here's a quick rundown:

```bash
# Assuming /tmp/xnnpack/xnnpack is the git root.
#
# Note: the useradd stuff is to avoid being root when editing files during debugging in the mounted folder.
docker run --name xnnpack -it -v /tmp/xnnpack:/mnt -u 0 -w /mnt \
  ghcr.io/google/xnnpack/standard:latest /bin/bash -c \
  "export CC=gcc-9 CXX=g++-9 && useradd -m -u $(id -u) ${USER} && exec su ${USER} -s /bin/bash"
```

Then within the container, in `/mnt`.

```bash
# Clone depot tools and add them to the PATH.
git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git ~/depot_tools
export PATH=$HOME/depot_tools:$PATH
# Configure gclient and sync dependencies
cat <<EOF > /tmp/xnnpack/.gclient
solutions = [
  { "name"        : 'xnnpack',
    "url"         : 'https://github.com/google/XNNPACK',
    "deps_file"   : 'DEPS',
    "managed"     : False,
    "custom_deps" : {},
    "custom_vars": {},
  },
]
EOF
gclient sync --reset --force --upstream
# Build
cd xnnpack
gn gen --check --args='is_debug=false symbol_level=0 dcheck_always_on=true cc_wrapper="ccache" target_cpu="x64"' out/x64.dchecks
autoninja -C out/x64.dchecks
```

The container may be missing some build dependencies. You can either remove the `useradd` stuff from the `docker run` command or start another shell in the same container using `docker exec -it /bin/bash` and install the missing packages from there.

Note: this was written from head so ymmv, I may have missed a step. The GH actions are the SoT.

GregoryComer:
Thanks, I'll try to run through that locally. I also rebased onto master, which fixed CI on another PR.

There was one legitimate test failure for unary bf16 sqrt. With the subgraph rewrite, it falls back to f32 and the approximation that is used on x86 appears to break for very large values (test fails at like 10^19). The existing tests call this out and limit the range for f32 sqrt, so I've applied the same fix for bf16.

qukhan (approved):

GregoryComer:
I've updated the GN build and updated the sqrt tolerance (another rounding issue with very large values - test now limits to 10^30).

src/subgraph/rewrites/BUILD
dsharlet, line 20: This needs to add a dep on `"//:math"` (our internal build is failing with this PR due to this missing dep and blocks submitting this change).

GregoryComer: Thanks, updated.

PR #10633, opened 2026-07-01
Fix RISC-V rvvfp16 f16-vapproxgelu: use existing u4v unroll

velonica0:
PR #10403 changed the rvvfp16arith f16-vapproxgelu micro-kernel selected in src/configs/unary-elementwise-config.c from rational_6_4_div_u2v to _u8v. No u8v variant exists: the rvvfp16arith-rational-6-4 template only supports LMUL in {1, 2, 4}, so no u8v kernel is generated (LMUL=8 leaves no vector registers for the rational-6/4 division temporaries).

The production micro-kernel list is derived from the kernels referenced by src/configs/*-config.c, so pointing at the non-existent u8v kernel caused two failures whenever XNNPACK_ENABLE_RISCV_FP16_VECTOR is enabled:

  * link error: undefined reference to xnn_f16_vapproxgelu_ukernel__rvvfp16arith_rational_6_4_div_u8v
  * f16-vapproxgelu RVV kernels dropped out of the PROD list (u2v demoted to non-prod), so the shared/production library contains no RVV vapproxgelu kernel at all.

Select the largest unroll the template supports (u4v), fix element_tile, and regenerate the micro-kernel lists via tools/update-microkernels.py so the kernel is promoted back to PROD.


velonica0:
Hi, @fbarchard
Could you please take a look at this PR when you have a moment? It involves only a minor change. Thanks for your help.

fbarchard (approved):
Looks good.  m4 makes sense.
internally we cant test f16 due to a qemu config issue

PR #10661, opened 2026-07-06
Validate maxpool indirection buffer size

bg0d-droid:
Summary:
- validate max-pooling indirection buffer sizing with checked arithmetic before allocation
- return an early reshape error when the computed indirection buffer size cannot be represented
- add 32-bit regression coverage for the direct operator and subgraph/runtime reshape paths

Test plan:
- i386 ASAN/UBSAN build of `max-pooling-nhwc-test` and `max-pooling-2d-test`
- `./build-i386-asan-clang-full/test/operators/max-pooling-nhwc-test --gtest_filter=MAX_POOLING_NHWC_F32.indirection_buffer_size_overflow_32bit`
- `./build-i386-asan-clang-full/test/subgraph/max-pooling-2d-test --gtest_filter=MaxPooling2DQU8.IndirectionBufferSizeOverflow32BitSubgraph`

### DIFF
--- a/test/operators/max-pooling-nhwc.cc
+++ b/test/operators/max-pooling-nhwc.cc
@@ -8,6 +8,8 @@
 
 #include <cassert>
 #include <cstddef>
+#include <limits>
+#include <memory>
 
 #include <gtest/gtest.h>
 #include "include/xnnpack.h"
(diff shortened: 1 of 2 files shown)


test/subgraph/max-pooling-2d.cc
```diff
@@ -154,4 +165,71 @@ TEST(MaxPooling2DQU8, test) { TestImpl<quantized<uint8_t>>(); }
 TEST(MaxPooling2DF16, test) { TestImpl<xnn_float16>(); }
 TEST(MaxPooling2DF32, test) { TestImpl<float>(); }
 
+TEST(MaxPooling2DQU8, IndirectionBufferSizeOverflow32BitSubgraph) {
+  if (sizeof(void*) != 4) {
+    GTEST_SKIP() << "32-bit size_t overflow regression test";
+  }
```
qukhan, line 161: You are checking `sizeof(void*)` but the log messages talks about the size of `size_t`.

test/operators/max-pooling-nhwc.cc
```diff
@@ -16,6 +18,59 @@
...
 
+namespace {
+
+bool IsEarlyOverflowRejection(enum xnn_status status) {
+  return status == xnn_status_invalid_parameter ||
+         status == xnn_status_unsupported_parameter ||
+         status == xnn_status_out_of_memory;
+}
+
+}  // namespace
```
qukhan: There's only one test using this function. You're checking for 3 different return values. Which one are you expecting? Is it really necessary to add an extra helper function that is used only once? Why not directly check against the status value that you expect?

test/subgraph/max-pooling-2d.cc
```diff
@@ -26,6 +27,16 @@ using testing::Pointwise;
...
 
+namespace {
+
+bool IsMaxPoolOverflowEarlyRejection(enum xnn_status status) {
+  return status == xnn_status_invalid_parameter ||
+         status == xnn_status_unsupported_parameter ||
+         status == xnn_status_out_of_memory;
+}
+
+}  // namespace
```
qukhan: Same comment as in the other file.

qukhan (changes requested):

qukhan:
I would like a commit message that is a bit more descriptive than *"Add maxpool indirection buffer size tests"*. What conditions are the tests exercising?

bg0d-droid:
Updated, thanks. The commit message now describes the 32-bit maxpool indirection overflow condition, and the tests now check the exact expected `xnn_status_out_of_memory` status.

qukhan (approved):

PR #10686, opened 2026-07-08
KFI-543 Integrate SME2 kernels matmul_clamp_f16_qai8dxp_qsi8cxp

Colm-in-Arm:
This change integrates the aarch64 SME2 matmul_clamp_f16_qai8dxp_qsi8cxp path for FP16-output dynamically quantized fully connected workloads with QC8 weights. It enables packed-QP8 LHS execution where the graph and source  datatype match the kernel contract.

### KleidiAi kernels integrated
* kai_matmul_clamp_f16_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme2_mopa
* kai_matmul_clamp_f16_qai8dxp1x4_qsi8cxp4vlx4_1x4vl_sme2_dot

### Specific changes to note

**src/operator-run.c**
Updates QP8 GEMM output column stride handling to use the operator’s configured output element size instead of  assuming sizeof(float). This is needed because QP8 kernels can now produce either FP32 or FP16 outputs.

**src/subgraph.c**
Extends packed-LHS optimization for qdint8 -> fp16 fully connected nodes with QC8 weights to select the qp8_f16_qc8w path when the original pre-quantized input is FP16. The rewrite is gated on the convert source  datatype so FP32 sources remain on the existing QD8 path.

**src/xnnpack/simd/f16-scalar.h**
Corrects the scalar FP16 SIMD FMA capability flag. Plain scalar _Float16 expressions do not explicitly perform fused operations, so the scalar backend reports non-fused rounding semantics when native FP16 is  available.


src/configs/gemm-config.c
```diff
@@ -3007,6 +3009,44 @@ static void init_qp8_f32_qc8w_gemm_config(void) {
...
+        xnn_pack_kai_qs8_weights_and_biases;
+    qp8_f16_qc8w_gemm_config.packed_stride_weights_and_biases =
+        xnn_packed_stride_kai_qs8_weights_and_biases;
+    qp8_f16_qc8w_gemm_config.mr = mr;
+    qp8_f16_qc8w_gemm_config.nr = nr;
+    qp8_f16_qc8w_gemm_config.log2_kr = 2;
+    qp8_f16_qc8w_gemm_config.mr_packed = mr;
+#endif  // XNN_ENABLE_ARM_SME2
+  }
+  assert(qp8_f16_qc8w_gemm_config.mr <= XNN_MAX_MR);
```
qukhan: If `XNN_ARCH_ARM64 && XNN_ENABLE_KLEIDIAI` is true and `XNN_ENABLE_ARM_SME2` is false, then this assert will always fire.

src/qp8-f16-qc8w-gemm/qp8-f16-qc8w-gemm-minmax-1x64c4-neonsme2.c
```diff
@@ -0,0 +1,46 @@
...
+      (const struct xnn_f16_minmax_params*) minmax_params;
+  kai_run_matmul_clamp_f16_qai8dxp1x4_qsi8cxp4vlx4_1x4vl_sme2_dot(
+      m, n, k, lhs_packed, rhs_packed, dst, dst_stride_row, dst_stride_col,
+      xnn_float16_to_float(f16_minmax_params->scalar.min),
+      xnn_float16_to_float(f16_minmax_params->scalar.max));
+#else
+  assert(
+      "Calling KleidiAI microkernel wrapper, but XNNPACK was compiled without "
+      "`XNN_ENABLE_KLEIDIAI`." &&
+      0);
```
qukhan: The indentation is weird.

qukhan (changes requested):

src/configs/gemm-config.c
Colm-in-Arm: I think its ok. When XNN_ENABLE_ARM_SME2 is false `mr` remains 0. However, it looks a bit awkward. I'll move it inside the XNN_ENABLE_ARM_SME2 block.

src/qp8-f16-qc8w-gemm/qp8-f16-qc8w-gemm-minmax-1x64c4-neonsme2.c
Colm-in-Arm: Indeed. I'll reformat.

src/qp8-f16-qc8w-gemm/qp8-f16-qc8w-gemm-minmax-16x64c4-neonsme2.c
```diff
@@ -0,0 +1,47 @@
...
+      (const struct xnn_f16_minmax_params*) minmax_params;
+  kai_run_matmul_clamp_f16_qai8dxp1vlx4_qsi8cxp4vlx4_1vlx4vl_sme2_mopa(
+      m, n, k, lhs_packed, rhs_packed, dst, dst_stride_row, dst_stride_col,
+      xnn_float16_to_float(f16_minmax_params->scalar.min),
+      xnn_float16_to_float(f16_minmax_params->scalar.max));
+#else
+  assert(
+      "Calling KleidiAI microkernel wrapper, but XNNPACK was compiled without "
+      "`XNN_ENABLE_KLEIDIAI`." &&
+      0);
```
qukhan: The indentation is also weird here.

src/xnnpack/simd/f16-scalar.h
```diff
@@ -53,18 +53,11 @@ static XNN_INLINE xnn_simd_f16_t xnn_mul_f16(xnn_simd_f16_t a,
...
-#if ((XNN_ARCH_X86 || XNN_ARCH_X86_64) && defined(__FMA__)) || \
-    (XNN_ARCH_ARM64 && __ARM_FEATURE_FMA && defined(__ARM_FEATURE_FP16_FML))
-#define XNN_SIMD_HAS_NATIVE_FMA 1
-#else
 #define XNN_SIMD_HAS_NATIVE_FMA 0
-#endif  // ((XNN_ARCH_X86 || XNN_ARCH_X86_64) && defined(__FMA__)) ||
-        // (XNN_ARCH_ARM64 && __ARM_FEATURE_FMA &&
-        // defined(__ARM_FEATURE_FP16_FML))
 #else
+// Computing f16 fused ops via float rounds once at the final f16 conversion.
```
dsharlet: This doesn't seem right. computing fp16 fma with floats would do a double rounding, wouldn't it?

(Interestingly, doing fp16 fma with *fp64* does avoid double rounding, because the mantissa is big enough to hold any intermediate sum. But we don't do fma with fp64.)

Colm-in-Arm: This was quite the rabbit hole when I first encountered it. It started with these two test failures on M4:

[ RUN      ] F16SimdSCALARTest.Fmsub
test/simd/f16-simd-scalar.cc:132: Failure
  output_f32[k]
    Which is: 46.4375
  TruncToF16(inputs_f32[k] * inputs_f32[k + 1] - inputs_f32[k + 2 * 1])
    Which is: 46.46875
Google Test trace:
./test/replicable_random_device.h:92: To replicate this failure, re-run the test with `--gtest_random_seed=36814`.

[  FAILED  ] F16SimdSCALARTest.Fmsub (0 ms)
[ RUN      ] F16SimdSCALARTest.Fnmadd
test/simd/f16-simd-scalar.cc:156: Failure
  output_f32[k]
    Which is: -46.4375
  TruncToF16(-inputs_f32[k] * inputs_f32[k + 1] + inputs_f32[k + 2 * 1])
    Which is: -46.46875
Google Test trace:
./test/replicable_random_device.h:92: To replicate this failure, re-run the test with `--gtest_random_seed=36814`.

When I added some printouts all fingers pointed to double rounding and setting XNN_SIMD_HAS_NATIVE_FMA reliably solved the problem. However, when I look at it again today, things aren't quite so clear. Rolling back this change makes this failure quite intermittent 9/20 runs of it will fail on M4. I'll continue to investigate.

Colm-in-Arm: I believe I’ve found the root cause of these test failures.

The code above determines whether native FMA is available for the compilation target, and the tests use XNN_SIMD_HAS_NATIVE_FMA to decide whether to expect fused results. However, the workspace .bazelrc specifies --copt='-ffp-contract=off'. With that option, the compiler is not allowed to contract ordinary multiply/add expressions into FMA instructions, even if the target hardware supports FMA.

That means the implementation can legally compute the multiply and add/subtract as separate operations, while the test expects single-rounding fused behavior if it has determined the hardware supports it. For some random seeds this exposes a double-rounding difference, which explains the intermittent failures.

I’ll revert the change above. The follow-up question is whether the F16/F32 SIMD tests should be changed so that fused expectations are only used for implementations that explicitly perform FMA, rather than simply when the target has native FMA support.

src/qp8-f16-qc8w-gemm/qp8-f16-qc8w-gemm-minmax-16x64c4-neonsme2.c
Colm-in-Arm: I should have checked for that when I fixed the earlier one. Done now.

src/xnnpack/simd/f16-scalar.h
dsharlet: > However, the workspace .bazelrc specifies --copt='-ffp-contract=off'.

I think this is what we want, we want all fma usage to be explicitly opted-in to.

> while the test expects single-rounding fused behavior if it has determined the hardware supports it. 

I think the bug is here: https://github.com/google/XNNPACK/blob/f6d310ea256480361335d3370f3b61e0cea75a2a/src/xnnpack/simd/f16-scalar.h#L74-L80

It looks like we think that if `a`, `b` are a real `xnn_simd_f16_t`, that `a * b` is also a float16, but that's probably wrong. Maybe we should explicitly cast `a * b` to `xnn_simd_f16_t`, like the non-native branch below.

dsharlet: I actually, I mis-read it. The `#else` branch also does the arithmetic with a float intermediate.

I suspect this doesn't actually matter, and we should relax the test in the non-fma case to permit this implementation.

qukhan (approved):

qukhan:
Hi @Colm-in-Arm, merging is failing because of the following job failure: https://github.com/google/XNNPACK/actions/runs/30280144537/job/90024151613

src/xnnpack/simd/f16-scalar.h
Colm-in-Arm: To manage relaxing the tests I introduced the concept of XNN_SIMD_TEST_USES_EXPLICIT_FMA in `test/simd/f16-simd.cc.in`. For the generated tests where explicit FMA can be called this macro is enabled. For those tests we can then rely on XNN_SIMD_HAS_NATIVE_FMA to decide how the float rounding will be handled.

Colm-in-Arm:
> Hi @Colm-in-Arm, merging is failing because of the following job failure: https://github.com/google/XNNPACK/actions/runs/30280144537/job/90024151613

This should be fixed by the most recent commit.

qukhan (approved):

PR #10690, opened 2026-07-09
[RVV] add rvv f32 kernel for ppmm

velonica0:
Add RVV kernel for f32-ppmm with MR={1,2,4} and NR={m1,m2,m4} variants. Tested on SpacemiT K1 CPU and K3 CPU, both VLEN=256.

| Operator | Workload | K1 Scalar (ns) | K1 RVV (ns) | K1 Speedup | K3 Scalar (ns) | K3 RVV (ns) | K3 Speedup |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| f32-ppmm | ALBERT | 153,322,965 | 29,111,495 | 5.3x | 77,835,415 | 9,417,268 | 8.3x |
| f32-ppmm | MobileBERT | 26,477,852 | 5,963,739 | 4.4x | 12,942,549 | 1,552,870 | 8.3x |


cmake/gen/rvv_microkernels.cmake
```diff
@@ -167,10 +166,17 @@ SET(NON_PROD_RVV_MICROKERNEL_SRCS
...
+  src/f32-ppmm/gen/f32-ppmm-1x1v-minmax-rvv.c
+  src/f32-ppmm/gen/f32-ppmm-1x2v-minmax-rvv.c
+  src/f32-ppmm/gen/f32-ppmm-1x4v-minmax-rvv.c
+  src/f32-ppmm/gen/f32-ppmm-2x1v-minmax-rvv.c
+  src/f32-ppmm/gen/f32-ppmm-2x2v-minmax-rvv.c
+  src/f32-ppmm/gen/f32-ppmm-2x4v-minmax-rvv.c
+  src/f32-ppmm/gen/f32-ppmm-4x1v-minmax-rvv.c
+  src/f32-ppmm/gen/f32-ppmm-4x2v-minmax-rvv.c
+  src/f32-ppmm/gen/f32-ppmm-4x4v-minmax-rvv.c
   src/f32-qs8-vcvt/gen/f32-qs8-vcvt-rvv-u1v.c
```
ken-unger, line 180: Perhaps I'm wrong, but I think you need to run 'tools/update-microkernels.py' one more time which will move 4x4v into the PROD list here and for bzl.

ken-unger: Sorry, ignore my comment as I see that there is no config for f32-ppmm.  (so I don't think it is actually used?).

PR #10702, opened 2026-07-10
litert/tensor/arithmetic.h: bounds-check axis and index values before vector subscript

destro4evr-rgb:
Addresses VRP report https://issuetracker.google.com/issues/533344097

Several graph-building functions in `litert/tensor/arithmetic.h` use axis values or buffer-derived indices as direct subscripts into heap-allocated `std::vector` shape vectors without validating that the values are within bounds. `std::vector::operator[]` is unchecked in release builds.

**Concatenation** — `axis` is passed directly by the caller and used as a subscript into the output and input shape vectors with no prior validation. A negative value implicitly casts to a SIZE_MAX-class `size_t`, directing the read and write gigabytes past the heap allocation.

**ReduceMax, Mean, Sum** — Each function reads axes from a tensor buffer as `int32_t` values and uses them directly as subscripts into the output shape vector (`o_info.shape[b_data[0]]`, `o_info.shape[b_data[i]]`). A value of -1 in the buffer becomes SIZE_MAX after the implicit cast to `size_t`.

**Transpose** — The permutation buffer is read as `int32_t` values and used to index into `input_shape` inside a loop. A negative permutation value produces a SIZE_MAX-class index. Additionally, if the permutation buffer contains more entries than the input rank, the loop writes to `output_info.shape[i]` beyond the resized vector length.

**Pad and PadV2** — The loop reads `b_data[i * 2]` and `b_data[i * 2 + 1]` for every dimension, but the buffer length is not checked against `2 * input_rank` before the loop. A buffer shorter than `2 * input_rank` elements causes reads past the buffer's allocation.

## Fix

Validate axis/index values against `[0, shape.size())` before any subscript, and return an `ErrorTensor` on failure. For Pad and PadV2, confirm the locked buffer holds at least `2 * o_info.shape.size()` elements before entering the loop.

This follows the same pattern introduced in #10464 for `ExpandDims`, `Pack`, and `OneHot`.

## Affected functions

| Function | Issue | Fixed |
|---|---|---|
| `Concatenation` | `axis` unchecked before `shape[axis]` | ✓ |
| `ReduceMax` | `b_data[0]` / `b_data[i]` unchecked before `shape[b_data[*]]` | ✓ |
| `Mean` | same as ReduceMax | ✓ |
| `Sum` | same as ReduceMax | ✓ |
| `Transpose` | `perm_data[i]` unchecked; perm length vs input rank unchecked | ✓ |
| `Pad` | buffer length not checked against `2 * input_rank` before loop | ✓ |
| `PadV2` | same as Pad | ✓ |

## Precedent

PR #10464 ("Bounds-check data-derived axes and quantization params in tensor builder", merged 2026-07-02) fixed the same bug class in `ExpandDims`, `Pack`, and `OneHot` in the same file. The fix here applies the identical pattern to the remaining unchecked functions.

### DIFF
--- a/litert/tensor/arithmetic.h
+++ b/litert/tensor/arithmetic.h
@@ -589,6 +589,10 @@ Tensor<Mixins...> Pad(Tensor<Mixins...> a, Tensor<Mixins...> b,
   if (b_info.buffer) {
     const LockedBufferSpan<const int32_t> b_lock =
         b_info.buffer->Lock().As<const int32_t>();
+    if (b_lock.size() < 2 * o_info.shape.size()) {
+      return Tensor<Mixins...>(graph::ErrorTensor(absl::InvalidArgumentError(
+          "Padding buffer size must be at least 2 * input rank.")));
+    }
     const int32_t* b_data = b_lock.data();
     for (int i = 0; i < o_info.shape.size(); ++i) {
       o_info.shape[i] += b_data[i * 2] + b_data[i * 2 + 1];
@@ -617,6 +621,10 @@ Tensor<Mixins...> PadV2(Tensor<Mixins...> a, Tensor<Mixins...> b,
   if (b_info.buffer) {
     const LockedBufferSpan<const int32_t> b_lock =
         b_info.buffer->Lock().As<const int32_t>();
+    if (b_lock.size() < 2 * o_info.shape.size()) {
+      return Tensor<Mixins...>(graph::ErrorTensor(absl::InvalidArgumentError(
+          "Padding buffer size must be at least 2 * input rank.")));
+    }
     const int32_t* b_data = b_lock.data();
     for (int i = 0; i < o_info.shape.size(); ++i) {
       o_info.shape[i] += b_data[i * 2] + b_data[i * 2 + 1];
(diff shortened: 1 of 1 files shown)


destro4evr-rgb:
@qukhan  - flagging this for review. This PR addresses heap out-of-bounds reads in several litert/tensor/arithmetic.h graph-building functions (Concat, ReduceMax, Mean, Sum, Transpose, Pad) where attacker-controlled axis values or buffer indices are used as unchecked subscripts into shape vectors.

qukhan:
There was a hiccup in our import flow, I think it should be good now.

destro4evr-rgb:
Hi @qukhan , both Copybara PRs (#10754 on XNNPACK and [#8717](https://github.com/google-ai-edge/LiteRT/pull/8717) on LiteRT) have now completed all CI checks cleanly. 
Could you take a look and merge when you get a chance?

destro4evr-rgb:
Hey @qukhan , rebased the PR onto current master to resolve the conflict from the LockedBufferSpan refactor. Should be clean now."

qukhan:
Because #10765 got through and there's a merge conflict, this needs to be rebased again.

destro4evr-rgb:
Hi @qukhan , rebased again onto current master - conflict from #10765 is resolved. CI should be clean. Ready for merge when you get a chance.

PR #10710, opened 2026-07-11
Fix integer overflow in NHWC convolution kernel size calculation

YellowJune:
## Summary
This PR fixes a critical integer overflow vulnerability in the NHWC convolution operator. The overflow occurred during the calculation of the kernel size (kernel_height * kernel_width), leading to incorrect microkernel selection and undersized heap memory allocation.

## Problem
In src/operators/convolution-nhwc.c, the kernel size was calculated using 32-bit integer arithmetic before being widened to size_t. When the product of kernel_height and kernel_width exceeds 2^32 (e.g., 65536 x 65536), the value wraps to zero.

This zero-value propagates through the system, causing:
1. **Incorrect Microkernel Selection**: find_dwconv_ukernel(0) matches any registered kernel, regardless of the actual filter size.
2. **Undersized Allocation**: The packed_weights buffer is allocated only for bias terms, leaving no space for filter weights.
3. **Heap Buffer Overflow**: During weight packing, the actual (large) dimensions are used to write into the undersized buffer, resulting in an out-of-bounds heap write.

## Fix
1. **Added Overflow Guard**: Introduced a check in select_microkernel_type to reject kernel dimensions that would overflow size_t.
2. **Size_t Arithmetic**: Applied explicit size_t casting before multiplication in all critical paths (creation, reshape, and microkernel selection) to ensure 64-bit arithmetic is used.
3. **Hardened Reshape Path**: Updated effective_kernel_height/width calculations to prevent similar overflows when large dilations are applied.

This fix aligns with previous security patches applied to deconvolution (#9876) and pooling/NCHW convolution (commit 32e57d1).

## Verification Results
- Verified that 65536 x 65536 is correctly calculated as 4,294,967,296 instead of 0.
- Confirmed the overflow guard correctly triggers for values exceeding SIZE_MAX.

### DIFF
--- a/src/operators/convolution-nhwc.c
+++ b/src/operators/convolution-nhwc.c
@@ -563,7 +563,14 @@ struct convolution2d_nhwc_variant {
 static enum xnn_status select_microkernel_type(
     const struct convolution2d_nhwc_variant* variant,
     struct convolution2d_nhwc_context* context) {
-  const size_t kernel_size = context->kernel_height * context->kernel_width;
+  size_t kernel_size;
+  if (!xnn_safe_mul(context->kernel_height, context->kernel_width, &kernel_size)) {
+    xnn_log_error(
+        "failed to create %s operator: kernel size (%" PRIu32 "x%" PRIu32 ") overflows size_t",
+        xnn_operator_type_to_string(context->operator_type),
+        context->kernel_width, context->kernel_height);
+    return xnn_status_unsupported_parameter;
+  }
   const bool unit_subsampling = (context->subsampling_width | context->subsampling_height) == 1;
   const bool any_padding = (context->input_padding_left | context->input_padding_top |
                             context->input_padding_right | context->input_padding_bottom) != 0;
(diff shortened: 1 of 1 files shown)


src/operators/convolution-nhwc.c
```diff
@@ -563,7 +563,14 @@ struct convolution2d_nhwc_variant {
 static enum xnn_status select_microkernel_type(
     const struct convolution2d_nhwc_variant* variant,
     struct convolution2d_nhwc_context* context) {
-  const size_t kernel_size = context->kernel_height * context->kernel_width;
+  if ((size_t) context->kernel_height > SIZE_MAX / (size_t) context->kernel_width) {
```
qukhan: Please, use `xnn_safe_mul`.

qukhan (changes requested):

src/operators/convolution-nhwc.c
YellowJune: Done. Replaced the manual overflow checks with xnn_safe_mul in the affected code paths. Thanks for the suggestion!

qukhan (approved):

YellowJune:
Does the merge process automatically after copybara validation? Or, do you handle it manually after the full approval?

qukhan:
It will process automatically.

YellowJune:
> It will process automatically.

However, I see that "1 workflow awaiting approval", which means task requires approval from a maintainer.. Can you please recheck about this? Best regards,

YellowJune:
@qukhan hello, i think it is some.. waiting for being reported, and its stucked in there for much time

qukhan:
Sorry for the delay, I restarted the process to merge this PR.

YellowJune:
> Sorry for the delay, I restarted the process to merge this PR.

Thanks. Im first to merge PR this time, so idk about processing time and average merging time. Can you tell me estimated time for automatically merged?

qukhan:
The PR has been merged.

As some internal tests run on the PRs before being merged, there's no good estimation for the time it could take.

PR #10748, opened 2026-07-14
validate runtime input channels in convolution reshape paths

aizu-m:
check the input channel dimension against the operator's create-time channels in the convolution, depthwise-convolution and deconvolution reshape paths, otherwise an input reshaped to fewer channels via xnn_reshape_external_value leaves the indirection buffer striding past the end of the shorter input during the igemm read.

### DIFF
--- a/src/subgraph/convolution-2d.c
+++ b/src/subgraph/convolution-2d.c
@@ -692,8 +692,7 @@ enum xnn_status reshape_convolution_operator(struct xnn_operator_data* opdata,
         input_value->shape.dim[input_value->shape.num_dims - 1];
     if (input_channels != convolution_op->input_pixel_stride) {
       xnn_log_error("failed to reshape %s operator with input ID #%" PRIu32
-                    ": input channels (%zu) must match the operator's number "
-                    "of input channels (%zu)",
+                    ": mismatch at channel dimension (%zu != %zu)",
                     xnn_node_type_to_string(xnn_node_type_convolution_2d),
                     input_id, input_channels,
                     convolution_op->input_pixel_stride);
--- a/src/subgraph/convolution-2d.c
+++ b/src/subgraph/convolution-2d.c
@@ -682,6 +682,25 @@ enum xnn_status reshape_convolution_operator(struct xnn_operator_data* opdata,
   const size_t input_height = input_value->shape.dim[1];
   const size_t input_width = input_value->shape.dim[2];
 
+  // The indirection buffer strides across input pixels by the operator's
+  // create-time input_pixel_stride, so an NHWC input carrying fewer channels
+  // than that makes the microkernel read past the end of the input buffer.
+  const xnn_operator_t convolution_op = opdata->operator_objects[0];
+  if (convolution_op->type != xnn_operator_type_convolution_nchw_f16 &&
+      convolution_op->type != xnn_operator_type_convolution_nchw_f32) {
+    const size_t input_channels =
+        input_value->shape.dim[input_value->shape.num_dims - 1];
+    if (input_channels != convolution_op->input_pixel_stride) {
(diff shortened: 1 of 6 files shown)


src/subgraph/convolution-2d.c
```diff
@@ -682,6 +682,25 @@ enum xnn_status reshape_convolution_operator(struct xnn_operator_data* opdata,
...
+  // create-time input_pixel_stride, so an NHWC input carrying fewer channels
+  // than that makes the microkernel read past the end of the input buffer.
+  const xnn_operator_t convolution_op = opdata->operator_objects[0];
+  if (convolution_op->type != xnn_operator_type_convolution_nchw_f16 &&
+      convolution_op->type != xnn_operator_type_convolution_nchw_f32) {
+    const size_t input_channels =
+        input_value->shape.dim[input_value->shape.num_dims - 1];
+    if (input_channels != convolution_op->input_pixel_stride) {
+      xnn_log_error("failed to reshape %s operator with input ID #%" PRIu32
+                    ": input channels (%zu) must match the operator's number "
```
dsharlet: I think this error message should read ``"failed to reshape %s operator with input_a ID #%" PRIu32 " and input_b ID #%" PRIu32 ": mismatch at channel dimension (%zu != %zu)"`. The operator stuff is an internal implementation detail.

This error message should basically be almost the same as this one: https://github.com/google/XNNPACK/blob/44db232920d81d2993720e349a55918c4524e5b6/src/subgraph/batch-matrix-multiply.c#L331-L338

Same issue in the other two ops.

Also, it seems like for all of these (convolution-2d, deconvolution-2d, depthwise-convolution-2d, batch-matrix-multiply, fully-connected) the check should be in the operator reshape function, not the subgraph. Otherwise, this hole still exists in operator API usage (but direct operator API usage is rare, the subgraph API is our main public API).

test/subgraph/convolution-2d.cc
```diff
@@ -375,4 +375,48 @@ TEST(Convolution2DQD8F32QC8W, test) {
   TestImpl<float, qcint8, float>(/*convert_to=*/xnn_datatype_qdint8);
 }
 
+#ifndef XNNPACK_USE_YNNPACK
```
dsharlet: Did these tests actually fail with YNNPACK? I think it should validate this already.

src/subgraph/convolution-2d.c
aizu-m: Reworded all three to `mismatch at channel dimension (%zu != %zu)` like the batch-matrix-multiply check, and dropped the operator internals from the message text.

On where the check lives: batch-matrix-multiply and fully-connected both do this in the subgraph today, so I kept these three consistent with them. Moving it into the operator reshape means threading an input_channels argument through all the public xnn_reshape_convolution2d_nhwc_* and deconvolution entry points, which is a wider API change. I can do that as a separate follow-up across all five ops if you'd rather it sit there.

test/subgraph/convolution-2d.cc
aizu-m: Good question. I guarded them defensively because the check I added lives in src/subgraph, and I hadn't confirmed the YNNPACK runtime returns invalid_parameter for a channel mismatch (no YNNPACK build to hand here). Dropped the guard now on the basis that it validates already. If the internal build reports a different status for this case, let me know and I'll narrow it back.

src/subgraph/convolution-2d.c
dsharlet: Let's leave this as is for now, thanks.

test/subgraph/convolution-2d.cc
dsharlet: I'm pretty sure it will validate, but I'm less sure it will fail specifically with `xnn_status_invalid_parameter`. Let's see what happens.

PR #10756, opened 2026-07-15
Fix LUT fusion treating an already-fused LUT node as a plain unary op

sachinhambar:
`is_pure_unary_elementwise` treated any `xnn_node_type_unary_elementwise`
node as a plain single-input op, reading `inputs[0]` unconditionally.
But a node already fused into a LUT (via
`xnn_define_unary_elementwise_lut_in_place`) has `num_inputs == 2`
(tensor + LUT table) and `unary_operator == xnn_unary_invalid`. If the
fusion pass reaches that node again — the outer loop in
`xnn_subgraph_fuse_unary_quantized_into_lut` can revisit the same slot
within a single pass, and `xnn_subgraph_optimize` can also be called
more than once on the same subgraph — it would misinterpret the LUT
node as an ordinary unary op and try to fuse it again.

### DIFF
--- a/src/subgraph.c
+++ b/src/subgraph.c
@@ -1790,7 +1790,6 @@ static uint32_t is_pure_unary_elementwise(xnn_subgraph_t subgraph,
                                           uint32_t num_unary_values) {
   switch (node->type) {
     case xnn_node_type_unary_elementwise:
-      assert(node->num_inputs >= 1);
       // A pure unary elementwise function takes exactly one tensor input and
       // must have a valid operator. A node already fused into a LUT by this
       // same pass (see `xnn_define_unary_elementwise_lut_in_place`) has 2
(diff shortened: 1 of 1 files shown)


src/subgraph.c
```diff
@@ -1791,6 +1791,15 @@ static uint32_t is_pure_unary_elementwise(xnn_subgraph_t subgraph,
   switch (node->type) {
     case xnn_node_type_unary_elementwise:
       assert(node->num_inputs >= 1);
```
dsharlet: This `assert` should be moved after the new check.

dsharlet (changes requested):
Thanks for the PR!

src/subgraph.c
sachinhambar: Review comments incorporated

dsharlet (approved):

PR #10761, opened 2026-07-16
unpooling: guard indirection buffer size computation against integer overflow

destro4evr-rgb:
`xnn_reshape_unpooling2d_nhwc_x32` computed the indirection buffer size with a raw five-factor multiplication — `sizeof(void*) * batch_size * input_height * input_width * pooling_size` — with no overflow guard. When caller-supplied dimensions are large, the product silently wraps to a value far smaller than required. `xnn_reallocate_memory` returns a valid pointer to an undersized buffer, and `xnn_indirection_init_unpool2d` then writes the full (attacker-controlled) number of pointer entries out of bounds.

**Concrete overflow (64-bit, 2×2 pool):**
```
batch_size=1<<20, input_height=1<<20, input_width=1<<20, pooling_size=4
8 * (2^20 * 2^20 * 2^20 * 4) = 8 * 2^62 = 2^65 mod 2^64 = 0
```
`xnn_reallocate_memory(ptr, 0)` under glibc returns a non-NULL zero-byte allocation. `xnn_indirection_init_unpool2d` then writes 2^62 pointer-sized entries out of bounds.

**Fix:** replace the raw multiplication with chained `xnn_safe_mul` calls, matching the pattern already used in `argmax-pooling-nhwc.c` and `average-pooling-nhwc.c` after PR #10646.

### DIFF
--- a/src/operators/unpooling-nhwc.c
+++ b/src/operators/unpooling-nhwc.c
@@ -215,9 +215,22 @@ enum xnn_status xnn_reshape_unpooling2d_nhwc_x32(
 
   const size_t pooling_height = unpooling_op->convolution_op->kernel_height;
   const size_t pooling_width = unpooling_op->convolution_op->kernel_width;
-  const size_t pooling_size = pooling_height * pooling_width;
+  size_t pooling_size = 0;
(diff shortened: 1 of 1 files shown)


destro4evr-rgb:
Closing for now, will resubmit after further review.

PR #10762, opened 2026-07-16
litert/tensor/arithmetic.h: bounds-check axis in Split/Unpack and rank in Tile

destro4evr-rgb:
Three shape-inference functions in `litert/tensor/arithmetic.h` use attacker-controlled values as unchecked subscripts into heap-allocated shape vectors.

**Split** — after negative-axis normalization, `axis_val` is used as `input_info.shape[axis_val]` (OOB read) and `output_shape[axis_val] /= num_splits` (OOB write) without checking `axis_val < input_info.shape.size()`.

**Unpack** — `axis` is passed directly to `output_shape.erase(output_shape.begin() + axis)` with no range check. An out-of-range iterator to `erase()` is UB that corrupts the heap. The sibling `Pack()` already validates its axis (PR #10464); this aligns `Unpack` to the same pattern.

**Tile** — `output_info.shape` is resized to `input_shape.size()` (N elements) but the fill loop iterates `multiples_data.size()` (M) times. When M > N the loop writes `output_info.shape[i]` and reads `input_shape[i]` for `i >= N`, both past the end of their heap-allocated vectors.

Fixes follow the same error-return pattern used in PR #10464 and PR #10702.

### DIFF
--- a/litert/tensor/arithmetic.h
+++ b/litert/tensor/arithmetic.h
@@ -1429,6 +1429,10 @@ std::vector<Tensor<Mixins...>> Unpack(
   std::vector<Tensor<Mixins...>> outputs;
   outputs.reserve(num);
   const graph::TensorInformation& input_info = GetInfo(input.GetRaw()).value();
+  if (axis < 0 || axis >= static_cast<int>(input_info.shape.size())) {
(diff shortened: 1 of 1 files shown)


destro4evr-rgb:
Closing for now, will resubmit after further review.

PR #10787, opened 2026-07-18
Fix qb4 packw bias tail in neon

JakeStevens:
The kernel template allows for indexing past the end the bias, which can result in a SIGSEGV at packing. 

Here we handle the tail with a a memcpy + memset with proper bounds.

Added a test for these cases as well.

Also confirmed with actual usecases on device it resolves a SIGSEGV.


src/qb4-packw/gen/qb4-packw-x16c4-gemm-goi-aarch64-neondot.c
```diff
@@ -364,15 +365,20 @@ void xnn_qb4_packw_gemm_goi_ukernel_x16c4__aarch64_neondot(
...
+          vst1q_s32((int32_t*)out + 4, b1);
+          const int32x4_t b2 = vld1q_s32(&b[8]);
+          vst1q_s32((int32_t*)out + 8, b2);
+          const int32x4_t b3 = vld1q_s32(&b[12]);
+          vst1q_s32((int32_t*)out + 12, b3);
+        } else {
+          memcpy(out, b, n * sizeof(int32_t));
+          memset((int32_t*) out + n, 0, (16 - n) * sizeof(int32_t));
+        }
+        b += min((size_t) n, (size_t) 16);
```
qukhan: Could you update the template to integrate the `min` call into the branch you've added?

```suggestion
          b += 16;
        } else {
          memcpy(out, b, n * sizeof(int32_t));
          memset((int32_t*) out + n, 0, (16 - n) * sizeof(int32_t));
          b += n;
        }
```

qukhan (changes requested):

src/qb4-packw/gen/qb4-packw-x16c4-gemm-goi-aarch64-neondot.c
JakeStevens: done

qukhan (approved):

src/qb4-packw/gen/qb4-packw-x16c4-gemm-goi-aarch64-neondot.c
```diff
@@ -364,15 +365,21 @@ void xnn_qb4_packw_gemm_goi_ukernel_x16c4__aarch64_neondot(
...
-        const int32x4_t b0 = vld1q_s32(&b[0]);
-        vst1q_s32((int32_t*)out + 0, b0);
-        const int32x4_t b1 = vld1q_s32(&b[4]);
-        vst1q_s32((int32_t*)out + 4, b1);
-        const int32x4_t b2 = vld1q_s32(&b[8]);
-        vst1q_s32((int32_t*)out + 8, b2);
-        const int32x4_t b3 = vld1q_s32(&b[12]);
-        vst1q_s32((int32_t*)out + 12, b3);
-        b += 16;
+        if (n >= 16) {
```
fbarchard, line 368: this isnt right.  you copy 16 but not the rest.  it needs to be in a loop and copy n bias values.
ditto for scalar.

you can/should write the full NR (16 floats) but its important on the remainder to only read n to avoid out of bounds.
optional, but a good idea to clear the rest to zero.

fbarchard (changes requested):
wrong amount of bias values are copied.
the memcpy/memset version looks right.

src/qb4-packw/gen/qb4-packw-x16c4-gemm-goi-aarch64-neondot.c
fbarchard: disreguard... you are doing the neon right.  its a direct copy of 4 vectors when  n >= 16 
and memcpy/memset for remainder.

src/qb4-packw/gen/qb4-packw-x16c4-gemm-goi-scalar.c
```diff
@@ -327,23 +327,29 @@ void xnn_qb4_packw_gemm_goi_ukernel_x16c4__scalar(
...
+          ((uint32_t*) out)[6] = b[6];
+          ((uint32_t*) out)[7] = b[7];
+          ((uint32_t*) out)[8] = b[8];
+          ((uint32_t*) out)[9] = b[9];
+          ((uint32_t*) out)[10] = b[10];
+          ((uint32_t*) out)[11] = b[11];
+          ((uint32_t*) out)[12] = b[12];
+          ((uint32_t*) out)[13] = b[13];
+          ((uint32_t*) out)[14] = b[14];
+          ((uint32_t*) out)[15] = b[15];
```
fbarchard, line 346: this is fine and both this and memcpy(out, b, 16 * sizeof(uint32_t));
will vectorize to 4 vectors, but the unrolling exposes a pedantic compiler performance issue for very large NR ... e.g. NR=256 will be heavily unrolled and the compiler will spend a long time (more than 10 minutes) trying to compile this file.  It eventually does the right thing, which is a few vectors.

fbarchard (approved):
looks good.  I tried it on android and it runs smoothly

PR #10791, opened 2026-07-20
reallocate deconv igemm indirection buffer on output-size change

aizu-m:
Reshaping one f32 deconvolution twice, same 4x4 input, adjustment 0 then 1, under ASan:

    ERROR: AddressSanitizer: BUS on unknown address
    The signal is caused by a READ memory access.
    Hint: this fault was caused by a dereference of a high value address (x0 = 0xfffffffffffffff0)
      #0 xnn_f32_igemm_minmax_ukernel_4x2__asm_aarch64_neonfma_cortex_a75_prfm
      #2 xnn_run_operator_with_index operator-run.c:2240

reshape_igemm_path sizes the indirection buffer from the output,
round_up(output_h * output_w, mr) * kernel_size, but only reallocates and
refills it when the input height or width change. adjustment_height and
adjustment_width are reshape arguments and grow the output on their own, so the
same 4x4 input with adjustment 1 gives an 8x8 output while the input check stays
false. The buffer keeps its 49-pixel size and the igemm then walks it for 64
output rows, reading past the end and dereferencing the stale pointers there.

Found while diffing this against reshape_subconv2d_path lower in the file. The
subconv sibling already recognises an output or mr change as a resize; the igemm
gate only looked at the input. Same guard applied here, and the three tracking
fields it needs already exist on the operator.

Added a regression test under DECONVOLUTION_NHWC_F32: it faults on the old code
and passes with the change.

### DIFF
--- a/test/operators/deconvolution-nhwc.cc
+++ b/test/operators/deconvolution-nhwc.cc
@@ -4,9 +4,13 @@
 // LICENSE file in the root directory of this source tree.
 
 #include <cstddef>
+#include <limits>
+#include <memory>
 #include <vector>
 
 #include <gtest/gtest.h>
+#include "include/xnnpack.h"
+#include "src/xnnpack/buffer.h"
 #include "src/xnnpack/config-types.h"
 #include "src/xnnpack/config.h"
 #include "test/operators/deconvolution-operator-tester.h"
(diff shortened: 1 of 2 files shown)


aizu-m:
@googlebot I signed it.

aizu-m:
Rebased onto current master. The cla/google failure was a stale check from when the PR opened; everything is green now. Full deconvolution-nhwc test binary passes locally after the rebase, 1348 tests.

test/operators/deconvolution-nhwc.cc
```diff
@@ -3445,3 +3448,114 @@ TEST(DECONVOLUTION_NHWC, conversion_buffer_size_overflow) {
...
+  xnnpack::Buffer<float> input(
+      batch_size * input_height * input_width * groups * group_input_channels,
+      xnnpack::XnnExtraBytes);
+  for (size_t i = 0; i < input.size(); i++) {
+    input[i] = static_cast<float>(static_cast<int>(i % 11) - 5) * 0.125f;
+  }
+
+  const auto create_op = [&]() -> xnn_operator_t {
+    xnn_operator_t op = nullptr;
+    EXPECT_EQ(xnn_status_success,
```
qukhan: Please avoid yoda-style comparisons. Put the tested value on the left and the expected one on the right.

aizu-m: Done. Flipped all the assertions in the new test so the call under test sits on the left and xnn_status_success on the right, same for the dimension checks. Rebuilt and reran the full deconvolution-nhwc binary, 1348 tests pass.

qukhan:
The CLA check will not pass as long as you reference the AI tooling you've used as Co-Author in your commits.

Note: All submissions to Google Open Source projects need to follow Google’s Contributor License Agreement (CLA), which covers any original work of authorship included in the submission. This doesn’t prohibit the use of coding assistance tools, including tool-, AI-, or machine-generated code, as long as these submissions abide by the CLA's requirements.

aizu-m:
Understood. Traced it to a trailer on the last commit and a stray committer identity on the first one. Rewrote both to drop them; tree hash is unchanged, so the diff is exactly what was reviewed. cla/google is green now.

PR #10794, opened 2026-07-20
fix(build): correct explicit_false alias target in define_build_option

### LINKED ISSUE #9676: Bug: define_build_option function has logic error in native.alias
## Summary
The `define_build_option` function in `ynnpack/build_defs.bzl` has a logic error in the `native.alias` definition (around line 33-41).

## Issue
When the `explicit_false` condition is selected, the alias incorrectly points to `":" + explicit_true` instead of `":" + explicit_false`:

```starlark
native.alias(
    name = name,
    actual = select({
        explicit_true: ":" + explicit_true,
        explicit_false: ":" + explicit_true,  # ← BUG: should be explicit_false
        "//conditions:default": ":" + default,
    }),
)

RohanMali2003:
Fixes #9676

This PR corrects a copy-paste typo in the `define_build_option` function within `build_defs.bzl`. 

Previously, the `explicit_false` key in the `native.alias` select dictionary was incorrectly mapped to `":" + explicit_true`. This update corrects the mapping to point to `":" + explicit_false` as intended.

### DIFF
--- a/ynnpack/build_defs.bzl
+++ b/ynnpack/build_defs.bzl
@@ -39,7 +39,7 @@ def define_build_option(name, default_all = [], default_any = []):
         name = name,
         actual = select({
             explicit_true: ":" + explicit_true,
-            explicit_false: ":" + explicit_true,
+            explicit_false: ":" + explicit_false,
             "//conditions:default": ":" + default,
         }),
     )


dsharlet:
I don't think this is wrong, bazel is just weird. See https://github.com/google/XNNPACK/pull/10189 for discussion.

RohanMali2003:
Ah, okay! Thanks for pointing out the discussion in #10189. I didn't realize this was a known Bazel quirk and just assumed it was a copy-paste error based on the issue being open. I'll close the PR.

PR #10851, opened 2026-07-27
Fix bad-free in rewrite_dequant_bmm_at on fixed-capacity scratch subgraphs

wangw-1991:
`rewrite_dequant_bmm_at()` reserves node/value capacity before checking the node type. The reserve can `realloc()` the subgraph's arrays, which is unsafe for the stack-allocated scratch subgraph built by `xnn_subgraph_fuse_unary_quantized_into_lut()` (`unary_nodes`/`unary_values`, capped at `XNN_MAX_UNARY_FUSION_NODES`). When a fusion chain fills all slots, the reserve calls `realloc()` on a stack pointer → bad-free.

Fix: check for `batch_matrix_multiply` before reserving. A scratch subgraph contains only unary elementwise ops, so it never matches and the stray `realloc()` is avoided. Real heap-backed graphs with a BMM node are unaffected (they still reserve upfront).

See chromium issue: https://issues.chromium.org/issues/539052047

### DIFF
--- a/src/subgraph.c
+++ b/src/subgraph.c
@@ -4242,6 +4242,15 @@ static bool is_qs8_to_f32_dequant(const struct xnn_subgraph* subgraph,
 }
 
 static bool rewrite_dequant_bmm_at(xnn_subgraph_t subgraph, uint32_t node_id) {
+  // Only batch-matrix-multiply nodes are rewritten here. Check the node type
+  // *before* reserving any space: reserving may reallocate the subgraph's node
+  // and value arrays, and some callers optimize a fixed-capacity scratch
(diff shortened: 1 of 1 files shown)


qukhan (changes requested):
The change is fine however note that your commit message is "Initial upload.", which is not very descriptive and which is what will be seen in the commit history.

Could you (force) push a commit with an actual description of the change?

wangw-1991:
> The change is fine however note that your commit message is "Initial upload.", which is not very descriptive and which is what will be seen in the commit history.
> 
> Could you (force) push a commit with an actual description of the change?

Done, thanks. Please take a look, again.

qukhan (approved):

PR #10852, opened 2026-07-27
memory-planner: add NULL checks after xnn_allocate_zero_memory

destro4evr-rgb:
## Summary

`xnn_init_value_allocation_tracker` and `xnn_plan_value_allocation_tracker` each call `xnn_allocate_zero_memory` for internal scratch buffers but do not check the returned pointer before using it. On memory-constrained devices any of the three allocations can return NULL, producing an immediate null pointer dereference.

Three affected sites in `src/memory-planner.c`:

- `xnn_init_value_allocation_tracker` (line 151): `tracker->usage` passed directly to `populate_value_lifecycle`, which indexes it unconditionally
- `xnn_plan_value_allocation_tracker` (line 210): `sorted_usage` written to in the immediately following loop
- `xnn_plan_value_allocation_tracker` (line 221): `current_live_mem_blocks` written to inside the planning loop

Both functions are changed to return `enum xnn_status`. `xnn_plan_memory` in `runtime.c` is updated to check both return values and propagate `xnn_status_out_of_memory` to its caller (`xnn_reshape_runtime` / `xnn_setup_runtime_v2`).

## Changes

- `src/xnnpack/memory-planner.h`: update declarations of `xnn_init_value_allocation_tracker` and `xnn_plan_value_allocation_tracker` to return `enum xnn_status`
- `src/memory-planner.c`: add NULL guards and early returns; add `src/xnnpack/log.h` include for `xnn_log_error`
- `src/runtime.c`: check return values of both planning functions in `xnn_plan_memory`

### DIFF
--- a/src/xnnpack/memory-planner.h
+++ b/src/xnnpack/memory-planner.h
@@ -51,7 +51,10 @@ struct xnn_value_allocation_tracker {
 };
 
 // Initialize the memory allocation tracker for xnn_values.
-XNN_INTERNAL void xnn_init_value_allocation_tracker(
+// Returns xnn_status_out_of_memory if the internal usage array cannot be
+// allocated. On failure the tracker must not be used and does not need to be
+// released.
+XNN_INTERNAL enum xnn_status xnn_init_value_allocation_tracker(
     struct xnn_value_allocation_tracker* tracker,
     const struct xnn_runtime* runtime);
 
@@ -85,7 +88,9 @@ XNN_INTERNAL void xnn_mark_tensor_as_reuse(
 
 // Plan the exact the memory allocation for intermediate tensors according to
 // the xnn_value allocation tracker.
-XNN_INTERNAL void xnn_plan_value_allocation_tracker(
+// Returns xnn_status_out_of_memory if an internal scratch buffer cannot be
+// allocated.
+XNN_INTERNAL enum xnn_status xnn_plan_value_allocation_tracker(
     struct xnn_value_allocation_tracker* tracker);
 
 #ifdef __cplusplus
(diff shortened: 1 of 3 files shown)


src/xnnpack/memory-planner.h
```diff
@@ -51,7 +51,10 @@ struct xnn_value_allocation_tracker {
 };
 
 // Initialize the memory allocation tracker for xnn_values.
-XNN_INTERNAL void xnn_init_value_allocation_tracker(
+// Returns xnn_status_out_of_memory if the internal usage array cannot be
```
qukhan, line 55: Can you add a blank comment line between the first-line comment and the rest of the description?

```suggestion
//
// Returns xnn_status_out_of_memory if the internal usage array cannot be
```

src/xnnpack/memory-planner.h
```diff
@@ -85,7 +88,9 @@ XNN_INTERNAL void xnn_mark_tensor_as_reuse(
 
 // Plan the exact the memory allocation for intermediate tensors according to
 // the xnn_value allocation tracker.
-XNN_INTERNAL void xnn_plan_value_allocation_tracker(
+// Returns xnn_status_out_of_memory if an internal scratch buffer cannot be
```
qukhan, line 93:
```suggestion
//
// Returns xnn_status_out_of_memory if an internal scratch buffer cannot be
```

qukhan (changes requested):

src/xnnpack/memory-planner.h
destro4evr-rgb: Applied directly to the branch

src/xnnpack/memory-planner.h
destro4evr-rgb: Applied directly to the branch

destro4evr-rgb:
Hi @qukhan , I've updated the changes in the files directly

qukhan (approved):

PR #10853, opened 2026-07-27
litert/arithmetic: add bounds checks to Gather, Slice, and TopK shape inference

destro4evr-rgb:
## Summary

`litert/tensor/arithmetic.h` contains three shape-inference functions with missing input validation, following the same pattern fixed for ten other functions by PRs #10464 and #10702.

- **Gather**: `axis` and `batch_dims` were used as direct loop bounds and iterator offsets against `std::vector<int>` shape vectors without validating them against the actual tensor rank. An out-of-range `axis` caused heap out-of-bounds reads via `input_info.shape[i]`; an out-of-range `batch_dims` produced an invalid iterator passed to `insert`.

- **Slice**: The size tensor buffer length was not checked against the input tensor rank before the `-1` sentinel loop. When `size_data.size() > input_info.shape.size()` and a `-1` element appeared at index `i >= input rank`, the code read `input_info.shape[i]` past the end of the allocated vector.

- **TopK**: The `k` tensor's buffer pointer was dereferenced without a null check (every comparable function — `Sum`, `ReduceMax`, `Mean`, `ArgMax` — checks `buffer == nullptr` first), and `k_lock.data()[0]` was accessed without verifying the buffer contained at least one element.

## Fix

Added the same `InvalidArgumentError` guard pattern used throughout the rest of the file:
- `Gather`: validate `0 <= axis < input rank` and `0 <= batch_dims <= indices rank` before the loop and insert.
- `Slice`: validate `size_data.size() == input_info.shape.size()` before assigning the output shape.
- `TopK`: check `buffer != nullptr` and `k_lock.size() >= 1` before reading `k_lock.data()[0]`.

### DIFF
--- a/litert/tensor/arithmetic.h
+++ b/litert/tensor/arithmetic.h
@@ -2195,23 +2195,30 @@ Tensor<Mixins...> Gather(Tensor<Mixins...> input, Tensor<Mixins...> indices,
   graph::TensorInformation& output_info = *GetInfo(output.GetRaw());
   output_info.type = input_info.type;
   output_info.shape.clear();
-  if (axis < 0 || axis >= static_cast<int>(input_info.shape.size())) {
+  int resolved_axis = axis;
+  if (resolved_axis < 0) {
+    resolved_axis += static_cast<int>(input_info.shape.size());
+  }
+  if (resolved_axis < 0 ||
+      resolved_axis >= static_cast<int>(input_info.shape.size())) {
     return Tensor<Mixins...>(graph::ErrorTensor(absl::InvalidArgumentError(
         "The Gather axis is out of range.")));
   }
+  op->axis = resolved_axis;
   if (batch_dims < 0 ||
       batch_dims > static_cast<int>(indices_info.shape.size())) {
     return Tensor<Mixins...>(graph::ErrorTensor(absl::InvalidArgumentError(
         "The Gather batch_dims is out of range.")));
   }
   int i = 0;
-  for (; i < axis; ++i) {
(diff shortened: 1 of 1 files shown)


qukhan (approved):

destro4evr-rgb:
Hi @qukhan, just following up - the PR has been approved and all CI is green. Is there anything blocking the merge on your end?

qukhan:
Yes, an internal test is failing. Let me come back to you.

litert/tensor/arithmetic.h
```diff
@@ -2179,6 +2195,15 @@ Tensor<Mixins...> Gather(Tensor<Mixins...> input, Tensor<Mixins...> indices,
   graph::TensorInformation& output_info = *GetInfo(output.GetRaw());
   output_info.type = input_info.type;
   output_info.shape.clear();
+  if (axis < 0 || axis >= static_cast<int>(input_info.shape.size())) {
```
qukhan: The `axis < 0` checks makes one of our internal tests fail. When `axis` is negative, it should be counted from the end.

qukhan (changes requested):

litert/tensor/arithmetic.h
destro4evr-rgb: Hi @qukhan , Fixed - negative `axis` is now normalized to `axis + rank` before the bounds check (standard TF/numpy convention). The resolved value is also stored in `op->axis` so execution sees a non-negative index. Updated in c6edda3.

qukhan (approved):

PR #10874, opened 2026-07-28
Add `f16_dwconv` AVX512FP16 kernels

rahulchaphalkar:
I'm starting to implement a few F16 kernels which lack AVX512FP16 implementations, and `dwconv` seems like a good place to start as its used quite a bit in `mobilenet` benchmarks. This patch implements `9p32c` tile size with `acc` and `acc2` both.


rahulchaphalkar:
Bump

fbarchard (approved):

PR #10910, opened 2026-07-31
use valid_batch_size to free zero buffers in xnn_destroy_operator

aizu-m:
xnn_destroy_operator walks convolution_op->zero_buffers with op->batch_size while valid_batch_size is what tracks that array's length, so a reshape rejected after the qd8 wrapper has already resized it leaves the destructor freeing seven slots past the end of a one-entry array; found while diffing the destructor against the three reshape free loops, which all iterate valid_batch_size.

### DIFF
--- a/test/operators/convolution-nhwc.cc
+++ b/test/operators/convolution-nhwc.cc
@@ -6,10 +6,14 @@
 // This source code is licensed under the BSD-style license found in the
 // LICENSE file in the root directory of this source tree.
 
+#include <cstddef>
 #include <utility>
 #include <vector>
 
 #include <gtest/gtest.h>
+#include "include/xnnpack.h"
+#include "src/xnnpack/allocator.h"
+#include "src/xnnpack/params.h"
 #include "test/operators/convolution-operator-tester.h"
 
 namespace {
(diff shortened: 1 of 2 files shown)


test/operators/convolution-nhwc.cc
```diff
@@ -1096,6 +1100,89 @@ CREATE_CONVOLUTION_TESTS(CONVOLUTION_NHWC_QD8_F16_QC8W, TestNHWCxQD8F16QC8W)
...
+              xnn_reshape_convolution2d_nhwc_qd8_f32_qc8w(
+                  convolution_op, /*batch_size=*/8, /*input_height=*/0,
+                  /*input_width=*/4, &workspace_size, &output_height,
+                  &output_width, /*threadpool=*/nullptr));
+
+    ASSERT_EQ(xnn_status_success, xnn_delete_operator(convolution_op));
+  }
+
+  const size_t leaked = live_aligned_allocations;
+  xnn_params.allocator = saved_allocator;
```
qukhan: If an assert fails, this will not be restored. Use an RAII wrapper to swap the allocator back on function exit.

qukhan (changes requested):

test/operators/convolution-nhwc.cc
aizu-m: Good point. Swapped the manual save/restore for a small RAII guard, so the allocator is put back on scope exit even when an assert bails out early. That also let the leak check become a plain assert at the end instead of the deferred snapshot. Test still passes locally.

test/operators/convolution-nhwc.cc
```diff
@@ -1096,6 +1100,98 @@ CREATE_CONVOLUTION_TESTS(CONVOLUTION_NHWC_QD8_F16_QC8W, TestNHWCxQD8F16QC8W)
 
 CREATE_CONVOLUTION_TESTS(CONVOLUTION_NHWC_QD8_F32_QC8W, TestNHWCxQD8F32QC8W)
 
+// Counts the aligned allocations that are still live, so that a zero buffer
+// the operator forgets to release is observable.
+size_t live_aligned_allocations = 0;
```
qukhan: I think you don't need to create a global variable. You can use the `CountingAllocatorGuard` as the allocator context to hold that info.

I'm thinking of something like this (untested):
```cpp
class CountingAllocatorGuard {
 public:
  CountingAllocatorGuard() : saved_allocator_(xnn_params.allocator) {
    xnn_params.allocator.context = this;
    xnn_params.allocator.aligned_allocate = AlignedAllocate;
    xnn_params.allocator.aligned_deallocate = AlignedDeallocate;
  }
  ~CountingAllocatorGuard() {
    xnn_params.allocator = saved_allocator_;
  }
  // Non-copyable/non-movable because xnn_params.allocator.context holds `this`.
  // You could also store the context in a unique_ptr or shared_ptr to relax this.
  CountingAllocatorGuard(const CountingAllocatorGuard&) = delete;
  CountingAllocatorGuard& operator=(const CountingAllocatorGuard&) = delete;
  size_t live_aligned_allocations() const { return live_aligned_allocations_; }
 private:
  static void* AlignedAllocate(void* context, size_t alignment, size_t size) {
    auto* self = static_cast<CountingAllocatorGuard*>(context);
    void* pointer = self->saved_allocator_.aligned_allocate(
        self->saved_allocator_.context, alignment, size);
    if (pointer != nullptr) {
      self->live_aligned_allocations_++;
    }
    return pointer;
  }
  static void AlignedDeallocate(void* context, void* pointer) {
    auto* self = static_cast<CountingAllocatorGuard*>(context);
    if (pointer != nullptr) {
      self->live_aligned_allocations_--;
    }
    self->saved_allocator_.aligned_deallocate(
        self->saved_allocator_.context, pointer);
  }
  const struct xnn_allocator saved_allocator_;
  size_t live_aligned_allocations_ = 0;
};
```

aizu-m: Done, took your version pretty much as is: the counter now lives in the guard and the hooks reach it through the allocator context, so the global is gone. Forwarding through saved_allocator_ also means the hooks wrap whatever allocator was installed rather than assuming the default one, which let me drop the allocator.h include. Test still passes locally.

test/operators/convolution-nhwc.cc
```diff
@@ -1100,39 +1099,47 @@ CREATE_CONVOLUTION_TESTS(CONVOLUTION_NHWC_QD8_F16_QC8W, TestNHWCxQD8F16QC8W)
...
-    xnn_params.allocator.aligned_deallocate = CountingAlignedDeallocate;
-    live_aligned_allocations = 0;
+    xnn_params.allocator.context = this;
+    xnn_params.allocator.aligned_allocate = AlignedAllocate;
+    xnn_params.allocator.aligned_deallocate = AlignedDeallocate;
   }
   ~CountingAllocatorGuard() { xnn_params.allocator = saved_allocator_; }
 
+  // Non-copyable/non-movable because xnn_params.allocator.context holds
+  // `this`.
```
qukhan, line 1116: Having the `context` hold an implementation class through a `unique_ptr` would solve the non-movable status.

This is just a remark and doesn't need to be acted on.

qukhan (approved):

qukhan:
There are conflicts with the main branch. Could you rebase on top of it?

aizu-m:
Rebased onto master (b3bcd40). The only conflict was the include block in test/operators/convolution-nhwc.cc, where the new overflow tests had added <array>, <cstdint> and <limits> next to my <cstddef>. Kept both sets, nothing else moved, and the three commits are otherwise unchanged. Rebuilt convolution-nhwc-test and ran the *QD8_F32_QC8W* set plus the two new overflow tests, all 85 pass. Left the guard non-movable for now, per your note that the unique_ptr version was optional.

PR #10934, opened 2026-08-05
Fix integer overflow in qc2w/qc4w requantization scale allocation

prasanna8585:
setup_scale_params_qx8_qcyw computed the requantization-scale buffer size as an unguarded context-> output_channels * sizeof(float), where output_channels is a direct parameter of the public API xnn_create_fully_connected_nc_qs8_qc2w (and its qc4w/qd8/qdu8 siblings sharing this scale-params setup function). A caller supplying a large output_channels value overflows size_t, causing the allocation to wrap to a small value; the subsequent per-channel scale computation loop then reads and writes using the real, unwrapped output_channels.\n\nConfirmed via ASan (heap-buffer-overflow read and write) with output_channels = (1 << 62) + 16, which wraps the allocation to exactly 64 bytes.\n\nApplies xnn_safe_mul to the computation, consistent with the pattern already used for the main packed-weights allocation in this same file (line 264) and in convolution-nhwc.c / deconvolution-nhwc.c.

### DIFF
--- a/src/operators/fully-connected-nc.c
+++ b/src/operators/fully-connected-nc.c
@@ -952,11 +952,19 @@ static enum xnn_status setup_scale_params_qs8(const struct fc_variant* variant,
 
 static enum xnn_status setup_scale_params_qx8_qcyw(const struct fc_variant* variant,
                                               struct fc_context* context) {
-  context->requantization_scale.f32 = xnn_allocate_simd_memory(context->output_channels * sizeof(float));
+  size_t requantization_scale_bytes = 0;
+  if (!xnn_safe_mul(context->output_channels, sizeof(float), &requantization_scale_bytes)) {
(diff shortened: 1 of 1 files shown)


prasanna8585:
Hi Team,

is there any update require from my side on this ?

Aelphy:
Hi, we are waiting for internal approval. Nothing is required on your side.

prasanna8585:
Hi @Aelphy ,

Just following up to check if this is ready to be merged.

Aelphy:
Hm, could you please try rebase and push?

prasanna8585:
Done rebased onto current master and pushed. Let me know if anything else needs adjusting.

prasanna8585:
Hi Team,
Let me know if anything else needs adjusting to merge this PR.

PR #10928, opened 2026-08-04
convolution/deconvolution: fix NULL deref after xnn_reallocate_memory for zero_buffers

destro4evr-rgb:
Three reshape functions for quantized NHWC convolution and deconvolution store the result of `xnn_reallocate_memory` directly into `zero_buffers` and then immediately dereference it - a NULL dereference (SIGSEGV) if `realloc` fails under memory pressure.

**Affected functions:**
- `reshape_convolution2d_nhwc_qx8_f16_qc8w` (`src/operators/convolution-nhwc.c`)
- `reshape_convolution2d_nhwc_qx8_f32_qc8w` (`src/operators/convolution-nhwc.c`)
- `reshape_deconvolution2d_nhwc_qx8_f32_qc8w` (`src/operators/deconvolution-nhwc.c`)

**Before:**
```c
convolution_op->convolution_op->zero_buffers =
    xnn_reallocate_memory(convolution_op->convolution_op->zero_buffers,
                          batch_size * sizeof(void*));
convolution_op->convolution_op->zero_buffers[0] =   // NULL deref if realloc failed
    convolution_op->zero_buffer;
```

**After:**
```c
void** new_zero_buffers = xnn_reallocate_memory(
    convolution_op->convolution_op->zero_buffers,
    batch_size * sizeof(void*));
if (new_zero_buffers == NULL) {
  xnn_log_error("failed to reallocate %zu bytes for zero_buffers",
                batch_size * sizeof(void*));
  return xnn_status_out_of_memory;
}
convolution_op->convolution_op->zero_buffers = new_zero_buffers;
convolution_op->convolution_op->zero_buffers[0] = convolution_op->zero_buffer;
```

The fix saves to a temporary, checks for NULL, and propagates `xnn_status_out_of_memory` - consistent with the guard pattern used for other allocations in the same files.

PR #10941, opened 2026-08-07
convolution/deconvolution: fix NULL deref in per-batch zero_buffers allocation loop

destro4evr-rgb:
`xnn_allocate_simd_memory` is a `posix_memalign` / `_mm_malloc` wrapper that returns `NULL` on OOM. In three reshape functions the return value was not checked, so `NULL` was silently stored in `zero_buffers[i]`. At the next inference call `xnn_compute_dq_zero_buffer_igemm` passes `zero_buffers[batch_index]` directly to `memset`, producing a write to address 0x0 and a reliable SIGSEGV.

**Affected functions**
- `reshape_convolution2d_nhwc_qx8_f16_qc8w` (`src/operators/convolution-nhwc.c`)
- `reshape_convolution2d_nhwc_qx8_f32_qc8w` (`src/operators/convolution-nhwc.c`)
- `reshape_deconvolution2d_nhwc_qx8_f32_qc8w` (`src/operators/deconvolution-nhwc.c`)

**Fix**: add a `NULL` guard after `xnn_allocate_simd_memory` inside each per-batch loop and return `xnn_status_out_of_memory` immediately so the error propagates to the caller before inference begins.

**Relation to #10928**: PR #10928 fixes the `xnn_reallocate_memory` NULL deref when resizing the `zero_buffers` pointer array itself. This PR fixes the separate NULL deref inside the subsequent per-batch element allocation loop, which is not addressed by #10928.

### DIFF
--- a/src/operators/convolution-nhwc.c
+++ b/src/operators/convolution-nhwc.c
@@ -3402,6 +3402,12 @@ enum xnn_status reshape_convolution2d_nhwc_qx8_f16_qc8w(
     for (size_t i = 1; i < batch_size; ++i) {
       convolution_op->convolution_op->zero_buffers[i] =
           xnn_allocate_simd_memory(convolution_op->convolution_op->zero_size);
+      if (convolution_op->convolution_op->zero_buffers[i] == NULL) {
+        xnn_log_error(
+            "failed to allocate %zu bytes for zero buffer for batch %zu",
+            convolution_op->convolution_op->zero_size, i);
+        return xnn_status_out_of_memory;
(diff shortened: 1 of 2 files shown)


destro4evr-rgb:
Hi @qukhan, just checking in on this one - it's been sitting for a couple of weeks with all checks have passed. Let me know if any changes are required or if there's anything blocking it on your end.

qukhan:
I was out for some time. I've retriggered the import, I'll let you know if there are some sync issues.

PR #7941, opened 2025-03-04
Update KleidiAI to use GitHub

felix-johnny:
Update to release v1.4.0 and use
GitHub mirror repository

PR #10943, opened 2026-08-07
Add KleidiAI Advanced SIMD PF16 6x32 GEMM kernel

Colm-in-Arm:
### Summary
This change adds the KleidiAI Advanced SIMD (Arm® Neon™) PF16 6×32 GEMM kernels and selects it for packed FP16 GEMM when SME or SME2 kernels are not available. Two kernels are added. A general advanced SIMD and one targeting Arm(R) Cortex(TM)-A55.

The existing conventional PF16 6×16 kernels remain available for operations that cannot use the packed PF16 path, including convolution/IGEMM and cases where packed-LHS optimization is disabled.

This change also updates KleidiAI to version 1.28.

### Stack / Dependencies
The 6×32 kernel exists in KleidiAI 1.25, but this integration also requires the NXK RHS packer, which was part of 1.26.

DEPS now references the official GitHub repository using the v1.28.0 tag. Bazel and CMake use the corresponding immutable archive commit and SHA-256. This is in line with an earlier PR: https://github.com/google/XNNPACK/pull/7941

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -527,6 +527,9 @@ if (current_cpu == "arm64") {
       ":microkernel_defs",
       ":microkernel_headers",
     ]
+    if (xnnpack_enable_arm_kleidiai) {
+      deps += [ "//third_party/kleidiai" ]
+    }
     sources = ALL_NEONFP16_MICROKERNEL_SRCS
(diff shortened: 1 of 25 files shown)


Aelphy:
Please resolve conflicts

qukhan:
Hi, our internal tooling report a merge conflict. Could you please rebase onto main?

PR #10947, opened 2026-08-08
packing: fix NULL deref in KleidiAI SME bias-substitute allocation

destro4evr-rgb:
`calloc` returns NULL on OOM. Three KleidiAI SME weight-packing functions allocate a zero-initialized bias substitute when the caller passes NULL bias (a standard configuration for bias-free FC layers), but none check the `calloc` return value. On failure the NULL pointer is passed directly to `kai_run_*` as the bias argument - no intervening allocation, no early-return path - and the kernel dereferences it → SIGSEGV.

XNNPACK's own comment at each site documents why the substitute is needed:

```
// Some packing kernels assume that the bias is non-null. Allocate a zero
// initialized array as a workaround if bias is null.
```

**Affected functions:**
- `xnn_pack_kai_qs8_qc8w_weights_and_biases_sme` (src/reference/packing.cc ~line 2665)
- `xnn_pack_kai_f16_weights_and_biases` (src/reference/packing.cc ~line 2821)
- `xnn_pack_kai_f32_weights_and_biases` (src/reference/packing.cc ~line 2901)

**Fix:** add a NULL guard after each `calloc`. On OOM, log the error and return early before the kernel call.

Note: this is independent of #10942, which guards `tmp_data` transpose buffer allocations in a different set of functions.

### DIFF
--- a/src/reference/packing.cc
+++ b/src/reference/packing.cc
@@ -2675,6 +2675,7 @@ void xnn_pack_kai_qs8_qc8w_weights_and_biases_sme(
       xnn_log_error(
           "failed to allocate %zu bytes for KleidiAI SME bias substitute buffer",
           output_channels * sizeof(int32_t));
+      assert(false);
       return;
     }
     free_accumulator_init = true;
@@ -2838,6 +2839,7 @@ void xnn_pack_kai_f16_weights_and_biases(
       xnn_log_error(
           "failed to allocate %zu bytes for KleidiAI SME bias substitute buffer",
           output_channels * sizeof(float));
+      assert(false);
       return;
     }
     free_accumulator_init = true;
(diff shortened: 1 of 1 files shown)


Aelphy:
If return is happening prematurely due to OOM, execution would just continue with unfinished packing, may be put assert(false) before return for simpler debugging. Same for #10942

destro4evr-rgb:
Hi @Aelphy , added assert(false) before each early return here and in #10942.

destro4evr-rgb:
Hi @Aelphy the copybara check is still failing on this one - is there anything needed on my end to resolve it?

Aelphy:
Hm, interesting. Can you try sync to master?

destro4evr-rgb:
Hi @Aelphy already synced to master, still failing on copybara though.

Aelphy:
Ok, found the issue, resolving.

Aelphy (approved):

destro4evr-rgb:
Hi @qukhan , this PR needs an approval to proceed, could you please check it once?

Aelphy (approved):

qukhan (approved):

destro4evr-rgb:
Hi @Aelphy @qukhan, I've rebased with master to resolve the CI failure.

PR #10960, opened 2026-08-11
Repair shared workspace users after setup errors

rootd4ddy:
## Summary

- Preserve the first setup failure while finishing shared-workspace pointer and context repair.
- Add a regression test for a sibling setup failure during workspace growth.

## Why

The shared arena is replaced before sibling setup callbacks run. Returning on the first callback error leaves later runtimes pointing into released storage. Complete the repair pass before returning the first error.

## Test

- `workspace-test`: 9/9 passed.
- The original ASan reproducer exits cleanly with this change.


rootd4ddy:
@googlebot rescan

PR #10985, opened 2026-08-14
Fix integer overflow in convolution conversion buffer sizing

anagnorisis2peripeteia:
## Summary

Several public convolution and deconvolution constructors convert FP16 kernels
or biases into temporary buffers. They multiply caller-controlled dimensions to
compute the allocation size, then use the unwrapped element count in the
conversion loop.

The unchecked products can wrap `size_t`. The allocation can then be smaller
than the conversion loop, which causes a heap-buffer-overflow.

Affected paths:

- `xnn_create_convolution2d_nhwc_f16`
- `xnn_create_convolution2d_nhwc_pf16`
- `xnn_create_convolution2d_nhwc_f32_f16`
- `xnn_create_convolution2d_nhwc_pf32_f16`
- `xnn_create_deconvolution2d_nhwc_f32_f16`

## Fix

Use `xnn_safe_mul` for each element-count and byte-size product. Return
`xnn_status_invalid_parameter` before allocation or conversion when a product
does not fit in `size_t`. Release the temporary kernel buffer when a later bias
size check fails.

## Tests

- Added regression tests for kernel-conversion and FP32-static-bias overflow in
  the convolution and deconvolution operator tests.
- Built the two test binaries with CMake and Ninja.
- Ran all 20 convolution and deconvolution test shards with CTest. All passed.
- The same overflow inputs reproduce heap-buffer-overflow under ASan on the
  base revision and return `xnn_status_invalid_parameter` on this branch.

### DIFF
--- a/src/operators/convolution-nhwc.c
+++ b/src/operators/convolution-nhwc.c
@@ -2306,8 +2306,16 @@ enum xnn_status xnn_create_convolution2d_nhwc_f16(
   const void* bias_to_use = bias;
   void* allocated_bias = NULL;
   if (bias != NULL && (flags & XNN_FLAG_FP32_STATIC_BIASES)) {
-    const size_t bias_size = groups * group_output_channels;
-    allocated_bias = xnn_allocate_memory(bias_size * sizeof(xnn_float16));
+    size_t bias_size;
+    size_t bias_bytes;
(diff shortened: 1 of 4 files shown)


qukhan (approved):

anagnorisis2peripeteia:
@qukhan some CI failure?

PR #10990, opened 2026-08-14
[Security] Fix heap buffer overflow via integer overflow in depth-to-space reshape on 32-bit targets

rishuranjanofficial:
On 32-bit targets, size_t is 32-bit. Several multiplications in reshape_depth_to_space_nhwc() in src/operators/transpose-nd.c were unguarded and silently overflow with large input dimensions. This produces incorrect buffer strides passed to the transpose operation, causing writes beyond the allocated heap buffer.

A crafted model with large spatial dimensions can trigger this on any 32-bit deployment of XNNPACK (Android ARMv7, x86 32-bit, RISC-V RV32).

Fixed by replacing bare multiplications with xnn_safe_mul() calls, consistent with overflow handling already present in the codebase. 
Returns xnn_status_invalid_parameter on overflow instead of proceeding with corrupt strides.

### DIFF
--- a/src/operators/transpose-nd.c
+++ b/src/operators/transpose-nd.c
@@ -1157,9 +1157,9 @@ static enum xnn_status reshape_depth_to_space_nhwc(
     return xnn_status_invalid_parameter;
   }
 
-  size_t output_stride_0, output_stride_1, temp;
-  if (!xnn_safe_mul((size_t)block_size, input_width, &temp) ||
-      !xnn_safe_mul(temp, block_output_pixel_stride, &output_stride_0)) {
+  size_t output_stride_0, output_stride_1;
+  if (!xnn_safe_mul((size_t)block_size, input_width, &output_stride_0) ||
+      !xnn_safe_mul(output_stride_0, block_output_pixel_stride, &output_stride_0)) {
     xnn_log_error("failed to reshape %s operator: overflow in output_stride[0]",
         xnn_operator_type_to_string(expected_operator_type));
     return xnn_status_invalid_parameter;
--- a/src/operators/transpose-nd.c
+++ b/src/operators/transpose-nd.c
@@ -1136,31 +1136,63 @@ static enum xnn_status reshape_depth_to_space_nhwc(
     return xnn_status_success;
   }
 
-  const size_t block_output_pixel_stride = block_size * output_channels;
+  size_t block_output_pixel_stride;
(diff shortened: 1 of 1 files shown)


src/operators/transpose-nd.c
```diff
@@ -1136,31 +1136,63 @@ static enum xnn_status reshape_depth_to_space_nhwc(
...
+
+  size_t input_row_stride;
+  if (!xnn_safe_mul(input_width, input_channels, &input_row_stride)) {
+    xnn_log_error("failed to reshape %s operator: overflow in input_width * input_channels",
+        xnn_operator_type_to_string(expected_operator_type));
+    return xnn_status_invalid_parameter;
+  }
+
+  size_t output_stride_0, output_stride_1, temp;
+  if (!xnn_safe_mul((size_t)block_size, input_width, &temp) ||
```
qukhan: You don't need `temp`, you can directly accumulate into `output_stride_0`.

qukhan (changes requested):

src/operators/transpose-nd.c
rishuranjanofficial: Done - accumulated directly into output_stride_0 and removed temp. Thanks!

qukhan (approved):

rishuranjanofficial:
Addressed the review: accumulated both multiplications directly into output_stride_0 and removed temp.

Note on the failing check: gn-linux-arm is a CI environment issue, not this change. Every suite aborts before running with qemu-arm64: Could not open '/lib/ld-linux-aarch64.so.1': No such file or directory - the aarch64 loader is missing on the runner, so qemu can't exec any arm64 test binary. All CMake/Bazel builds including cmake-linux-aarch64, cmake-linux-armhf, and bazel-linux-aarch64-* are green. This PR is a behavior-identical variable rename in reshape_depth_to_space_nhwc.

qukhan:
There was an issue with the gn linux arm builds last week. You'll probably have to rebase onto the main branch to get it to pass.

rishuranjanofficial:
Rebased onto master. gn-linux-arm passes now, but cmake-wasm fails on an unrelated f16 microkernel - f16-vapproxgelu-wasmrelaxedsimd-… errors with always_inline function 'wasm_f16x8_pmin' requires target feature 'fp16'. That's in the WASM f16 relaxed-SIMD kernels, not this depth-to-space change, and looks like a pre-existing break on master. 

Happy to rebase again once it's sorted, or to merge past it if you prefer.

qukhan (approved):

PR #10991, opened 2026-08-14
guard NCHW FP16 conversion buffer sizes

anagnorisis2peripeteia:
## Summary

Guard the temporary FP32 buffers used by the NCHW FP16 convolution constructor.

## Problem

The constructor multiplied the group, channel, and kernel dimensions without checking for `size_t` overflow. It then allocated the wrapped byte count but converted the full logical element count. The same issue affected the temporary bias buffer. A caller that supplies dimensions near the `size_t` limit could therefore trigger a heap buffer overflow during weight conversion.

## Fix

Use `xnn_safe_mul` for each kernel and bias element-count product and for each byte-count conversion. Return `xnn_status_invalid_parameter` before allocation when a count is not representable. Use the checked counts for both allocation and conversion loops.

## Tests

- Added focused NCHW constructor tests for overflow at every kernel product stage and for both bias overflow stages.
- Added a normal-size bias conversion test.
- Ran the affected `CONVOLUTION_NCHW.*` tests under AddressSanitizer and UndefinedBehaviorSanitizer. The base revision reproduces the heap-buffer overflow; this change returns `xnn_status_invalid_parameter`.
- Built the test target and ran CTest. 495 of 496 tests passed; the only failure was the existing Apple scalar tolerance failure in `f32-vgelu-test`.

The paired constructor check produced these results:

```text
base: AddressSanitizer: heap-buffer-overflow in xnn_create_convolution2d_nchw_f32_f16
head: overflow status=2 op=0x0; exit_code=0
```

Review history: https://gist.github.com/8d068e7e6647a814d5898654c55b8f52

### DIFF
--- a/src/operators/convolution-nchw.c
+++ b/src/operators/convolution-nchw.c
@@ -1238,32 +1238,12 @@ enum xnn_status xnn_create_convolution2d_nchw_f32_f16(
   const void* kernel_for_cache_key = kernel;
   size_t num_kernel_entries;
   size_t kernel_bytes;
-  if (!xnn_safe_mul(groups, group_input_channels, &num_kernel_entries)) {
-    xnn_log_error(
-        "failed to create %s operator: kernel size overflows size_t",
-        xnn_operator_type_to_string(xnn_operator_type_convolution_nchw_f32));
-    return xnn_status_invalid_parameter;
-  }
-  if (!xnn_safe_mul(num_kernel_entries, group_output_channels,
-                    &num_kernel_entries)) {
-    xnn_log_error(
-        "failed to create %s operator: kernel size overflows size_t",
-        xnn_operator_type_to_string(xnn_operator_type_convolution_nchw_f32));
-    return xnn_status_invalid_parameter;
-  }
-  if (!xnn_safe_mul(num_kernel_entries, kernel_width, &num_kernel_entries)) {
-    xnn_log_error(
-        "failed to create %s operator: kernel size overflows size_t",
-        xnn_operator_type_to_string(xnn_operator_type_convolution_nchw_f32));
-    return xnn_status_invalid_parameter;
-  }
-  if (!xnn_safe_mul(num_kernel_entries, kernel_height, &num_kernel_entries)) {
-    xnn_log_error(
-        "failed to create %s operator: kernel size overflows size_t",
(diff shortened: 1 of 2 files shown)


src/operators/convolution-nchw.c
```diff
@@ -1236,11 +1236,40 @@ enum xnn_status xnn_create_convolution2d_nchw_f32_f16(
...
+        xnn_operator_type_to_string(xnn_operator_type_convolution_nchw_f32));
+    return xnn_status_invalid_parameter;
+  }
+  if (!xnn_safe_mul(num_kernel_entries, sizeof(float), &kernel_bytes)) {
+    xnn_log_error(
+        "failed to create %s operator: kernel size overflows size_t",
+        xnn_operator_type_to_string(xnn_operator_type_convolution_nchw_f32));
+    return xnn_status_invalid_parameter;
+  }
+  float* fp32_kernel_buffer = (float*)xnn_allocate_memory(kernel_bytes);
```
qukhan: All of these are the same multiplication and accumulate in the same variable, you should chain them in one `if`.

```cpp
if (!xnn_safe_mul(groups, group_input_channels, &num_kernel_entries) ||
    !xnn_safe_mul(num_kernel_entries, group_output_channels,
                    &num_kernel_entries) ||
   ...)
```

qukhan (changes requested):

src/operators/convolution-nchw.c
anagnorisis2peripeteia: Implemented in commit 8cec1a29a5. I chained the kernel and bias checked multiplications into short-circuit conditions, as suggested. The focused convolution-nchw-test passes.

qukhan (approved):

anagnorisis2peripeteia:
Fixed in 35aaf8ae70: `CONVOLUTION_NCHW.bias_conversion` now skips when the NCHW F32 configuration is unsupported. Could you please re-approve the PR to start the CI matrix on the updated head?

qukhan (approved):

anagnorisis2peripeteia:
@qukhan srry but this copybara thing perplexes me

qukhan:
That's the mechanism that integrates PRs into our codebase. It imports the PR into an internal review tool and then this is re-exported as an internal PR so that both sides stay in sync, which leads to a new PR being created.

PR #11007, opened 2026-08-18
fix: check bias allocation size for overflow in xnn_create_convolution2d_nhwc_pqs8_qs8_qs8

prasanna8585:
`groups * group_output_channels * sizeof(float)` was computed directly and passed to `xnn_allocate_simd_memory` with no overflow check. `groups` and `group_output_channels` are operator-creation parameters that typically originate from a loaded model file; a crafted model declaring values whose product, multiplied by `sizeof(float)`, wraps around `size_t` causes an undersized (or zero-byte) allocation, while the subsequent write loop iterates using the un-overflowed, correctly-computed channel count — a heap buffer overflow.

This is the same defect class already fixed in the f16/pf16/f32_f16/pf32_f16 convolution and deconvolution variants in this codebase (google/XNNPACK@5ecde1f9, "fix/conversion-buffer-size-overflow"), which switched to `xnn_safe_mul` for the identical reason. `xnn_create_convolution2d_nhwc_pqs8_qs8_qs8` has its own separate, inline bias-broadcast allocation that wasn't covered by that fix. This applies the same `xnn_safe_mul`-based check here.

Verified with a standalone AddressSanitizer reproduction of the exact allocation-then-write logic: the pre-fix code produces a heap-buffer-overflow (WRITE past a zero/near-zero-byte allocation) when `groups * group_output_channels` overflows `size_t` after multiplying by `sizeof(float)`; the fixed code correctly rejects operator creation with `xnn_status_invalid_parameter` instead.

### DIFF
--- a/ynnpack/kernels/dequantize_dot/generic.h
+++ b/ynnpack/kernels/dequantize_dot/generic.h
@@ -6,7 +6,6 @@
 #ifndef XNNPACK_YNNPACK_KERNELS_DEQUANTIZE_DOT_GENERIC_H_
 #define XNNPACK_YNNPACK_KERNELS_DEQUANTIZE_DOT_GENERIC_H_
 
-#include <algorithm>
 #include <cassert>
 #include <cstddef>
 #include <cstdint>
@@ -24,7 +23,7 @@ template <size_t N, typename T>
 struct broadcast_to_load {
   broadcast_to_load(const T*& ptr, size_t stride) {
     if (stride == 0) {
-      std::fill_n(data, N, *ptr);
+      simd::store(data, simd::broadcast<N>(*ptr));
       ptr = data;
     } else {
       assert(stride == sizeof(T));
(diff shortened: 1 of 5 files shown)


qukhan (approved):

src/operators/convolution-nhwc.c
```diff
@@ -2245,17 +2245,27 @@ enum xnn_status xnn_create_convolution2d_nhwc_pqs8_qs8_qs8(
...
     int8_t output_max, uint32_t flags, xnn_weights_cache_t weights_cache,
     xnn_operator_t* convolution_op_out) {
+  size_t bias_size;
+  size_t bias_bytes;
+  if (!xnn_safe_mul(groups, group_output_channels, &bias_size) ||
+      !xnn_safe_mul(bias_size, sizeof(float), &bias_bytes)) {
+    xnn_log_error(
+        "failed to create %s operator: bias size overflows size_t",
+        xnn_operator_type_to_string(
+            xnn_operator_type_convolution_nhwc_pqs8_qs8_qs8));
```
Aelphy: xnn_operator_type_convolution_nhwc_pqs8_qs8_qs8 -> xnn_operator_type_convolution_nhwc_pqs8_qs8_qc8

Aelphy: I am wrong. add missing xnn_operator_type_convolution_nhwc_pqs8_qs8_qs8

prasanna8585:
@qukhan is there any update required from my side to merge this PR?

qukhan:
There's seems to be a merge conflict with the main branch. Can you try rebasing?

PR #11021, opened 2026-08-19
Check xnn_allocate_memory return value in xnn_create_threadpool_v2

destro4evr-rgb:
`xnn_create_threadpool_v2` in `src/runtime.c` calls `xnn_allocate_memory` but does not check the return value before immediately writing through the pointer. 
If the allocation fails, `*threadpool_out` is NULL and the next line (`(*threadpool_out)->scheduler = scheduler`) is a guaranteed NULL dereference. The function also unconditionally returns `xnn_status_success`, so the caller has no way to detect the failure.

Add a NULL check after the allocation and return `xnn_status_out_of_memory` on failure, consistent with how every other allocation in the runtime is handled.


qukhan (approved):

Aelphy:
our internal tests show merge conflict. Could you please rebase with master?

destro4evr-rgb:
Hi @Aelphy, I've rebased with master

PR #11022, opened 2026-08-19
fix dwconv-config

rahulchaphalkar:
- Added back a check for fp16 supported in dwconv-config
- Refactored helper function for scalar init into inline
- removed scalar fallback from non x86 archs

### DIFF
--- a/src/configs/dwconv-config.c
+++ b/src/configs/dwconv-config.c
@@ -33,28 +33,6 @@ XNN_INIT_ONCE_GUARD(qu8_dwconv);
   (xnn_dwconv_ukernel_fn) ukernel;       \
   xnn_log_info("Using dwconv microkernel '%s'.", #ukernel);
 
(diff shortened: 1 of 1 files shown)


fbarchard (approved):

src/configs/dwconv-config.c
```diff
@@ -990,7 +978,7 @@ static void init_qu8_dwconv_config(void) {
 
 const struct xnn_dwconv_config* xnn_init_f16_dwconv_config() {
   const struct xnn_hardware_config* hardware_config = xnn_init_hardware_config();
-  if (hardware_config == NULL) {
+  if (hardware_config == NULL || !xnn_is_f16_compatible_config(hardware_config)) {
```
fbarchard, line 981: { should be on previous line

fbarchard (changes requested):

fbarchard (approved):

PR #11043, opened 2026-08-21
   Fix unguarded integer overflow in reshape_reduce_nd() workspace sizing

Benjamin-Asare-Agyapong:
### What

`reshape_reduce_nd()` in `src/operators/reduce-nd.c` computes the workspace
size for a reduce operator (and, in one branch, a zero-padding buffer size)
by multiplying caller-supplied tensor dimensions with no overflow check.

If the product overflows `size_t`, `*workspace_size` wraps to a value much
smaller than what's actually required. A caller allocates a buffer of that
(wrong) size, and the reduce kernel then writes based on the true,
non-overflowed per-dimension bounds — writing past the undersized buffer.

### Why this pattern specifically

13 of the 14 operator files in `src/operators/` that compute a buffer size
by multiplying tensor dimensions already guard it with the library's own
`xnn_safe_mul()` helper (`src/xnnpack/math.h`) — e.g. `argmax-pooling-nhwc.c`,
`average-pooling-nhwc.c`, `convolution-nhwc.c`, `fully-connected-nc.c`.
`reduce-nd.c` was the sole exception.

### Reachability

This is reached via the public, documented low-level operator API
(`xnn_create_reduce_nd` / `xnn_reshape_reduce_nd` / `xnn_setup_reduce_nd`),
which any embedding application can call directly with its own shape
metadata. The subgraph/model-parsing path (`xnn_define_tensor_value()` in
`src/tensor.c`) independently bounds the full tensor's element count before
any reshape happens, so callers going through that front door aren't
affected — this is specific to direct use of the low-level API with shape
data that doesn't originate from a `xnn_subgraph`.

### The fix

Applies the same `xnn_safe_mul()` pattern used everywhere else in the
codebase to both unguarded multiplications in `reshape_reduce_nd()`, plus an
explicit check that the subsequent left-shift by
`log2_accumulator_element_size` / `log2_data_element_size` doesn't itself
overflow. On overflow, returns `xnn_status_unsupported_parameter` and logs
an error, matching the convention already used by e.g.
`argmax-pooling-nhwc.c` for the identical failure mode.

No behavior change for any shape where the multiplication doesn't overflow.

### DIFF
--- a/src/operators/reduce-nd.c
+++ b/src/operators/reduce-nd.c
@@ -240,7 +240,17 @@ static XNN_NO_SANITIZE_FUNCTION enum xnn_status reshape_reduce_nd(
   size_t num_reduction_elements;
   if (normalized_reduction_axes[num_reduction_axes - 1] == num_input_dims - 1) {
     if (workspace_size != NULL) {
-      const size_t num_output_elements = normalized_input_shape[0] * normalized_input_shape[2] * normalized_input_shape[4];
+      // Fix: guard against size_t overflow when multiplying tensor dimensions
+      // together to compute the workspace size (CWE-190 -> CWE-131 -> CWE-787).
+      size_t num_output_elements;
+      if (!xnn_safe_mul(normalized_input_shape[0], normalized_input_shape[2], &num_output_elements) ||
+          !xnn_safe_mul(num_output_elements, normalized_input_shape[4], &num_output_elements) ||
+          num_output_elements > (SIZE_MAX >> log2_accumulator_element_size) - XNN_EXTRA_BYTES) {
+        xnn_log_error(
+            "failed to reshape %s operator: workspace size computation overflowed",
+            xnn_operator_type_to_string_v2(reduce_op));
+        return xnn_status_unsupported_parameter;
+      }
       *workspace_size = (num_output_elements << log2_accumulator_element_size) + XNN_EXTRA_BYTES;
     }
     num_reduction_elements = normalized_input_shape[1] * normalized_input_shape[3] * normalized_input_shape[5];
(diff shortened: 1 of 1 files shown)


Benjamin-Asare-Agyapong:
This fixes a potential security issue: `reshape_reduce_nd()` computes a
workspace buffer size by multiplying tensor dimensions with no overflow
check, unlike every other operator in `src/operators/` that does the same
kind of size calculation (all 13 of which already use `xnn_safe_mul()`). If
the product overflows `size_t`, the buffer allocated ends up much smaller
than what the reduce kernel actually writes.

I've also reported this through Google's OSS VRP portal for tracking. Happy
to share more detail on request, but I'll keep specifics off this public
thread pending review — glad to discuss privately if useful for triage.

Would appreciate a look when you have a chance — thanks!

src/operators/reduce-nd.c
```diff
@@ -240,7 +240,17 @@ static XNN_NO_SANITIZE_FUNCTION enum xnn_status reshape_reduce_nd(
   size_t num_reduction_elements;
   if (normalized_reduction_axes[num_reduction_axes - 1] == num_input_dims - 1) {
     if (workspace_size != NULL) {
-      const size_t num_output_elements = normalized_input_shape[0] * normalized_input_shape[2] * normalized_input_shape[4];
+      // Fix: guard against size_t overflow when multiplying tensor dimensions
```
qukhan, line 243: Please, re-read your agent's output before sending PRs. This comment and the ones below have obviously no use to the code comprehension and are specific to the change that you are making.

qukhan (changes requested):

PR #11084, opened 2026-08-27
Wasm F16 avgpool and scalar fallback

rahulchaphalkar:


fbarchard (approved):
overall good.  top suggestion is we use f16c on most platforms for native and avx512skx would help, using f16-f32acc approach.  Done with portable template it shouldnt need new source, just generator
wasm version is only 13% of native on high end cpus, but really needs d8 to support f16 arithmetics, like arm, to close the gap.
we eventually need followup to detect fp16 support and choose the right kernel - f16, f16-f32acc for wasmrelaxedsimd

PR #11092, opened 2026-08-28
litert/xnnpack: reject negative tensor dimensions in DefineValue()

destro4evr-rgb:
## Problem

`DefineValue()` in `litert/tensor/backends/xnnpack/conversion.cc` converts the tensor shape from `std::vector<int32_t>` to `std::vector<size_t>` with an implicit cast:

```cpp
std::vector<size_t> dims(info.shape.begin(), info.shape.end());
```

A crafted TFLite model that declares a tensor dimension as `-1` (valid as an `int32_t`, common for dynamic shapes) causes the implicit conversion to produce `~1.8e19` (`SIZE_MAX`) for that element. The ballooned `size_t` value is then passed as the dimension count to `xnn_validate_channelwise_quantized_tensor()`, which loops over `dims[channel_dim]` entries in the scale/zero-point arrays. Iterating `SIZE_MAX` times reads arbitrarily far past the end of those arrays → SIGSEGV at model-load time, before any inference runs.

## Fix

Add an explicit negative-dimension check before inserting each element into `dims`, and return a descriptive `InvalidArgumentError` instead of crashing.

```cpp
std::vector<size_t> dims;
dims.reserve(info.shape.size());
for (int32_t d : info.shape) {
  if (d < 0) {
    return absl::InvalidArgumentError(absl::StrFormat(
        "%s: tensor has negative dimension %d; dynamic shapes are not "
        "supported in the XNNPACK backend",
        info.name, d));
  }
  dims.push_back(static_cast<size_t>(d));
}
```

This makes the error explicit and prevents the out-of-bounds read in `xnn_validate_channelwise_quantized_tensor()` (and analogously in `xnn_validate_blockwise_quantized_tensor()`).


litert/tensor/backends/xnnpack/conversion.cc
```diff
@@ -428,7 +428,17 @@ absl::StatusOr<uint32_t> XnnpackBuildContext::DefineValue(
...
+  dims.reserve(info.shape.size());
+  for (int32_t d : info.shape) {
+    if (d < 0) {
+      return absl::InvalidArgumentError(absl::StrFormat(
+          "%s: tensor has negative dimension %d; dynamic shapes are not "
+          "supported in the XNNPACK backend",
+          info.name, d));
+    }
+    dims.push_back(static_cast<size_t>(d));
+  }
```
qukhan: I'd rather you check the data first and then keep the vector creation as a range constructor call.

```suggestion
  if (absl::c_any_of(info.shape, [](int d){return d < 0;})) {
    return absl::InvalidArgumentError(absl::StrFormat(
             "%s: tensor has negative dimension; dynamic shapes are not "
             "supported in the XNNPACK backend", info.name));
  }
  std::vector<size_t> dims(info.shape.begin(), info.shape.end());
```

qukhan (changes requested):

litert/tensor/backends/xnnpack/conversion.cc
destro4evr-rgb: Hi @qukhan , changes has been made as requested.

destro4evr-rgb:
Hi @qukhan, changes are done as requested. Could you re-approve? Also feedback/copybara is failing on the internal side,

qukhan:
Here are the errors:

- In `/litert/tensor/backends/xnnpack/conversion.cc`
  1. Line 431
     > no header providing "absl::c_any_of" is directly included
  2. Line 432
     > no header providing "absl::StrFormat" is directly included

You'll also need to update the BUILD file with the correct dependencies.

qukhan (approved):

destro4evr-rgb:
Hi @qukhan, updated the BUILD file with the correct dependencies, could you please verify & approve it ?

destro4evr-rgb:
Hi @qukhan, All checks have passed, could you please have a look at this?

qukhan:
Hi, I'm sorry, there's a big change that is coming concerning these two files[^1] and that will make this PR obsolete. Given the difference in size. I'd rather drop this one and maybe re-create it later if needed.

[^1]: https://github.com/google/XNNPACK/commit/89efcd2a2dcc50f85189f0e63bcff024830dc54e

destro4evr-rgb:
Hi @qukhan, as suggested, re-created the fix on top of the refactored code in #11195.

PR #11105, opened 2026-08-31
litert/xnnpack: validate scale array size against channel/block count in DefineValue()

destro4evr-rgb:
`DefineValue()` in `litert/tensor/backends/xnnpack/conversion.cc` passes caller-supplied scale arrays to XNNPACK without checking that the array is large enough for the tensor's channel or block count.

**Channelwise path**: `pcq.scales.size()` was never checked against `dims[pcq.quantized_dimension]`. `xnn_validate_channelwise_quantized_tensor` then loops `dims[channel_dim]` times reading `scale[channel]`, causing an
out-of-bounds heap read when the scale array is shorter than the channel dimension - reachable from a crafted flatbuffer at model-load time.

**Blockwise path**: `bwq.scales.size()` was never checked against `dims[0] * dims[1] / block_size`. `xnn_define_blockwise_quantized_tensor_value_v2` loops `block_count` times reading the fp16 scale array, causing an
out-of-bounds heap read when the scale array is shorter than `block_count` - also reachable from a crafted flatbuffer at model-load time.

Fix: add explicit size checks before both calls and return `InvalidArgumentError` on mismatch.

### DIFF
--- a/litert/tensor/backends/xnnpack/conversion.cc
+++ b/litert/tensor/backends/xnnpack/conversion.cc
@@ -505,6 +505,10 @@ absl::StatusOr<uint32_t> XnnpackBuildContext::DefineValue(
           info.name,
           ": blockwise quantized tensor requires at least 2 dimensions"));
     }
+    if (bwq.block_size == 0) {
+      return absl::InvalidArgumentError(absl::StrCat(
+          info.name, ": blockwise quantized tensor block_size must be > 0"));
+    }
     const size_t expected_block_count =
         dims[0] * dims[1] / bwq.block_size;
     if (bwq.scales.size() < expected_block_count) {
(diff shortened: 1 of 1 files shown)


litert/tensor/backends/xnnpack/conversion.cc
```diff
@@ -478,6 +478,16 @@ absl::StatusOr<uint32_t> XnnpackBuildContext::DefineValue(
             value.flags, &value.id))
             << "Could not define a new tensor value after dequantization.";
       } else {
+        if (pcq.quantized_dimension >= dims.size() ||
```
qukhan: Should you also check the `< 0` case?

qukhan (changes requested):

litert/tensor/backends/xnnpack/conversion.cc
destro4evr-rgb: Hi @qukhan, changes made, added an explicit < 0 check.

qukhan (approved):

PR #11112, opened 2026-09-01
Optimize Arm bf16 qb4w fully connected conversion path

JakeStevens:
Two related changes:

Keep qd8_bf16_qb4w fully connected native on Arm only when an optimized i8mm or NEON dot-product kernel is available and the input, output, and weight scales use the supported BF16 types. Otherwise, retain the fp32 rewrite fallback.

Add NEON v8 BF16-to-QS8 conversion microkernels and select the u32 kernel for Arm packed-LHS conversion. This replaces the scalar-imagic conversion path on Arm while preserving scalar and x86 fallback behavior.

Test plan:
```
      cmake --build build-test --target bf16-qs8-vcvt-test -j8
      ./build-test/test/bf16-qs8-vcvt-test  # 64/64 pass
      cmake --build build-android-arm64 --target bf16-qs8-vcvt-test -j8
``` 
Benchmark (Pixel 10, Android 17/API 37, arm64-v8a):
Configured a Release build for Android API 24 with NDK r29/Clang 21.
```
      cmake --build build-android-arm64 --target vunary-bench -j8
      adb push build-android-arm64/bench/vunary-bench /data/local/tmp/
      adb shell /data/local/tmp/vunary-bench --benchmark_filter='bf16_qs8_vcvt.*(neonv8_u32|scalar_imagic_u4)'
```
Results (real time, CPU scaling and ASLR enabled):
```
        N=8192:  neonv8_u32 1110 ns, 22.1348 GB/s; scalar_imagic_u4 1182 ns, 20.7856 GB/s
        N=65536: neonv8_u32 8886 ns, 22.1246 GB/s; scalar_imagic_u4 9455 ns, 20.7941 GB/s
        neonv8_u32 was 1.064-1.065x faster than scalar_imagic_u4.
```


test/subgraph/bf16-to-fp32-fallback.cc
```diff
@@ -55,6 +57,78 @@ using litert::tensor::XnnpackGraph;
...
+struct InlineQd8Bf16Qb4wFullyConnected {
+  xnn_subgraph_t subgraph = nullptr;
+  uint32_t input_id = XNN_INVALID_VALUE_ID;
+  uint32_t convert_output_id = XNN_INVALID_VALUE_ID;
+  uint32_t weights_id = XNN_INVALID_VALUE_ID;
+  uint32_t output_id = XNN_INVALID_VALUE_ID;
+  std::vector<uint16_t> weights_scale = {0x3F80, 0x3F80};
+  std::vector<uint8_t> weights_data = std::vector<uint8_t>(32, 0x88);
+
+  void Build(enum xnn_datatype input_datatype,
```
qukhan, line 69: I'm wondering how difficult it would be to use the Tensor API to build this test graph like was done in https://github.com/google/XNNPACK/blob/a5acbbec8f21a1903bbe8ef711f4fc309970ee6d/test/subgraph/fp16-to-fp32-fallback.cc#L128-L159

JakeStevens: done

fbarchard (approved):
overall looks okay to me.  i wonder why it is neonv8 specific?  fp16 rounding is neonv8 but this is bf16 and normal neon should work, or neonbf16 if you want to optimize for the vcvt with rounding.

JakeStevens:
Thanks! The `neonv8` is for the FP32-to-S32 quantization step rather than the BF16 widening: we use `vcvtnq_s32_f32` for round-to-nearest, and that instruction requires Armv8. 

 A baseline-NEON variant could use magic-bias rounding approach but the FC path for this is already gated on dotprod/i8mm, so its target hardware is already Armv8+.

Aelphy:
Seems like this PR breaks Bf16ToFp32FallbackTest.RewritesScalarQd8Fallback when compiled for android_arm

Aelphy:
[XNNPACK/test/subgraph/bf16-to-fp32-fallback.cc:84]: Failure
Value of: litert_expected_value_or_error_84.ok()
  Actual: false
Expected: true
UNKNOWN: xnn_status=2

JakeStevens:
fixed the scalar fallback test

 gn-windows-x86_64 failure seems preexisting

Aelphy (dismissed):

qukhan (approved):

JakeStevens:
updated to incorporate name change (upstream switched to GetSubgraph()) that occurred between last PR update and CI running

JakeStevens:
Anything I need to do here?

PR #11123, opened 2026-09-02
Add WASM FP16 rmax

rahulchaphalkar:


fbarchard (approved):

fbarchard (approved):

src/configs/reduce-config.c
```diff
@@ -778,16 +778,16 @@ static void init_f16_rmax_config(void) {
...
     #if XNN_ENABLE_AVX512SKX
       if (hardware_config->arch_flags & xnn_arch_x86_avx512skx) {
         f16_rmax_config.ukernel = XNN_INIT_REDUCE_UKERNEL(xnn_f16_rmax_ukernel__avx512skx_u64_acc4);
       } else
     #endif
+    #if XNN_ENABLE_F16C
+      if (hardware_config->arch_flags & xnn_arch_x86_f16c) {
+        f16_rmax_config.ukernel = XNN_INIT_REDUCE_UKERNEL(xnn_f16_rmax_ukernel__f16c_u32);
+      } else
+    #endif
```
qukhan, line 790: Why do you move this block?

qukhan (changes requested):

src/configs/reduce-config.c
rahulchaphalkar: To preserve the dispatch priority: AVX512FP16, AVX512SKX with FP32 accumulation, then F16C. 
Previously, the independent AVX512SKX/scalar block overwrote the AVX512FP16 or F16C selection because   `else` was missing before the AVX512SKX check.

rahulchaphalkar:
CI job seems to be failing with unrelated issues. Copilot says -
The job is failing because 51 actions in build.yml are not pinned to specific commit hashes, The errors are scattered throughout the workflow, with examples at:

Line 32: uses: actions/checkout@v6
Line 35: uses: actions/cache/restore@v5
Line 53: uses: actions/cache/save@v5
Lines 215, 221, 247, 253, etc.

qukhan (approved):

PR #11156, opened 2026-09-07
litert/tensor: guard AveragePool2D/MaxPool2D/Conv2D/DepthwiseConv2D against zero stride in arithmetic.h

destro4evr-rgb:
AveragePool2D, MaxPool2D, Conv2DImpl, and DepthwiseConv2DImpl all divide by `stride_h` and `stride_w` to compute output spatial dimensions without first checking that either value is non-zero. On x86-64 and ARM, integer division by zero raises SIGFPE, crashing the process at graph-construction time before any inference runs.

Add a `stride > 0` guard in each function immediately before the shape reads, consistent with the validation pattern already used in `ComputeTransposeConvPadding` in `arithmetic.cc`.

**Affected functions:**
- `AveragePool2D` - lines 1226–1230
- `MaxPool2D` - lines 1266–1270
- `Conv2DImpl` - lines 1314–1320
- `DepthwiseConv2DImpl` - lines 1388–1394

### DIFF
--- a/litert/tensor/arithmetic.h
+++ b/litert/tensor/arithmetic.h
@@ -1217,6 +1217,10 @@ Tensor<Mixins...> AveragePool2D(
   graph::TensorInformation& output_info = *GetInfo(output.GetRaw());
   output_info.type = input_info.type;
 
+  if (stride_h <= 0 || stride_w <= 0) {
+    return Tensor<Mixins...>(graph::ErrorTensor(absl::InvalidArgumentError(
+        "AveragePool2D strides must be greater than zero.")));
+  }
   const int input_h = input_info.shape[1];
   const int input_w = input_info.shape[2];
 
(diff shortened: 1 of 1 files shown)


litert/tensor/arithmetic.h
```diff
@@ -1217,6 +1217,10 @@ Tensor<Mixins...> AveragePool2D(
   graph::TensorInformation& output_info = *GetInfo(output.GetRaw());
   output_info.type = input_info.type;
 
+  if (stride_h <= 0 || stride_w <= 0) {
```
qukhan, line 1220: Why should the strides not be negative?

destro4evr-rgb: The zero case is what causes the SIGFPE - I added <= 0 to also reject negative strides since they'd give a negative output shape, but that's beyond the scope of this fix. I can change it to == 0 if that's cleaner.

destro4evr-rgb:
Hi @qukhan, I've updated the guard to == 0 only (pure SIGFPE fix - negative strides are a separate concern). Can you check it once and approve it ?

PR #11171, opened 2026-09-08
FP16 GOI pack kernels

rahulchaphalkar:
wasmrelaxedsimdfp16 F16 GOI pack kernels
AVX512FP16 GOI pack native kernels
Updated gemm-config.cc to wire in kernels

### DIFF
--- a/src/configs/gemm-config.c
+++ b/src/configs/gemm-config.c
@@ -335,7 +335,14 @@ static void init_f16_gemm_config(void) {
         f16_gemm_config.nr = 32;
       } else
     #endif
-    if (hardware_config->arch_flags & xnn_arch_x86_avx2) {
+    if ((hardware_config->arch_flags & xnn_arch_x86_avx2)
+    #if XNN_ENABLE_AVX512SKX
+      && !(hardware_config->arch_flags & xnn_arch_x86_avx512skx)
+    #endif
+    #if XNN_ENABLE_AVX512FP16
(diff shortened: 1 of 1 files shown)


fbarchard:
i suggest 2 changes

1. fix config issue where AVX2 was always selected.  But use avx512skx with prfm
a new kernel (avx512fp16) should not be necessary
But if you want to try, i suggest looking for vbmi optimizations. (icelake).  It has permi2w and other instructions that can help with the transpose

2. enable wasmsimd gemm.  remove fp16 checks if possible by have a default f16_gemm using scalar as a fallback.  otherwise do else

src/configs/gemm-config.c
```diff
@@ -335,7 +335,14 @@ static void init_f16_gemm_config(void) {
         f16_gemm_config.nr = 32;
       } else
     #endif
-    if (hardware_config->arch_flags & xnn_arch_x86_avx2) {
+    if ((hardware_config->arch_flags & xnn_arch_x86_avx2)
```
fbarchard, line 338: the avx2 should be in #else not #endif so if avx512 is selected, the avx2 will not.

fbarchard (changes requested):

src/configs/gemm-config.c
rahulchaphalkar: Yes. I'm going to close this PR, it became too noisy and I think the existing AVX512SKX is fine as is. Separate PR for wasmsimdfp16.

PR #11186, opened 2026-09-10
fix(subgraph): reject fully-connected tensors exceeding XNN_MAX_TENSOR_DIMS (fixes #11036)

### LINKED ISSUE #11036: XNNPACK delegate accepts rank-7 FULLY_CONNECTED then fails during Prepare
## Summary

The default XNNPACK delegate claims a rank-7 `FULLY_CONNECTED` node and then
fails during delegate preparation.  This aborts interpreter setup, although
the same FlatBuffer runs correctly with the builtin kernels when default
delegates are disabled.

## Environment

- LiteRT: `ai-edge-litert 2.1.6`
- TensorFlow converter: `tf-nightly 2.22.0.dev20260808` (also reproduced on
  TensorFlow 2.20.0)
- Linux x86_64 CPU, Python 3.11

## Minimal reproduction

```python
import tensorflow as tf
from ai_edge_litert.interpreter import Interpreter, OpResolverType

x1 = tf.constant([[1.0, 2.0], [3.0, 4.0], [5.0, 6.0]], shape=[1, 3, 2])


class Model(tf.keras.Model):
    def __init__(self):
        super().__init__()
        self.w1 = tf.Variable([1.0] * 6)

    def call(self, x):
        w = tf.reshape(self.w1, (6, 1, 1, 1))
        v = tf.reshape(x, (1, 1, 1, 6))
        y = tf.tensordot(v, w, axes=1)
        return tf.expand_dims(y, axis=-1)  # output rank: 7


model = Model()
print("eager:", model(x1).shape, model(x1).numpy().flatten())
flatbuffer = tf.lite.TFLiteConverter.from_keras_model(model).convert()


def run(**kwargs):
    interpreter = Interpreter(model_content=flatbuffer, **kwargs)
    interpreter.allocate_tensors()
    input_detail = interpreter.get_input_details()[0]
    interpreter.set_tensor(input_detail["index"], x1.numpy())
    interpreter.invoke()
    return interpreter.get_tensor(interpreter.get_output_details()[0]["index"])


print(
    "builtin only:",
    run(
        experimental_op_resolver_type=
        OpResolverType.BUILTIN_WITHOUT_DEFAULT_DELEGATES
    ).flatten(),
)
print("default delegates:", run().flatten())
```

## Actual result

```text
eager: (1, 1, 1, 1, 1, 1, 1) [21.]
builtin only: [21.]
RuntimeError: failed to delegate FULLY_CONNECTED node #1
Node number 2 (TfLiteXNNPackDelegate) failed to prepare.
```

The default-delegate run raises from `allocate_tensors()` or `invoke()` rather
than returning a result.

## Expected result

The delegate should either support this node or decline to claim it during
partitioning/support checking.  In the latter case, the builtin CPU kernel
should execute the node automatically, as demonstrated by the successful
`BUILTIN_WITHOUT_DEFAULT_DELEGATES` run.  A preparation failure after the node
has been delegated makes an otherwise runnable model fail to load.

## Context

The same behavior was reproduced through both `tf.lite.Interpreter` and
`ai_edge_litert`.  LiteRT maintainers reproduced the failure and asked that it
be tracked in XNNPACK: https://github.com/google-ai-edge/LiteRT/issues/9215

jdymitarai:
### Problem Description

Fixes #11036

When a `FULLY_CONNECTED` operator is defined or reshaped with tensor rank exceeding `XNN_MAX_TENSOR_DIMS` (e.g. rank-7 tensors), the operator was accepted during define time without dimension checks, and subsequent prepare / reshape phases could fail or index `output->shape.dim` / `filter_value->shape.dim` out-of-bounds.

In LiteRT / TensorFlow Lite delegation, this caused the XNNPACK delegate to improperly claim rank-7 `FULLY_CONNECTED` nodes and fail during prepare/allocation, preventing clean fallback to builtin CPU kernels.

### Solution

1. In `src/subgraph/fully-connected.c`:
   - Added validation in `xnn_define_fully_connected` to reject `input_value`, `kernel_value`, `bias_value`, and `output_value` shapes with `num_dims > XNN_MAX_TENSOR_DIMS` with `xnn_status_unsupported_parameter`.
   - In `create_fully_connected_operator`, validated `filter_value->shape.num_dims` to ensure it is between 2 and `XNN_MAX_TENSOR_DIMS` before accessing `filter_value->shape.dim[filter_value->shape.num_dims - 1]`.
   - In `resize_fully_connected_output_tensor` and `reshape_fully_connected_operator`, added dimension checks to reject `num_dims > XNN_MAX_TENSOR_DIMS` with `xnn_status_unsupported_parameter`.
2. In `test/subgraph/fully-connected.cc`:
   - Added regression test `FullyConnectedF32.rank_exceeding_max_dims_rejected` verifying that `xnn_define_fully_connected` returns `xnn_status_unsupported_parameter` when a tensor's rank exceeds `XNN_MAX_TENSOR_DIMS`.

### DIFF
--- a/test/subgraph/fully-connected.cc
+++ b/test/subgraph/fully-connected.cc
@@ -1423,4 +1423,39 @@ TEST(FullyConnectedQD8F32QC8W, reshape_rejects_input_channel_mismatch) {
   ASSERT_NE(xnn_status_success, subgraph.Status());
 }
 
+// Regression test: fully-connected with tensor rank exceeding XNN_MAX_TENSOR_DIMS
+// must be rejected as unsupported parameter (fixes #11036).
+TEST(FullyConnectedF32, rank_exceeding_max_dims_rejected) {
+  ASSERT_EQ(xnn_status_success, xnn_initialize(nullptr));
+
+  xnn_subgraph_t subgraph = nullptr;
+  ASSERT_EQ(xnn_status_success, xnn_create_subgraph(3, 0, &subgraph));
+
+  const uint32_t input_id = 0;
+  subgraph->values[input_id].type = xnn_value_type_dense_tensor;
+  subgraph->values[input_id].datatype = xnn_datatype_fp32;
+  subgraph->values[input_id].shape.num_dims = XNN_MAX_TENSOR_DIMS + 1;
+
+  float filter = 1.0f;
+  size_t filter_dims[] = {1, 1};
+  uint32_t filter_id = XNN_INVALID_VALUE_ID;
+  ASSERT_EQ(xnn_status_success,
+            xnn_define_tensor_value(subgraph, xnn_datatype_fp32, 2,
+                                    filter_dims, &filter, XNN_INVALID_VALUE_ID,
+                                    0, &filter_id));
+
+  uint32_t output_id = XNN_INVALID_VALUE_ID;
+  ASSERT_EQ(xnn_status_success,
+            xnn_define_tensor_value(subgraph, xnn_datatype_fp32, 0, nullptr,
+                                    nullptr, 1, XNN_VALUE_FLAG_EXTERNAL_OUTPUT,
+                                    &output_id));
+
+  ASSERT_EQ(xnn_status_unsupported_parameter,
+            xnn_define_fully_connected(subgraph, -1e30f, 1e30f, input_id,
+                                       filter_id, XNN_INVALID_VALUE_ID,
+                                       output_id, 0));
+
+  xnn_delete_subgraph(subgraph);
+}
+
 }  // namespace xnnpack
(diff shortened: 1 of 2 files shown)


test/subgraph/fully-connected.cc
```diff
@@ -1423,4 +1423,39 @@ TEST(FullyConnectedQD8F32QC8W, reshape_rejects_input_channel_mismatch) {
   ASSERT_NE(xnn_status_success, subgraph.Status());
 }
 
+// Regression test: fully-connected with tensor rank exceeding XNN_MAX_TENSOR_DIMS
+// must be rejected as unsupported parameter (fixes #11036).
```
qukhan: This comment is useless. Please reign-in your LLMs before sending PRs.

qukhan (changes requested):

test/subgraph/fully-connected.cc
jdymitarai: Removed the comment in 96188e9. Thank you for the review.

PR #11191, opened 2026-09-10
Add native, WASM, and scalar F16 raddstoreexpminusmax kernels

rahulchaphalkar:
- Adds FP16 `raddstoreexpminusmax` kernels for scalar, x86, and wasm
- Implemented the required `reduce_add_f16` operations
- Required for making wasm mobilenet f16 variants work

### DIFF
--- a/src/configs/raddstoreexpminusmax-config.c
+++ b/src/configs/raddstoreexpminusmax-config.c
@@ -61,7 +61,14 @@ static void init_f16_raddstoreexpminusmax_config(void) {
     if (hardware_config->arch_flags & xnn_arch_riscv_vector_fp16_arith) {
       f16_raddstoreexpminusmax_config.ukernel = XNN_INIT_RADDSTOREEXPMINUSMAX_UKERNEL(xnn_f16_raddstoreexpminusmax_ukernel__rvvfp16arith_rr2_p2_u4v);
     }
+  #elif XNN_ARCH_WASMRELAXEDSIMDFP16
+    f16_raddstoreexpminusmax_config.ukernel = XNN_INIT_RADDSTOREEXPMINUSMAX_UKERNEL(xnn_f16_raddstoreexpminusmax_ukernel__wasmrelaxedsimdfp16_rr2_p2_u8);
+  #elif XNN_ARCH_WASMRELAXEDSIMD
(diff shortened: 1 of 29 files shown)


src/configs/raddstoreexpminusmax-config.c
```diff
@@ -61,7 +61,14 @@ static void init_f16_raddstoreexpminusmax_config(void) {
     if (hardware_config->arch_flags & xnn_arch_riscv_vector_fp16_arith) {
       f16_raddstoreexpminusmax_config.ukernel = XNN_INIT_RADDSTOREEXPMINUSMAX_UKERNEL(xnn_f16_raddstoreexpminusmax_ukernel__rvvfp16arith_rr2_p2_u4v);
     }
+  #elif XNN_ARCH_WASMRELAXEDSIMDFP16
+    f16_raddstoreexpminusmax_config.ukernel = XNN_INIT_RADDSTOREEXPMINUSMAX_UKERNEL(xnn_f16_raddstoreexpminusmax_ukernel__wasmrelaxedsimdfp16_rr2_p2_u8);
+  #elif XNN_ARCH_WASMRELAXEDSIMD
+    f16_raddstoreexpminusmax_config.ukernel = XNN_INIT_RADDSTOREEXPMINUSMAX_UKERNEL(xnn_f16_f32acc_raddstoreexpminusmax_ukernel__wasmrelaxedsimd_rr2_p5_u4);
   #endif
+  if (f16_raddstoreexpminusmax_config.ukernel == NULL) {
```
rahulchaphalkar, line 69: This is intentional, the `#endif` followed by a separate `if` instead of an `#else`. For example, on x86 if cpu lacks AVX2, then this will assign scalar instead of NULL.

src/xnnpack/simd/f16-avx512fp16.h
```diff
@@ -197,6 +197,22 @@ static XNN_INLINE xnn_simd_f16_t xnn_set1_f16(xnn_float16 v) {
 #endif  // XNN_HAVE_FLOAT16
 }
 
+static XNN_INLINE float xnn_reduce_add_f16(xnn_simd_f16_t a) {
```
fbarchard, line 200: nit: on arm there is 'addv' to add all f16 elements in a vector and return an f16.
i expect that is the intent of a reduction add, but is not supported directly by avx512fp16.  you could do divide and conquer to do first half added to second half, etc.  This is fine, but consider a comment or optimizing it to use add_ph or if intel ever adds it vdpphps or hadd_ph

PR #11195, opened 2026-09-11
litert/xnnpack: reject negative tensor dimensions in DefineTensorValue()

destro4evr-rgb:
## Problem

`XnnpackBuildContext::DefineTensorValue()` in `litert/tensor/backends/xnnpack/conversion.cc` converts `info.shape`
(a `std::vector<int>`) directly to `std::vector<size_t>` without checking for negative values:

```cpp
std::vector<size_t> dims(info.shape.begin(), info.shape.end());
```

A negative `int` dimension silently wraps to a large `size_t` value via the implicit signed-to-unsigned conversion. That large value is then passed to `xnn_define_tensor_value()`, which either crashes or produces a miscomputed tensor layout.

The previous code had a `NumElements()` helper that explicitly checked for negative dimensions before this conversion. That helper was removed in the refactor from commit `60bd2ebd`, leaving the conversion unguarded.

## Fix

Add an explicit negative-dimension check immediately before the conversion:

```cpp
for (int dim : info.shape) {
  if (dim < 0) {
    return absl::InvalidArgumentError(
        absl::StrCat(info.name, ": negative tensor dimension ", dim, "."));
  }
}
std::vector<size_t> dims(info.shape.begin(), info.shape.end());
```

## Note

This is a replacement for #11092, which became non-applicable after the refactor in `60bd2ebd` restructured `conversion.cc`. As suggested by @qukhan in that PR, re-creating the fix on top of the new code.


Aelphy:
I think we want to keep negative shape support, e.g., -1

destro4evr-rgb:
Hi @Aelphy, does -1 actually reach DefineTensorValue in practice, or is dynamic shape resolution expected to happen upstream before the XNNPACK backend processes tensors? If -1 can legitimately arrive here, I'll update the fix to allow it through while still blocking other negative values.

Aelphy:
Since you are rejecting negatives here, it already assumes -1 could reach here. Though I am not sure we should just reject negatives vs controlling memory overflow.

destro4evr-rgb:
Hi @Aelphy , I've updated the fix to address this. Instead of blanket-rejecting all negatives, the loop now maps `-1` to `0` (so XNNPACK sees a zero-size dynamic dimension rather than `SIZE_MAX`) and only rejects values `< -1` as invalid. This preserves `-1` dynamic shape support while still preventing the overflow.

destro4evr-rgb:
Hi @Aelphy, just a gentle ping - I pushed the updated fix on Sep 15 that maps -1 to 0 and only rejects < -1. Can you take a look and approve it ?

Aelphy:
Could you please rebase on top of master. Seems like import has staled.

destro4evr-rgb:
Hi @Aelphy, I've rebased it, can you check and approve it ?

PR #11203, opened 2026-09-11
Fix the output shape of the static_broadcast lowering

EylonKrause:
## Problem

`static_broadcast` has no runtime kernel (`reshape_copy_operator` returns `xnn_status_unsupported_parameter` for it), so `optimize_common_subgraphs_broadcast` is its only implementation. Both of its paths can produce the wrong output shape.

**1. Elision into a binary / batch-matmul consumer ignores the other operand.** The pass replaces the consumer's input with the *un-broadcast* value whenever the consumer broadcasts implicitly, without checking that the consumer's own broadcasting reaches the same shape:

```c
out = add(broadcast(s[1] -> [4, 5]), w[1, 5]);   // must be [4, 5]
```

becomes `add(s[1], w[1, 5])`, i.e. `[1, 5]`: 5 of the 20 output elements are written. (For `batch_matrix_multiply` only the batch dimensions broadcast, and `XNN_FLAG_NO_BROADCAST` disables even that.)

**2. The add-of-zero fallback aligns dimensions from the front.** Broadcasting aligns *trailing* dimensions, but the zero tensor's shape is built by comparing `new_shape[k]` with `old_shape[k]` at the same index:

```c
shape[k] = (new_shape[k] == 0 || new_shape[k] == old_shape[k]) ? 1 : new_shape[k];
```

so for `abs(broadcast(x[4] -> [4, 4]))` dimension 0 is taken to be "already there" (`new_shape[0] == old_shape[0] == 4`) and the result is `[1, 4]` instead of `[4, 4]`.

Both are wrong with the default flags; `XNN_FLAG_NO_OPERATOR_FUSION` fails the graph outright (no kernel), so there is no correct path today.

## Fix

- Elide the broadcast into a consumer only when doing so leaves the consumer's result shape unchanged, computed with a small `broadcast_leading_dims` helper (`batch_matrix_multiply` excludes the two matrix dimensions, which must not change, and is skipped entirely under `XNN_FLAG_NO_BROADCAST`).
- Build the fallback's zero tensor with trailing-dim alignment, so new leading dimensions and grown dimensions are both materialised.
- Count a consumer once even when it uses the broadcast for several inputs, and then rewrite all of those inputs together.

## Validation

- New `StaticBroadcastRewrite.ElidedIntoBinaryOnlyIfShapeUnchanged` and `StaticBroadcastRewrite.MaterializesRankIncreasingBroadcast` in `test/subgraph/broadcast.cc`: both fail on `master` and pass with this change. They also cover the cases that *should* still be elided.
- The existing randomized `Broadcast` tests (all datatypes, ranks 1..`XNN_MAX_TENSOR_DIMS`) still pass, as do `binary`, `fusion`, `rewrites`, `copy`, `subgraph`, `static-reshape`, `static-expand-dims`, `batch-matrix-multiply` and `split-fuse` (Linux x86-64, Release).
- Nine further hand-written control cases (equal-rank fallbacks, scalar fallback, rank-up with a coinciding dimension, `0`-marker shapes, broadcast as either operand, a consumer using the broadcast twice) match a plain C reference implementation of numpy-style broadcasting.
- Found by a random-subgraph differential fuzzer comparing the default runtime against `XNN_FLAG_NO_OPERATOR_FUSION`, `XNN_FLAG_SLOW_CONSISTENT_ARITHMETIC` and a node-by-node chain (~35 hits).

Disclosure: this contribution was authored with an AI coding assistant (Claude) and reviewed and validated before submission.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/subgraph.c
+++ b/src/subgraph.c
@@ -2696,6 +2696,81 @@ static enum xnn_status optimize_common_subgraphs_reduce_sum_to_square(
   return xnn_status_success;
 }
 
+// Computes the shape of the result of broadcasting `a` against `b` over all
+// but their last `num_trailing_dims` dimensions (which are left out of the
+// result). Returns false if the two are not broadcast-compatible.
+static bool broadcast_leading_dims(const struct xnn_shape* a,
+                                   const struct xnn_shape* b,
+                                   size_t num_trailing_dims,
+                                   struct xnn_shape* result) {
+  if (a->num_dims < num_trailing_dims || b->num_dims < num_trailing_dims) {
+    return false;
+  }
+  const size_t a_rank = a->num_dims - num_trailing_dims;
+  const size_t b_rank = b->num_dims - num_trailing_dims;
+  const size_t rank = max(a_rank, b_rank);
+  result->num_dims = rank;
+  for (size_t i = 0; i < rank; i++) {
+    // Broadcasting aligns the trailing dimensions; missing leading ones are 1.
+    const size_t a_dim = i + a_rank >= rank ? a->dim[i + a_rank - rank] : 1;
+    const size_t b_dim = i + b_rank >= rank ? b->dim[i + b_rank - rank] : 1;
+    if (a_dim != b_dim && a_dim != 1 && b_dim != 1) {
+      return false;
+    }
+    result->dim[i] = max(a_dim, b_dim);
+  }
+  return true;
+}
+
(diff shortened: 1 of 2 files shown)


dsharlet:
Thanks for the PR. I'm sure this is a real issue, but, we need to just remove static_broadcast, it is not used by anything in production, and should not be.

PR #11215, opened 2026-09-14
Add scalar and WASM FP16 VADD and VADDC kernels

rahulchaphalkar:
Vadd / Vaddc kernels for enabling wasm mobilenet variants.

- Direct wasm f16 kernels `wasmrelaxedsimdfp16`
- F16-f32acc wasm kernels `wasmrelaxedsimd`
- scalar fallback

PR #11217, opened 2026-09-14
Add scalar and WASM FP16 VMUL and VMULC kernels

rahulchaphalkar:
Vmul and Vmulc kernels for enabling wasm mobilenet variants.

- Direct wasm f16 kernels wasmrelaxedsimdfp16
- F16-f32acc wasm kernels wasmrelaxedsimd
- scalar fallback
Note - The templates here like `*vop.c.in` and `*vopc.c.in` are already included in the `Vadd` PR https://github.com/google/XNNPACK/pull/11215 , so only 1 copy would be retained. I'll remove when one of the PR merges.


fbarchard (approved):

fbarchard (changes requested):
missing f16-vmulc.inc   and  f16-vmul.inc

fbarchard (approved):

Aelphy:
failing fp32 fallback is a red herring here. I think that vectorized fp32 path is better than scalar fp16.

PR #11218, opened 2026-09-14
Add scalar and WASM FP16 VHSwish kernels

rahulchaphalkar:
`vhswish` kernels for enabling wasm mobilenet variants.

- Direct wasm f16 kernels wasmrelaxedsimdfp16
- F16-f32acc wasm kernels wasmrelaxedsimd
- scalar fallback


src/configs/unary-elementwise-config.c
fbarchard, line 4305: !xnn_is_f16_compatible_config(hardware_config)) can be removed because there is always a scalar version?

fbarchard (approved):

src/f16-vhswish/gen/f16-f32acc-vhswish-wasmrelaxedsimd-u8.c
```diff
@@ -0,0 +1,89 @@
...
+  XNN_SIMD_CONST_F32(vone, 1.0f);
+  XNN_SIMD_CONST_F32(vzero, 0.0f);
+
+  xnn_simd_f32_t vacc = xnn_fmadd_f32(vx, vsixth, vhalf);
+  vacc = xnn_max_f32(vacc, vzero);
+  vacc = xnn_min_f32(vacc, vone);
+  return xnn_mul_f32(vacc, vx);
+}
+
+void xnn_f16_f32acc_vhswish_ukernel__wasmrelaxedsimd_u8(
```
fbarchard, line 55: i think these should also be wasmrelaxedsimdfp16 because the conversion wont work without fp16 experiment?

fbarchard (changes requested):

src/configs/unary-elementwise-config.c
rahulchaphalkar: This is for `log_config`, the `hswish_config` doesn't seem to have this check

src/f16-vhswish/gen/f16-f32acc-vhswish-wasmrelaxedsimd-u8.c
rahulchaphalkar: This works without `--experimental-wasm-fp16` in d8

PR #11243, opened 2026-09-18
Fix YNNPACK broadcast partial reductions

jasperjonkhans:
## Summary

Fix YNNPACK partial-output cropping when Slinky omits trailing singleton broadcast dimensions. Iterate over every reduction bound rather than stopping at the input buffer's physical rank, so parallel tasks do not write another task's partial output.

This reproduces the wrong, nondeterministic CPU sums reported in tensorflow/tensorflow#126812. The fix was initially carried as a dependency patch in tensorflow/tensorflow#127196; that review requested an upstream submission here.

Add `Reduce.BroadcastPartialReduction`, which directly invokes a partial-reduction tile and checks that another tile's output remains untouched, followed by 100 multithreaded graph invocations. Use the current scheduler/runtime test helpers and declare the direct Slinky Bazel dependencies.

## Validation

- Before the production fix: the deterministic sentinel check fails at partial-output index 32, and the graph returns incorrect sums.
- With the fix: all 45 native reduction tests pass.
- Ten regression-test repetitions: 1,000 correct parallel graph invocations, each returning exactly 2097152.
- Native CMake build using GCC 16.2.1 and generic YNNPACK kernels.

Bazel validation is blocked before loading the test target by an existing module compatibility conflict: the pinned googletest requires abseil-cpp compatibility level 0, while the root module requires level 1. No dependency versions were changed. ISA-specific kernels and full TensorFlow integration were not tested.


ynnpack/subgraph/test/reduce.cc
```diff
@@ -26,13 +31,108 @@
...
+  const size_t output_dims[] = {2, 1024, 1024};
+  ASSERT_EQ(ynn_define_static_broadcast(subgraph.get(), 3, output_dims,
+                                        expanded, &broadcast, 0),
+            ynn_status_success);
+  const int32_t axes[] = {0, 1, 2};
+  ASSERT_EQ(ynn_define_reduce(subgraph.get(), ynn_reduce_sum, 3, axes,
+                              broadcast, YNN_INVALID_VALUE_ID, &output_id, 0),
+            ynn_status_success);
+
+  Runtime runtime(subgraph.get(), &scheduler, /*flags=*/0, /*optimize=*/false);
```
dsharlet, line 78: I don't think this test should or needs to dig into the internal implementation like this. Just running the reproducer and validating the result should be enough.

dsharlet (changes requested):
Thanks for the fix!

ynnpack/subgraph/test/reduce.cc
dsharlet: Also, we've been putting specific issue reproducers in tests in this folder: https://github.com/google/XNNPACK/tree/master/ynnpack/subgraph/test/reproducers

